Issue #67, opened 2019-01-14
Implement modal dialog for table properties in HTML editor


alansemenov:
1) Remove the native "Cell properties" dialog (or at least hide it from the context menu).
2) Implement our own Table properties dialog which will have the following properties from the native one:
* Rows/Columns (readonly for existing table)
* Headers
* Cell spacing
* Cell padding
* Border thickness
* Alignment

PR #71, opened 2019-01-16
Implement modal dialog for table properties in HTML editor #67

ashklianko:


ashklianko:
> Looks/works great but it seems like you used `dialog.link.tabname.anchor` text resource for the "Headers" and "Alignment" fields :)
> Feel free to translate literals to russian as well.

Updated :)

PR #105, opened 2019-01-25
Improvements of the detail panel #8

edloidas:
* Renamed details panel to context panel: both classes and queries.
* Updated position of the context panel.
* Updated context panel dropdown to support filtering and icons.
* Replace the details panel icon with the "cog" icon.
* Passed repository and branch names as parameters, when querying the widgets.


src/main/resources/assets/js/app/view/context/WidgetsSelectionRow.ts
```diff
@@ -118,3 +124,50 @@ export class WidgetViewOption {
...
+    }
+
+    resolveSubName(object: WidgetViewOption): string {
+        const name = object.getWidgetView().getWidgetName();
+        const description = object.getWidgetView().getWidgetDescription() || 'No description';
+        switch (name) {
+        case 'Properties': return 'Properties of the content';
+        case 'Version history': return 'History of the content\'s changes';
+        case 'Dependencies': return 'Dependencies tree of the content';
+        default:
```
alansemenov: You cannot rely on hardcoded widget names to resolve description or icon because they are localised. Try switching your browser from English to any other language and it will stop working. We need something like widgetId for internal widgets that won't depend on browser language.

src/main/resources/assets/js/app/view/context/WidgetsSelectionRow.ts
```diff
@@ -118,3 +124,50 @@ export class WidgetViewOption {
...
+    }
+
+    resolveIconUrl(object: WidgetViewOption): string {
+        return object.getWidgetView().getWidgetIconUrl();
+    }
+
+    resolveIconClass(object: WidgetViewOption): string {
+        const name = object.getWidgetView().getWidgetName();
+        switch (name) {
+        case 'Properties': return 'icon-info';
```
alansemenov: Let's change the icon of this widget to the same icon that we used to have for the detail panel toggler (``icon-list``). Also remove the icon-info since we no longer need it.

src/main/resources/assets/js/app/view/context/WidgetsSelectionRow.ts
```diff
@@ -118,3 +124,50 @@ export class WidgetViewOption {
...
+
+    resolveDisplayName(object: WidgetViewOption): string {
+        return object.getWidgetView().getWidgetName();
+    }
+
+    resolveSubName(object: WidgetViewOption): string {
+        const name = object.getWidgetView().getWidgetName();
+        const description = object.getWidgetView().getWidgetDescription() || 'No description';
+        switch (name) {
+        case 'Properties': return 'Properties of the content';
```
alansemenov: Descriptions of the internal widgets must be localised, just like their names. Correct english descriptions:
1) _Content details_
2) _History of content changes_
3) _References to or from other items_

src/main/resources/assets/js/app/view/context/ContextView.ts
```diff
@@ -258,7 +259,7 @@ export class DetailsView extends api.dom.DivEl {
 
     private initCommonWidgetViews() {
 
-        this.defaultWidgetView = WidgetView.create().setName(i18n('field.details')).setDetailsView(this)
+        this.defaultWidgetView = WidgetView.create().setName(i18n('field.contextPanel')).setContextView(this)
```
alansemenov: Not a very correct name for this resource - should be similar to the two other widgets. Let's rename this to ``field.contextPanel.details`` and let's change the text itself from "_Properties_" to "_Details_".

Issue #135, opened 2019-02-04
Show first available action from the Publish menu button in the toolbar


alansemenov:
We now show "Create Issue..." button when nothing is selected in the Grid or user doesn't have Publish rights. Otherwise we show the Publish button with the dropdown of actions.

Change this logic so that we always show the first available action from the dropdown when something is selected in the grid (or inside the Wizard), otherwise (when nothing is selected) show "Create Issue...".

* Add "Publish" as the first action inside the dropdown so that the logic is consistent (if publishing is available, both the Publish button and the Publish action inside the dropdown will be enabled)
* Pick the first available action from the dropdown and show it as the main action button inside the toolbar. For example, if a published content is selected then the main action button will be "Publish Tree" etc.
* Just like we use green color for "PUBLISH" button and the dropdown toggler, implement the following colors for other actions: 
- green (same as for Publish) for "Publish Tree"; 
- grey (same as "Create Issue" now) for "Unpublish";
- blue (same color as the "Create Issue" button inside the Issue List dialog) for "Create Issue"
* When nothing is selected in the Grid, hide the dropdown toggler and vertical border to the left of the button
* Otherwise (when something is selected or in the Wizard) always show the dropdown toggler and never disable it (there will always be at least one available option in the dropdown)

PR #141, opened 2019-02-06
Show first available action from the Publish menu button in the toolb…

ashklianko:
…ar #135

Also done:
-using classes/styles to minimize/maximize button instead of manually toggling it
-fixed issues to be shwon in menu when needed

To be done:
-we probably need to refactor MenuButton class also to get rid off manual minimize/ maximize, at least I don't see why it's working like this; Also could make same behaviour as for ContentPublishMenuButton to not keep main action but show first enabled action as main.
But not in priority right now

### DIFF
--- a/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
+++ b/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
@@ -687,14 +687,6 @@ export class ContentBrowsePanel
             showCreateIssueButtonByDefault: true
         });
 
-        this.contextSplitPanel.onMobileModeChanged((isMobile: boolean) => {
-            if (isMobile) {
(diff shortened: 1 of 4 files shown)


alansemenov:
@ashklianko It works fine, but in between the actions (for example after publishing) there's a brief moment when previous action is replaced by "CREATE ISSUE" before being replaced again by the next action. 

It most likely happens when we re-render the grid after content state changes, but can we avoid this somehow by debouncing? Happens in the Wizard as well. There should be no flickering, the previous action must be replaced by the next action.

ashklianko:
Ok!

ashklianko:
The thing is we already have debouncing (200ms), but that seems not enough. Will set to 500ms for now, other option is to rework grid update routine

PR #155, opened 2019-02-11
-Adjusted focus behaviour to not let focus out of dialogs

ashklianko:
-Divided modal dialog constructor init process to several phases:

1. Initializing variables (initElements() method), pure vars initialization
2. Doing actions not related to vars initializing, like setting click ignored element or setting element to focus on open or whatever that is supposed to be done after elements initialized (postInitElements() method)
3. Initializing event listeners (initListeners() method)

-Moved dom related actions, like appendChild() or addClass(), to doRender()

-Added mask/unmask methods to remove hardcoded addClass('masked')

-Added setElementToFocusOnShow() to set focus on correct element when dialogs is opened, taking into account if dialog is already rendered or not

-Saving dialog config in var


ashklianko:
Depends on https://github.com/enonic/lib-admin-ui/pull/929

Issue #148, opened 2019-02-07
Sort content dialog reopens if it's closed before all of the items are loaded 


runarmyklebust:
1. Use Firefox (happens sometimes in Chrome also, but always in FF)
2. In the content grid select an item that has a lot of child items (over 200)
3. Open the Sort dialog
4. **While the list of child items is being loaded**, drag some of the top items to change the order and quickly close the dialog (click Save or Cancel buttons or X icon)

The dialog will reopen again.

[sort-dialog-issue.mov.zip](https://github.com/enonic/xp/files/2842599/sort-dialog-issue.mov.zip)

PR #158, opened 2019-02-12
Sort content dialog reopens if it's closed before all of the items ar…

ashklianko:
…e loaded #148

-Made SortContentTreeGrid not trigger load when added to dom (it was triggering load when opening browse panel )
-Fixed loading mask to be shown when sort dialog is opened and grid is loading
-Adding/removing grid loaded listener on dialog open/close (that was issue origin, loaded handler was opening dialog even after dialog is closed)


ashklianko:
One more thing to  note:  grid loads all items in dialog ( same issue with InstallAppDialogs , empty node is visible and triggers grid loading all items). 
Separate issue for this?

Issue #29, opened 2019-01-11
ACL-entries should be consistently sorted by name


jsi:
@sgauruseu commented on [Tue Nov 20 2018](https://github.com/enonic/app-contentstudio-old/issues/549)

1. Open a folder , press "Edit Permissions" button 

![image](https://user-images.githubusercontent.com/3728712/48767200-3e241800-ecc7-11e8-962a-bd2b88339a57.png)

PR #163, opened 2019-02-13
ACL-entries should be consistently sorted by name #29

ashklianko:
-added sorting function for acl entries views


ashklianko:
Depends on https://github.com/enonic/lib-admin-ui/pull/936

Issue #174, opened 2019-02-14
Publish Tree action - implement a check for unpublished child items 


alansemenov:
Publish Tree action in the publish button menu is now always enabled if selected item has children, no matter whether they are published or not. Sometimes it looks a bit strange, for example when you know you have published the entire structure - in this case we should show "Unpublish".

We should check whether selected item(s) have any unpublished child items and enable/disable "Publish Tree" action accordingly. We already perform this check in the Publish dialog, so just reuse the same request.

In the example below a published item is selected in the grid and all of its children are published as well, however Publish Tree is enabled and shown as the main action. But if we click it, the "Publish" button in the dialog is disabled.

![image](https://user-images.githubusercontent.com/11612490/52793839-ef401180-306e-11e9-9be5-265a3c39cdf7.png)

![image](https://user-images.githubusercontent.com/11612490/52793854-f6ffb600-306e-11e9-9797-6f9459a72bca.png)

ashklianko:
During fixing this issue I've encountered multiple difficulties figuring out what is going on in ContentWizardToolbarPublishControls and ContentWizardPanel, I believe that at least rework of ContentWizardToolbarPublishControls  is required:
https://github.com/enonic/app-contentstudio/issues/200

PR #199, opened 2019-02-20
Publish Tree action - implement a check for unpublished child items #174

ashklianko:


ashklianko:
Hm, I have button states updated immediately after item with tree is published; Using folder with a child folder

Issue #208, opened 2019-02-21
Publishing Wizard - dependent items should be hidden by default 


sgauruseu:
1. Select a folder with children and click on `Publish Tree...` menu item (publish menu)

![image](https://user-images.githubusercontent.com/3728712/53163289-cd024280-35de-11e9-8d0f-41d86dd42ee9.png)

2. Modal dialog appears, click on `Show dependent items` link
3. Click on `Cancel` button .
4. click on `Publish Tree...` menu item.

Dialog opens again.
**Expected:  dependent items should be hidden**
 **Actual** :  dependent items is expanded and `Hide dependent items` link is present 

![image](https://user-images.githubusercontent.com/3728712/53163528-57e33d00-35df-11e9-9848-c68d7eb6934a.png)

PR #215, opened 2019-02-21
Publishing Wizard - dependent items should be hidden by default #208

ashklianko:
-Default values of instance variables were set after initElements invoked; Need to init instance vars in init elements

### DIFF
--- a/src/main/resources/assets/js/app/dialog/DependantItemsDialog.ts
+++ b/src/main/resources/assets/js/app/dialog/DependantItemsDialog.ts
@@ -44,15 +44,15 @@ export abstract class DependantItemsDialog
 
     private dependantsHeaderText: string;
 
-    protected loading: boolean = false;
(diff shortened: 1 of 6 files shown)


ashklianko:
> It's unclear to me when it happened (maybe even in 6.x) but `showDependantList` config of a modal dialog doesn't seem to be working properly now.
> 
> Earlier it was used to configure whether the list of dependant items should be expanded or collapsed by default when the dialog opens. The only dialog where we should expand the list by default is ContentDeleteDialog (user should see all of the items that will be deleted). Which means that `showDependantList` property should be `false` by default and only be set to `true` in the config of the ContentDeleteDialog.
> 
> However, the value of this property doesn't seem to do anything now. It's implicitly set to `false` in ContentPublishDialog and explicitly to `true` for ContentPublishDialog but in both dialogs the list is collapsed by default.

Updated!

Issue #209, opened 2019-02-21
Part/layout configs: values  in inspection panel are not set after page reload


ashklianko:
1. Open site (Features)
2. Add part(s) with configs (Image filters test in Features)
3. Set some values for part's config in inspection panel 
4. Save and reload
5. Select part(s) and check inspection panel

Result: values are not set, but they present in repo (checked via toolbox)

PR #221, opened 2019-02-25
Part/layout configs: values in inspection panel are not set after pag…

ashklianko:
…e reload #209

-Refactored Part and Layout inspection panels to remove duplicated code
-Making config's form view to be rendered for inspection panels immediately via formView.setLazyRender(false), otherwise form view not rendered when inspection panel is hidden (panel is sliding with animation and lazy rendering thinks that panel is not in viewport yet)


ashklianko:
Depends on https://github.com/enonic/lib-admin-ui/pull/943

alansemenov:
Nice refactoring!

Issue #260, opened 2019-03-04
Error in console after publishing a shortcut with an image in target


sgauruseu:
1. Open new shortcut-wizard .
2. Type a name , select an image in target (bug does not reproduce when target is site)
3. Save the shortcut 
4. Click on Publish button in wizard and publish the content 

BUG -  error appears in console 
![image](https://user-images.githubusercontent.com/3728712/53733072-d9b15100-3e90-11e9-9a1b-993b869c7c35.png)

`_all.js:1 TypeError: Cannot read property 'init' of undefined
    at ContentWizardPanel.initPageModel (http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1412994)
    at http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1401372
From previous event:
    at ContentWizardPanel.loadDefaultModelsAndUpdatePageModel (http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1401296)
    at http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1402266
From previous event:
    at ContentWizardPanel.createTemplateUpdatedPromise (http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1402226)
    at ContentWizardPanel.handleOtherContentUpdate (http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1400652)
    at updateHandler (http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1396018)
    at http://localhost:8080/admin/tool/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:1551699758/js/bundle.js:1:1397036
    at Array.forEach (<anonymous>)
    at contentUpdatedHandler (http://loca`

PR #272, opened 2019-03-06
Error in console after publishing a shortcut with an image in target #260

edloidas:
Disabled `liveEditModel` update for content without the page editor.

### DIFF
--- a/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -1246,10 +1246,10 @@ export class ContentWizardPanel
         return new ContentWizardDataLoader().loadDefaultModels(site, this.contentType.getContentTypeName()).then(
             defaultModels => {
                 this.defaultModels = defaultModels;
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1246,17 +1246,20 @@ export class ContentWizardPanel
...
-                    const needsReload = !this.isSaving(); // pageModel is updated so we need reload unless we're saving already
-                    if (livePanel) {
-                        livePanel.setModel(this.liveEditModel);
-                        if (needsReload && reloadPage) {
-                            this.debouncedEditorRefresh(true);
-                        }
-                    }
-                    return needsReload;
-                });
+                const livePanel = this.getLivePanel();
```
alansemenov: Why did you move this line out of the `then` handler? It won't be used if `this.liveEditModel` is undefined, so I don't see why we need it here.

edloidas: You are right. Moved it inside the `then` scope.

Issue #48, opened 2019-01-11
Error appears after changing a template in Inspection Panel 


jsi:
@sgauruseu commented on [Sat Dec 29 2018](https://github.com/enonic/app-contentstudio-old/issues/657)

1. Navigate to Content Studio, expand `Features`  and add one new template: 
  select `Site` in supports, select  `default page` in controllers , then insert `Layout component`  and save the template. (Or insert  an image instead of the layout )

2. Open Features in wizard, open `Inspection Panel` and select this new page-template
3. Select  'main region only' template  

![image](https://user-images.githubusercontent.com/3728712/50538835-7c92d980-0b87-11e9-8ca7-42998fed4c0d.png)

Issue 1 : 
Confirmation Dialog did not appear when you select 'main region only'
Issue 2: 
Button `Save` is displayed on the toolbar , but expected  is 'Saved'. 
Alert dialog appears when you try to close the site : 

![image](https://user-images.githubusercontent.com/3728712/50538870-f4f99a80-0b87-11e9-96e7-1fc94db8334e.png)

Issue 3: 
LiveEdit closes and error notification message appears 

![image](https://user-images.githubusercontent.com/3728712/53876466-e44a2280-4017-11e9-811a-2349ad3c854b.png)

PR #279, opened 2019-03-07
Error appears after changing a template in Inspection Panel #48

ashklianko:
-Now switching to template from controller will correctly update Saved button. Issue occurred because of how we treat empty config and regions page objects.


ashklianko:
Also needs https://github.com/enonic/xp/pull/6988

Issue #274, opened 2019-03-07
iframe and script tags should by default not be allowed in HTML Editor


alansemenov:
1. Disallow `iframe` and ``script`` tags in HTML Editor by default. 

2. Read a property from the **app-contentstudio** config file (`com.enonic.app.contentstudio.cfg`) which will allow `script` tag:

`htmlinput.allowScripts=true`

If it's _true_, allow ``script`` element in the editor.

For `iframe` users can use the iframe macro.

3. Document the new property (separate task)

ashklianko:
Where to document this? https://github.com/enonic/xp-doc ?

sigdestad:
Nope.. 
Strictly speaking it should be documented in app-contentstudio

PR #285, opened 2019-03-11
iframe and script tags should by default not be allowed in HTML Edito…

ashklianko:
…r #274

### DIFF
--- a/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
+++ b/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
@@ -811,7 +811,7 @@ class HtmlEditorConfigBuilder {
             removePlugins: this.getPluginsToRemove(),
             removeButtons: this.toolsToExlcude,
             extraPlugins: 'macro,image2,tableresize',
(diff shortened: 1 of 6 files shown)


src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
```diff
@@ -855,6 +855,14 @@ class HtmlEditorConfigBuilder {
         return 'resize,image';
     }
 
+    private getExtraAllowedContent(): string {
+        if (this.editorParams.isScriptAllowed()) {
+            return 'code address dl dt dd blockquote script;img[data-src]';
+        }
+
+        return 'code address dl dt dd blockquote;img[data-src]';
+    }
```
alansemenov: This is a bit lazy :)
We shouldn't duplicate the entire string here. Just create a ``const`` that will be either `` script`` or empty string depending on ``this.editorParams.isScriptAllowed()`` and then use that const inside the string as a template literal:
```
return `code address dl dt dd blockquote${scriptTag};img[data-src]`;
```

ashklianko: He-he, ok :)

Issue #290, opened 2019-03-11
Floating context panel is shifted down whenever modal dialog opens


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/54133915-3d9ac300-4417-11e9-8b8c-cb5f1be9541b.png)

![image](https://user-images.githubusercontent.com/11612490/54133852-1ba14080-4417-11e9-8c02-931f195c7442.png)

PR #294, opened 2019-03-13
Floating context panel is shifted down whenever modal dialog opens #290

edloidas:
* Updated transition to be applied to the sliding (`right`) only.
* Updated position of the floating panel, when the upper container is blurred, since 
  > ... value other than none for the filter property results in the creation of a containing block for absolute and fixed positioned descendants.

  Thus, the position of the floating panel must be adjusted.


alansemenov:
Nice catch with the filter 👍

Issue #275, opened 2019-03-07
Change logic for default values in "Insert Image" dialog's "Caption" and "Alt text" fields 


alansemenov:
Change logic default values in the Caption and Alt Text fields after selecting a new image from the selector.

![image](https://user-images.githubusercontent.com/11612490/53957705-55084200-40df-11e9-9f40-d94211c808a3.png)

**Caption field**: 
Use value from the content's `data.caption` field. If `data.caption` doesn't exist or is empty, use value from `x.media.imageInfo.description`.

**Alt text field**:
Use value from `data.alt_text` field. If `data.alt_text` doesn't exist or is empty, use image file name.

ashklianko:
Pushed into https://github.com/enonic/app-contentstudio/pull/305 because issues closely related

PR #305, opened 2019-03-14
Set default value for the "Caption" field in ImageInspectionPanel #304

ashklianko:
-Have to fetch full image content out of ImageComponent class, otherwise circular dependency occurs during build


ashklianko:
Also pushed https://github.com/enonic/app-contentstudio/issues/275 here because closely related.

alansemenov:
Sweet!

PR #318, opened 2019-03-14
Attachment binary is not deleted after the attachment was removed fro…

vbradnitski:
…m content #310


alansemenov:
I fixed it for the 6.15.6/lib-admin-ui myself

Issue #325, opened 2019-03-18
Context panel auto-opens in the image content wizard


alansemenov:
When opening image content for edit, the context panel always opens automatically. It doesn't for site, folder etc., and it shouldn't depend on content type at all, only on browser width.

Also, there's a problem with overlapping icons in the top right corner.

This is in 800px wide browser window:
 
![image](https://user-images.githubusercontent.com/11612490/54525981-afbe5b00-4975-11e9-9f3f-e5a3a11aec0a.png)

alansemenov:
I mentioned it works correctly for the site, but not quite - the panel won't open for an existing site, but when creating a new site it auto-opens as well (and display name field loses focus).

alansemenov:
This is still not fixed :(

Try this:
1. Open wizard for an image
2. Change any field, like "Caption"
3. Save

The goddamn panel opens after save. Very irritating.

PR #329, opened 2019-03-19
Context panel auto-opens in the image content wizard #325

edloidas:
* Disabled panel auto-expansion on initialization stage.
* Fixed toolbar buttons position update, after Publish controls are added after a delay.

### DIFF
--- a/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
+++ b/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
@@ -439,7 +439,7 @@ export class LiveFormPanel
         return this.pageView;
     }
 
-    setModel(liveEditModel: LiveEditModel) {
+    setModel(liveEditModel: LiveEditModel, showPanel?: boolean) {
 
         this.liveEditModel = liveEditModel;
 
@@ -474,7 +474,7 @@ export class LiveFormPanel
         this.pageModel.unComponentPropertyChangedEvent(this.componentPropertyChangedHandler);
         this.pageModel.onComponentPropertyChangedEvent(this.componentPropertyChangedHandler);
 
-        this.clearSelection();
+        this.clearSelection(showPanel);
 
         this.handleContentUpdatedEvent();
     }
@@ -510,7 +510,7 @@ export class LiveFormPanel
         if (this.pageSkipReload === false && !this.pageLoading) {
 
             if (clearInspection) {
-                this.clearSelection();
+                this.clearSelection(false);
             }
 
             this.pageLoading = true;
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -439,7 +439,7 @@ export class LiveFormPanel
         return this.pageView;
     }
 
-    setModel(liveEditModel: LiveEditModel) {
+    setModel(liveEditModel: LiveEditModel, showPanel?: boolean) {
```
alansemenov: Do we ever call ``setModel()`` with ``showPanel = true``?
I found 1 case where we pass ``false`` and 2 cases where nothing is passed (so it's ``undefined``). So why do we need this parameter at all?
The same goes for ``clearSelection()``, actually - we pass either ``false`` or undefined, never ``true``.

image

edloidas: `InspectEvent` accepts `showPanel` param and set it to `true`, if it is `undefined`. So, setting a model with `undefined` value will result in calling the `InspectEvent` with the default (`true`) value. But the last time (see the screenshot) we explicitly set the `showPanel` param to `false`.

alansemenov: Ok, I'll rephrase - why don't we do the same for ``showPanel`` argument in ``setModel`` and ``clearSelection`` methods in terms of declaring it with default ``true`` value instead of optional? It's very confusing now that ``undefined`` value of the ``showPanel`` argument in fact means ``true``.

edloidas: While the `undefined` (`?`) parameters were never meant to be of specific value, I can now see, that sometimes it can be confusing for the boolean values, when people assume, that `undefined` parameter will be equals to `false` in checks further in the code. I caught myself one time thinking this way.
Okay, I'll suggest that I'll be fixing this in another pull #338 since there were a lot of refactoring, covering some classes from this PR.
As for the `undefined` boolean parameters, let us follow this rule:
* Don't assign the `true` to the `undefined` parameters further in the call stack. `undefined` should be treated like `false` (`!!undefined`) instead:
```diff
function foo(silet?: boolean) {
  bar(silent);
}
function bar(silent?: boolean) {
  baz(silent);
}
- function baz(silent: boolean = true) {
-   if (silent) {
-     // ...
-   }
- }
+ function baz(silent: boolean = false) {
+   if (!silent) {
+     // ...
+   }
+ }
```
* Assigning `true` to boolean parameter is possible, but only when calling the function without the arguments, or with defined arguments:
```diff
function foo(silent: boolean = true) {
  // ...
}

function bar(value?: boolean) {
- foo(value);

+ foo();

+ foo(false);

+ const defined = true;
+ foo(defined);
}
```

Issue #356, opened 2019-03-27
Endless spinner after clicking on Create Issue button


sgauruseu:
1. Do logout (if is  logged in) 
2. Do log in again.
3. Press on `Create Issue` button on the grid-toolbar

**Bug**:
Dialog opens and endless spinner  in the dialog 

![image](https://user-images.githubusercontent.com/3728712/55077965-15af8e80-50aa-11e9-8a4c-bc28d2505d9d.png)

PR #366, opened 2019-03-28
Endless spinner after clicking on Create Issue button #356

ashklianko:
-Moving load mask show/hide related code out of dialogs to their common parent
-Showing modal dialog only when dialog is rendered and shown, otherwise waiting for dialog to be shown


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/972

Issue #390, opened 2019-04-04
Context menu position doesn't change on PageView resize


edloidas:
* When Page Editor changes size, the context menu is not repositioned at all.
* When the arrow changes position on the click, close to the border, it doesn't restore the position afterward, when clicking, for example, in the center of the editor.

PR #395, opened 2019-04-04
Context menu position doesn't change on PageView resize #390

edloidas:


src/main/resources/assets/js/page-editor/ItemViewContextMenu.ts
```diff
@@ -24,6 +35,14 @@ export class ItemViewContextMenu
 
     private orientationListeners: { (orientation: ItemViewContextMenuOrientation): void }[] = [];
 
+    private notClicked: boolean = false;
+
+    private keepOrientation: boolean = false;
+
+    private targetSize: TargetSize;
+
```
alansemenov: Why not use ``private showAtOptions: ShowAtOptions = {notClicked: false, keepOrientation: false};`` instead of three separate properties?

edloidas: While it will take less code to define and assign those options, it will be harder to operate with them further in code. Using options object alongside the `x` `y` values in the `showAt`-like functions are logical and are commonly used in the code, that manipulates with 3D graphics. When we use `Coordinates`, we group values, but with `ShowAtOptions` we are just black boxing them. I can still commit the updated version if it's needed.

Issue #399, opened 2019-04-08
Site wizard - Save button on toolbar gets enabled after version rollback


sgauruseu:
1. Open new site-wizard, type a name and save it.
2. Select Features in app-selector, save it .
3. Add  second application (Superhero), save it.
4. Open Version History widget and do rollback the 2 last versions:

![image](https://user-images.githubusercontent.com/3728712/55710229-d579d580-59f2-11e9-8893-0d225698539e.png)

PR #442, opened 2019-04-17
Site wizard - Save button on toolbar gets enabled after version rollback

ashklianko:
…rsion #399

-Issue with XData not updated correctly after version rollback


alansemenov:
@vbradnitski Please review this

alansemenov:
@vbradnitski Please review the latest changes. Fix of the original bug is tested, seems to work fine.

Issue #394, opened 2019-04-04
Pdf document is not updated in Preview Panel after updating the document in wizard or version rollback


sgauruseu:
**Case 1**
 Open existing pdf-document and upload another pdf-file, close the wizard 
Bug:
Bug: the document is not updated  Preview Panel

**Case 2:**
- select a pdf-document, open Versions Widget and rollback the previos version
**BUG:**
Document is not updated in Preview Panel :  

![image](https://user-images.githubusercontent.com/3728712/55557689-1623d700-56f3-11e9-8d31-f3044c58bcf4.png)

PR #443, opened 2019-04-17
Pdf document is not updated in Preview Panel after updating the docum…

vbradnitski:
…ent in wizard or version rollback #394

### DIFF
--- a/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
+++ b/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
@@ -300,7 +300,14 @@ export class ContentItemPreviewPanel
             if (allows) {
                 this.setPreviewType(PREVIEW_TYPE.MEDIA);
                 if (this.isVisible()) {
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -300,7 +300,14 @@ export class ContentItemPreviewPanel
...
                 if (this.isVisible()) {
-                    this.frame.setSrc(api.util.UriHelper.getRestUri(`content/media/${contentSummary.getId()}?download=false#view=fit`));
+                    const uri = api.util.UriHelper.getRestUri(`content/media/${contentSummary.getId()}?download=false#view=fit`);
+
+                    if(this.frame.getSrc() !== uri) {
+                        this.frame.setSrc(uri);
+                    } else {
+                        this.frame.refresh();
+                    }
+
```
alansemenov: I suggest moving this logic (`if...else...`) inside `setSrc` method of the `IFrame` class, because we need this kind of refresh to happen not only in `ContentItemPreviewPanel` but in general for any iframe, whenever URL has changed. Which means there will be no changes here, but only in `api.dom.IFrameEl`

PR #446, opened 2019-04-19
Exception after unpublishing a deleted content item inside the Wizard…

edloidas:
… #385

* Enabled unpublish for the deleted items.
* Removed the unnecessary promise execution, when there is nothing to update.


edloidas:
Updated commits and the description. The branch is forced-pushed, so remove the local one, before you check out the one from `origin`.

Issue #462, opened 2019-05-09
Missing app info is not updated in the Site config after installation


alansemenov:
**App is missing:**
![image](https://user-images.githubusercontent.com/11612490/57459987-90a4c100-7274-11e9-8c17-6f959fa8ea4c.png)

**App is installed (the box is no longer red, but app icon/name/description are not shown):**
![image](https://user-images.githubusercontent.com/11612490/57460541-98189a00-7275-11e9-9410-1149dea7a035.png)

PR #466, opened 2019-05-13
Missing app info is not updated in the Site config after installation…

ashklianko:
… #462


ashklianko:
> Two issues:
> 
> 1. Icon and description are now correctly updated after the missing app is installed, but not app's display name
> 2. We should do the same (update app info) when the app is started. Try this: stop the app used by the site, open the site for edit, notice the app marked with red and missing the info, start the app.

Done

alansemenov:
Sweet

PR #468, opened 2019-05-13
Document keyboard shortcuts in Content Studio #467

edloidas:


sigdestad:
Hi, great job with docs so far.
Did you see that adoc has this: https://mrhaki.blogspot.com/2015/04/awesome-asciidoctor-display-keyboard.html

edloidas:
@sigdestad Oh, nice. Didn't know about this option. Will try to update the table and compile it to see if it is still readable. Will probably need to use the capital letters.

Issue #480, opened 2019-05-16
Move PropertyHelper class to lib-admin-ui


ashklianko:
@GlennRicaud commented on [Tue May 14 2019](https://github.com/enonic/app-users/issues/234)

The issue https://github.com/enonic/app-users/issues/207 was only halfway done it seems.
I create an ID Provider with the application, save and close -> Looks good
I open and close the ID Provider -> The apps says that there is a difference. And it saves with an emnpty property set...

ashklianko:
Updating contentstudio to use PropertyTreeHelper from lib-admin-ui

PR #481, opened 2019-05-16
ID Provider creation - Empty property set #480

ashklianko:
Updating contentstudio to use PropertyTreeHelper from lib-admin-ui


ashklianko:
https://github.com/enonic/lib-admin-ui/pull/1000 required for it

PR #539, opened 2019-06-13
Refactor ContentTreeGrid #1007

ashklianko:
-Broke all extra large functions into smaller ones
-Removed redundant code wherever possible


ashklianko:
Will refactor more in parallel because still some things could be improved, just takes much time to get into the code.

Issue #551, opened 2019-06-19
Show state icon to the left of the Publish button in the Content Wizard


alansemenov:
See also https://github.com/enonic/app-contentstudio/issues/512

In Content Wizard the icon representing state of the content item will be shown in the toolbar, to the left of the Publish button, and not over the content type icon as in other places. The same goes for the "invalid content" icon as well.

The icon should react to changes of item state and validity.

Center the last modifier/date in the toolbar.

![image](https://user-images.githubusercontent.com/11612490/59767781-54b54080-92a3-11e9-9597-cb3f5bf5e272.png)

![image](https://user-images.githubusercontent.com/11612490/59768943-816a5780-92a5-11e9-9b40-793daa83f3e4.png)

Icons should have the same tooltips as described in #512

PR #577, opened 2019-06-26
Show state icon to the left of the Publish button in the Content Wiza…

ashklianko:
…rd #551


alansemenov:
Some issues with layout of controls inside the toolbar on various widths of browser window. Also too much space to the right of the state icon. To be fixed in later commits.

image

Issue #646, opened 2019-07-15
In Publish dialog warning about items in progress is shown when there are none


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/61227033-cf973d00-a723-11e9-807b-3b80bb2c1dbf.png)

PR #648, opened 2019-07-16
In Publish dialog warning about items in progress is shown when there…

ashklianko:
… are none #646

### DIFF
--- a/src/main/resources/assets/js/app/publish/PublishProcessor.ts
+++ b/src/main/resources/assets/js/app/publish/PublishProcessor.ts
@@ -202,9 +202,15 @@ export class PublishProcessor {
     }
 
     public containsItemsInProgress(): boolean {
-        return this.itemList.getItems().some(
-            (item: ContentSummaryAndCompareStatus) => !item.isOnline() && item.getContentSummary().isInProgress()) ||
-               this.dependantList.getItems().some((item: ContentSummaryAndCompareStatus) => item.getContentSummary().isInProgress());
+        return this.itemList.getItems().some(this.isItemInProgress) || this.dependantList.getItems().some(this.isDependantItemInProgress);
+    }
+
+    private isItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
+        return !item.isOnline() && item.getContentSummary().isValid() && item.getContentSummary().isInProgress();
+    }
+
+    private isDependantItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
+        return item.getContentSummary().isValid() && item.getContentSummary().isInProgress();
     }
 
     public getDependantIds(): ContentId[] {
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/js/app/publish/PublishProcessor.ts
```diff
@@ -202,9 +202,15 @@ export class PublishProcessor {
     }
 
     public containsItemsInProgress(): boolean {
-        return this.itemList.getItems().some(
-            (item: ContentSummaryAndCompareStatus) => !item.isOnline() && item.getContentSummary().isInProgress()) ||
-               this.dependantList.getItems().some((item: ContentSummaryAndCompareStatus) => item.getContentSummary().isInProgress());
+        return this.itemList.getItems().some(this.isItemInProgress) || this.dependantList.getItems().some(this.isDependantItemInProgress);
+    }
+
```
alansemenov: I understand that it looks nicer this way but it feels wrong to have a function called ``isDependantItemInProgress`` which verifies _any_ item, not a _dependant_ item. It's also almost identical to `isItemInProgress`, except for one condition.

So I suggest a combination of the old and the new versions:

````

public containsItemsInProgress(): boolean {
  return this.itemList.getItems().filter(item => !item.isOnline()).some(this.isItemInProgress) ||
    this.dependantList.getItems().some(this.isItemInProgress);
}

private isItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
  const content = item.getContentSummary();
  return content.isValid() && content.isInProgress();
}

````

Comment (ashklianko) [reply src/main/resources/assets/js/app/publish/PublishProcessor.ts]: One more option then:

```
public containsItemsInProgress(): boolean {
        return this.itemList.getItems().some(this.isItemInProgress) || this.dependantList.getItems().some(this.isDependantItemInProgress);
    }

    private isItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
        return !item.isOnline() && this.isDependantItemInProgress(item);
    }

    private isDependantItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
        return item.getContentSummary().isValid() && item.getContentSummary().isInProgress();
    }
```

Comment (ashklianko) [reply src/main/resources/assets/js/app/publish/PublishProcessor.ts]: Also we could adjust functions names to better reflect its purpose

Comment (alansemenov) [reply src/main/resources/assets/js/app/publish/PublishProcessor.ts]: It's the same problem - there's nothing in `isDependantItemInProgress` indicating that the passed `item` is a _dependent_ item. The method checks _any_ item, and this should be reflected in the method name.

Then it should be like this:

```
public containsItemsInProgress(): boolean {
        return this.itemList.getItems().some(this.isOfflineItemInProgress) || this.dependantList.getItems().some(this.isItemInProgress);
}

private isItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
        return item.getContentSummary().isValid() && item.getContentSummary().isInProgress();
}

private isOfflineItemInProgress(item: ContentSummaryAndCompareStatus): boolean {
        return !item.isOnline() && this.isItemInProgress(item);
}

```

Issue #565, opened 2019-06-20
Display issue type icon in the details dialog


alansemenov:
Show icon representing issue type (regular or publish request) over the status button as shown in the screenshots below (should be clickable just as the button):

![image](https://user-images.githubusercontent.com/11612490/60185514-5abe9a80-982a-11e9-9995-d74d528323fd.png)

![image](https://user-images.githubusercontent.com/11612490/60185580-77f36900-982a-11e9-8251-0f578421ca7a.png)

PR #652, opened 2019-07-16
Display issue type icon in the details dialog #565 

edloidas:
* Added issue icon next to the status selection.
* Added small adjustments to the visual style.
* Fixed details dialog info position, when there is no Back button.
* Fixed localization.


src/main/resources/assets/js/app/issue/IssueType.ts
```diff
@@ -13,4 +13,15 @@ export class IssueTypeFormatter {
             return null;
         }
     }
+
```
alansemenov: This was probably done in accordance with `IssueStatus` class (which is also wrong), but...

Both `parseType` and `parseTypeName` are now very fragile - both will break if we add a new issue type. Also, we should always return `STANDARD` as default, to make sure UI won't break if for example someone creates a new issue with unknown issue type via API.

`parseType` could simply be `return IssueType[value] || IssueType.STANDARD;`
`parseTypeName`: `return (IssueType[value] || IssueType[IssueType.STANDARD]).toLowerCase();` (you'll have to change "-" to "_" in expected return value but this shouldn't be a problem

src/main/resources/assets/js/app/issue/view/IssueStatusSelector.ts
```diff
@@ -1,15 +1,20 @@
...
 
 export class IssueStatusSelector
     extends TabMenu {
 
     private static OPTIONS: IssueOption[] = [
-        {value: IssueStatus.OPEN, name: 'Open'},
-        {value: IssueStatus.CLOSED, name: 'Closed'}
+        {value: IssueStatus.OPEN, name: 'open'},
+        {value: IssueStatus.CLOSED, name: 'closed'}
     ];
```
alansemenov: Here we should also drop `OPTIONS` (and `IssueOption` interface) and work directly with `IssueStatus` enum (since we always use all statuses).

alansemenov:
Also, one unrelated thing: when all of the options inside the filter are disabled AND the toggler is disabled (= there are no issues on this tab at all), we should hide both the filter and the toggler (and show a message in the panel, as mentioned in a previous review).

image

src/main/resources/assets/js/app/issue/view/IssueStatusSelector.ts
edloidas: I agree. I completely refactored the IssueStatus usage in IssueStatusSelector and removed the IssueOption.

src/main/resources/assets/js/app/issue/IssueType.ts
edloidas: Yes. I've refactored `IssueTypeFormatter` and `IssueStatusFormatter`, which made some functions simpler. However, `parseType` may return `undefined` instead of the default value. I think it's more obvious if this function will be used elsewhere and some invalid value is passed. The default value is set in the `IssueSummaryBuilder` if the server returns unknown value or `null` instead of type.

PR #679, opened 2019-07-23
Implement Layers widget

ashklianko:


ashklianko:
> Let's change the logic so that **in the base layer** when the item is selected the widget always shows info about the layer, and not about the item. `inherited` property doesn't make sense in the base layer. I'll update the task description.

Updated

Issue #668, opened 2019-07-22
Implement Layers List modal dialog


alansemenov:
Modal dialog with the list of layers can be opened either from the "Layers" widget (when there's only base layer) or from the Layer details dialog (when "back" icon is clicked).

1) The list should be tree-structured
2) The base layer cannot be deleted (disable the X icon)
3) A layer that has child layers cannot be deleted (disable the X icon)
4) Confirm deletion by asking user to type in ID of the layer to be deleted (see below).
5) "Pencil" icon will open the Layer details dialog
6) "+" icon in the header will open the New layer dialog

![image](https://user-images.githubusercontent.com/11612490/61629446-28675800-ac85-11e9-9861-a4d9a9202862.png)

![image](https://user-images.githubusercontent.com/11612490/61629455-30bf9300-ac85-11e9-96d0-a4fa03a3047e.png)

![image](https://user-images.githubusercontent.com/11612490/61629855-5e590c00-ac86-11e9-9da4-67129c34c0a4.png)

PR #687, opened 2019-07-25
Implement Layers List modal dialog #668

ashklianko:
NB: Added test data for dialog list for presentation purposes, this list to be removed after 'Create' dialog implemented


ashklianko:
Sure!

Issue #671, opened 2019-07-22
Implement Layer selector


alansemenov:
When there are other layers than _base_, we'll show a layer selector in the Content Studio header for switching between the layers.

![image](https://user-images.githubusercontent.com/11612490/61631430-a24e1000-ac8a-11e9-9c16-6f9a0da42a4b.png)

1. The list of layers is tree-structured (same as in the modal dialog with the list of layers)
2. Current layer is displayed in the header and highlighted with blue background in the selector dropdown
3. The list shows display name + id in brackets, description and icon (or default icomoon icon if none set)

Selecting a layer should change the context of the app (endpoint base in `ContentResourceRequest` will change to go via selected layer) and reload the entire grid:

``this.resourcePath = Path.fromParent(super.getRestPath(), `cms/default/${layerId}` );``

PR #730, opened 2019-08-06
Implement Layer selector #671

ashklianko:


ashklianko:
Few things to implement later:
-Listen for server layer events (Create/Update/Delete)
-Update Layer Selector on layer server events
-Show Layer Selector when 2nd layer added / Hide when only base layer left

Currently all changes for layers reflected in Layer Selector after reload

Issue #739, opened 2019-08-07
Pass current layer via URL in Content Studio


alansemenov:
We should pass layer Id via URL in Content Grid and Content Wizard

Grid:
`/main#/<layer>/browse`

Wizard:
`/main#/<layer>/edit`

That's where current layer should be retrieved from to be preset in the Content Studio. Things to test:
1) If `<layer>` from URL is not found, default to "base"
2) Layer selector should be preset based on `<layer>`
3) Layers widget should display correct info about the layer
4) URL should change after current layer is changed via selector

Issue #746, opened 2019-08-08
Adapt XData resource request to fetch from required layer


ashklianko:
Currently rendering wizard panel for non base content fails because XData is not adapted for layers.

Also backend task is required as  XDataResource.java is also not awared of layers.

PR #749, opened 2019-08-08
Pass current layer via URL in Content Studio #739

ashklianko:


ashklianko:
Note that opening Content Wizard for non base content layer won't fork until  https://github.com/enonic/app-contentstudio/issues/746 is done.

Issue #757, opened 2019-08-08
Numbers on tabs in the Issue List dialog shouldn't include inactive issues


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/62713001-47434980-b9fc-11e9-8348-1c88d77fc898.png)

![image](https://user-images.githubusercontent.com/11612490/62713015-4ca09400-b9fc-11e9-8d00-06174ebc748c.png)

PR #770, opened 2019-08-12
Numbers on tabs in the Issue List dialog shouldn't include inactive i…

edloidas:
…ssues #757

Updated tab labels to display only the active issues maximum count.

### DIFF
--- a/src/main/resources/assets/js/app/issue/view/IssueListDialog.ts
+++ b/src/main/resources/assets/js/app/issue/view/IssueListDialog.ts
@@ -71,19 +71,22 @@ export class IssueListDialog
 
         this.allPanel = this.createIssuePanel({
             storage,
-            noIssuesMessage: i18n('dialog.issue.noIssuesAndPublishRequests')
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/js/app/issue/view/IssueListDialog.ts
```diff
@@ -288,8 +282,32 @@ export class IssueListDialog
...
+
+    private updateIssuesTabLabel() {
+        const total = this.issuesPanel.getActiveTotal();
+        const label = IssueListDialog.createTabLabel(i18n('field.issues'), total);
+        this.issuesTab.setLabel(label);
+    }
+
+    private static createTabLabel(label: string, count: number) {
+        return count > 0 ? (label + ' (' + count + ')') : label;
     }
```
alansemenov: `${label} (${count})` maybe?

edloidas: Yeah, using template string will be better. Fixed.

Issue #787, opened 2019-08-14
Content should remain in state "Ready" after page component(s) were saved as fragment


sgauruseu:
1. Open Features(Ready for Publishing)
2. Open Page Component View and click on the text component and select `Save as Fragment`

BUG: The site gets Work in Progress after saving the fragment.
![image](https://user-images.githubusercontent.com/3728712/63029154-ba473700-beb8-11e9-9371-030dbbc07743.png)

alansemenov:
1. If some other changes are made to the content (without save) and then a fragment is created, then the content should be "In progress", because content itself was changed. Now it will always mark as ready whenever a fragment is created. Basically, we have to check content state at the moment a fragment is created. If the content is already in progress, then don't mark as ready.

2. When a fragment is disassembled (with "Detach from fragment" action in the context menu), there should be the same logic as when creating a fragment. Now content will always go to "In progress".

PR #844, opened 2019-08-27
Content should remain in state "Ready" after page component(s) were s…

edloidas:
…aved as fragment #787

Updated checks for the changes in the content, preventing setting the status to Ready if something, except fragment, is updated.
Updated unpublish action title in the wizard.

### DIFF
--- a/src/main/resources/assets/js/app/page/Page.ts
+++ b/src/main/resources/assets/js/app/page/Page.ts
@@ -16,6 +16,8 @@ import {LayoutComponentType} from './region/LayoutComponentType';
 import {ComponentTypeWrapperJson} from './region/ComponentTypeWrapperJson';
 import {ComponentFactory} from './region/ComponentFactory';
 import {PageJson} from './PageJson';
+import {ComponentPath} from './region/ComponentPath';
+import {ComponentType} from './region/ComponentType';
 
 export class Page
     implements api.Equitable, api.Cloneable {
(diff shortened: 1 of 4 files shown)


src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -698,6 +698,11 @@ export class LiveFormPanel
         this.liveEditPageProxy.onComponentDetached((event: ComponentDetachedFromFragmentEvent) => {
             api.notify.showSuccess(i18n('notify.component.detached', event.getComponentView().getName()));
 
+            const componentPath = event.getComponentView().getComponentPath();
+            const canMarkContentAsReady = this.canMarkContentAsReady(componentPath);
+            if (canMarkContentAsReady) {
+                this.contentWizardPanel.setIsMarkedAsReady(true);
+            }
             this.saveAndReloadOnlyComponent(event.getComponentView());
         });
```
alansemenov: Let's extract the duplicated code (marked) into a separate method

image

src/main/resources/assets/js/app/page/Page.ts
```diff
@@ -163,6 +165,60 @@ export class Page
             });
         });
     }
+
+    findComponentById(id: ContentId, regions?: Region[]): Component {
```
alansemenov: This method is not used

src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -745,6 +748,45 @@ export class LiveFormPanel
...
     }
 
+    private canMarkContentAsReady(componentPath: ComponentPath): boolean {
+        const isReady = this.content.isReady();
+
```
alansemenov: Let's exit the method if `isReady` is false, instead of wrapping the entire body inside `if...`

```
private canMarkContentAsReady(componentPath: ComponentPath): boolean {
  if (!this.content.isReady()) {
    return false;
  }
  const persistedContent = this.contentWizardPanel.getPersistedItem();
     ...
  return viewedPage.equals(serverPage);
}
```

Issue #857, opened 2019-08-28
Refresh the Layer Widget and set readonly mode after deleting a local copy from inside the Wizard


alansemenov:
1. Open an inherited content in the Wizard
2. Say "yes" when asked about creating a local copy
3. Open the Layers widget, it says that the item is local - correct.
4. Click "Delete" button to remove the local copy.

After the copy is deleted, we need to:
a) refresh the widget so that it says that the item is now inherited
b) switch the editor to readonly mode (with notification)

PR #910, opened 2019-09-05
Refresh the Layer Widget and set readonly mode after deleting a local…

ashklianko:
… copy from inside the Wizard #857

-Updated Branch enum to use strings
-Updated ContentServerEventsHandler to be aware of layers and its' branch naming 'draft-<layerName>' (ContentDeletedEvent was not triggered because of that when deleting non-base layer items)
-In ContentWizardPanel trying to handle all possible variations with Deleted event: same content might be deleted on any layer


ashklianko:
Also commited https://github.com/enonic/app-contentstudio/issues/911 in here because depends on code changes made in this PR.

PR #926, opened 2019-09-06
ContentTypeFilter is not aware of layers #924

ashklianko:


ashklianko:
Depends on https://github.com/enonic/xp/pull/7369

Issue #875, opened 2019-08-30
Fetch component descriptors for all applications with one request


runarmyklebust:
When creating a new content, there are a lot of repeated request triggered.
This is an example from helsenett:

![Screen Shot 2019-08-30 at 11 00 57](https://user-images.githubusercontent.com/356026/64007999-848b8a80-cb15-11e9-8222-4e9786b9f85f.png)

There are 173 request on the initial opening.

pmi:
Okay, there is indeed a lot of requests with the same `applicationKey` and `by_application` endpoint, but they are actually different even though the last bit is the same.

Following requests are made per each application (notice the full path): 

- http://localhost:8080/admin/rest/content/page/descriptor/list/by_application ( this one's done 2 times per app, first by page inspection panel layout(), and then by live page editor in a separate iframe )
- http://localhost:8080/admin/rest/content/page/part/descriptor/list/by_application ( on part inspection panel layout() )
- http://localhost:8080/admin/rest/content/page/layout/descriptor/list/by_application ( on layout inspection panel layout() )

That resulted in 24 requests with 6 apps added to site.

sigdestad:
Could we handle this so that number of requests dont increase with the number of apps?

pmi:
@sigdestad We can reduce requests number to 4 sharp If we pass all application keys as array.

alansemenov:
We should also not fetch descriptors (page/part/layout) until user actually needs them. Created a [separate issue](https://github.com/enonic/app-contentstudio/issues/918) to fix that.

PR #942, opened 2019-09-10
Fetch component descriptors for all applications with one request #875

pmi:
- adapted ApplicationBasedCache to work with multiple app keys and cache requests with empty responses as well
- changed GetDescriptorsByApplication*S* requests for Page/Layout/Part to make requests to corresponding endpoints instead of doing multiple GetDescriptorsByApplication requests
- also debounced descriptors reload on applications change that reduced No of requests in case of selecting multiple applications with checkboxes


pmi:
Or should @runarmyklebust review it maybe ?

Issue #986, opened 2019-09-18
HTML Area - Implement toggle button for selecting default paste mode


alansemenov:
By default Cmd-V pastes text with formatting, and Cmd-Shift-V without. 
Default state of the button will be "off" and the behaviour will be as above.
In "on" state the behaviour will change the other way around - Cmd-V will paste text without formatting, and Cmd-Shift-V with.
We don't have to remembe the state of the button, it will always be "off" by default providing current behaviour.

PR #1005, opened 2019-09-22
HTML Area - Implement toggle button for selecting default paste mode …

ashklianko:
…#986

-Using ckeditor 'pastetext' plugin's icon and it's command to paste plain text when paste mode changed; 'pastetext' plugin itself does very close thing to what we need but (see https://ckeditor.com/cke4/addon/pastetext) but might be rather switched on constantly during editor's init process or might be used via toolbar button that doesn't really work
-Renamed macro plugin file correctly


ashklianko:
One more question: is it OK that toolbar is now on 2 lines?

Issue #1012, opened 2019-09-24
Enable caption for table elements


ashklianko:
@alansemenov commented on [Wed Sep 18 2019](https://github.com/enonic/app-users/issues/249)

Add "_Caption_" field to our table properties dialog. When filled, it should add `<caption></caption>` element inside the table. It's a part of the CKEditor's native dialog and we should have it as well.

![image](https://user-images.githubusercontent.com/11612490/65154767-5d491e80-da2c-11e9-8a86-cd3e43326986.png)

PR #1014, opened 2019-09-24
Enable caption for table elements #1012

ashklianko:


ashklianko:
Works for me, maybe some special case?

ashklianko:
Merged master into, please try now

Issue #990, opened 2019-09-18
Set cursor to the same position when switching HTML Editor to Full Screen mode


alansemenov:
After switching the editor to the Full Screen mode set cursor to the same position where it was before the switch.

PR #1035, opened 2019-09-26
Set cursor to the same position when switching HTML Editor to Full Sc…

ashklianko:
…reen mode #990

-Setting cursor to a selected element; unfortunately can't set it to exact position


alansemenov:
@ashklianko can you review my changes? For some reason I cannot set you as a reviewer if you were an author of the original PR.

ashklianko:
> @ashklianko can you review my changes? For some reason I cannot set you as a reviewer if you were an author of the original PR.
Sure

ashklianko:
Works like a charm

PR #1045, opened 2019-09-27
fix for NPE, owner might be null

ashklianko:


src/main/resources/assets/js/app/wizard/UpdatePersistedContentRoutine.ts
```diff
@@ -90,7 +91,7 @@ export class UpdatePersistedContentRoutine
     private hasContentChanged(persisted: Content, viewed: Content): boolean {
         return !persisted.dataEquals(viewed.getContentData()) ||
                !persisted.extraDataEquals(viewed.getAllExtraData()) ||
-               !persisted.getOwner().equals(viewed.getOwner()) ||
+               !ObjectHelper.equals(persisted.getOwner(), viewed.getOwner()) ||
```
alansemenov: Why not the same for other properties? Any of them might not be set.

Issue #1021, opened 2019-09-25
Inspect Panel - incorrect selected option after changing template selection


sgauruseu:
Case 1:
1. Open Features, open Page Component View, click on `main region ` an open Inspect panel.

![image](https://user-images.githubusercontent.com/3728712/65594026-84f33600-df9a-11e9-9223-7b86c0ce5c79.png)

2.  Select `All input types` option in Inspect panel:

![image](https://user-images.githubusercontent.com/3728712/65594099-bf5cd300-df9a-11e9-87be-7d50a2c44c57.png)

3. Select  `Attachment functions example` option in the selector.

BUG:  Incorrect selected option in the  PageTemplateAndController Selector
Actual  - `Automatic  
Expected: `Attachment functions example`

![image](https://user-images.githubusercontent.com/3728712/65594516-a4d72980-df9b-11e9-8276-c278b3bdc39e.png)

Case 2:
1. Open new site-wizard.
2. Select Features.
3. Select 'main region' controller
4. Open Inspect panel and expand options in the template-selector:
BUG-   two `Automatic`  are present in options

![image](https://user-images.githubusercontent.com/3728712/65754501-dbd34980-e119-11e9-8ba6-da0b208ec39a.png)

PR #1061, opened 2019-10-03
Inspect Panel - incorrect selected option after changing template sel…

ashklianko:
…ection #1021

-Resetting config after controller changed
-Correctly reloading options


alansemenov:
Works fine now 👍

Issue #1069, opened 2019-10-07
Give a different message when a version was not reverted


anatol-sialitski:
To test this:

Open Content Studio
Edit some content, save changes
Revert to previous version in the Version History widget. You'll get a message saying "Version was changed to ..."
Now select the same version as the last time and click "Revert" again. No new version will be created because that version is identical to the current one, but the message is still "Version was changed to ..." which is confusing since nothing has changed.
We should give a different message in this case: "No changes to revert". Ideally we shouldn't reload the list of versions in the widget in this case either.

PR #1070, opened 2019-10-07
Give a different message when a version was not reverted #1069

anatol-sialitski:
implemented

Note: added the `notify.revert.no.changes=No changes to revert.` to i18n files with default text.

This PR depends on https://github.com/enonic/xp/pull/7509


src/main/resources/assets/js/app/view/context/widget/version/VersionsView.ts
```diff
@@ -199,9 +199,13 @@ export class VersionsView
                 .onExecuted((action: api.ui.Action) => {
                     if (!isActive) {
                         new RevertVersionRequest(item.id, this.getContentId().toString()).sendAndParse().then(
-                            (contentId: ContentId) => {
-                                api.notify.NotifyManager.get().showFeedback(i18n('notify.version.changed', item.id));
-                                new ActiveContentVersionSetEvent(this.getContentId(), item.id).fire();
+                            (contentVersionId: string) => {
+                                if (contentVersionId === this.activeVersion.id) {
+                                    api.notify.NotifyManager.get().showFeedback(i18n('notify.revert.no.changes'));
```
alansemenov: * Change the key _notify.revert.no.changes_ to _notify.revert.noChanges_
* Remove this phrase from all phrase resource files, except for `phrases.properties` (but change the key there) and you can also translate it to russian in _phrases_ru.properties_. You should never add english text into non-english resource files, because then these new phrases will not be fetched when we will be translating texts before the next release.

Issue #1093, opened 2019-10-10
Multiple issues when pasting an image into HTML Editor


alansemenov:
Open [this page ](https://enonic.com/blog/best-practices-for-content-management) and copy-paste the top image **with** some text underneath into HTML Editor

1. Left-click the image
Console error:
![image](https://user-images.githubusercontent.com/11612490/66596293-e9b6af00-eb9c-11e9-8e5a-d92bdb7442d7.png)

2. Double-click the image
Native dialog for the image opens:
![image](https://user-images.githubusercontent.com/11612490/66596589-73ff1300-eb9d-11e9-8c5e-dca75e896144.png)

PR #1120, opened 2019-10-22
Multiple issues when pasting an image into HTML Editor #1093

ashklianko:
-Adjusted image's plugin to detect which elements to be upcasted to an image widget (excluded standalone img tags)
-Removed one of two code modifications of image plugin


ashklianko:
Will update

alansemenov:
The only problem now is that if I add a link to an external image, the link button in the toolbar won't be highlighted if I set a cursor on the image. Minor issue though.

Issue #1103, opened 2019-10-15
Selection is not reset in in the Inspect panel's descriptor selector


sgauruseu:
1. Open Features and open Page component View.
2. Insert `layout component`, then select ` 3-col layout`, save the site.
3. Click on the inserted layout in Page Component View, then click on `Remove` menu item, save the site.
4. Insert `layout component` again .
5. Click on the dropdown handle in the layouts-selector in Page Inspection panel.

BUG:  `3-col` layout remains 'selected' in the expanded options:
![image](https://user-images.githubusercontent.com/3728712/66826522-4b886780-ef55-11e9-87e7-82f737ad29f5.png)

**The same issue in part-selector**

PR #1144, opened 2019-10-31
Selection is not reset in in the Inspect panel's descriptor selector …

ashklianko:
…#1103

-Grid is used for holding options in inspection panel selector, 'selected' and 'active' cell classes were not set correctly when switching between components of same kind


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1135

Issue #1151, opened 2019-11-05
Site Wizard - controller selector does not appear after selecting an application in the saved site


sgauruseu:
1. Open new site wizard,type a name.
2. Press `Save` button.
3. Select Features in app-selector.

**Expected**: page descriptor dropdown should appear after selecting the applications

**Actual** :  Application is selected but  "No page controller found" is displayed in Page Editor

![image](https://user-images.githubusercontent.com/3728712/68214836-3ef4ad80-ffef-11e9-94f1-7ba996cbf38f.png)

 **Page descriptor dropdown  appears after clicking on  button Save**

alansemenov:
Note: it works as it should if the content is not saved before selecting the app.

PR #1178, opened 2019-11-12
Site Wizard - controller selector does not appear after selecting an …

ashklianko:
…application in the saved site #1151

-Site model and wizard are listening for data updates of the same content object, however after saving changes site model is getting updated with new content object, but wizard keeps old one, thus updates in wizard are not reaching site model. Fixed by triggering wizard step forms update after wizard saved and live form updated

### DIFF
--- a/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -530,12 +530,16 @@ export class ContentWizardPanel
         this.isFirstUpdateAndRenameEventSkiped = false;
         new BeforeContentSavedEvent().fire();
         return super.saveChanges().then((content: Content) => {
-
+            let isUpdateWizardStepFormsNeeded: boolean = false;
             const persistedItem = content.clone();
             if (liveFormPanel) {
                 this.liveEditModel.setContent(persistedItem);
                 if (this.reloadPageEditorOnSave && this.pageEditorUpdatedDuringSave) {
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -530,12 +530,16 @@ export class ContentWizardPanel
...
         new BeforeContentSavedEvent().fire();
         return super.saveChanges().then((content: Content) => {
-
+            let isUpdateWizardStepFormsNeeded: boolean = false;
             const persistedItem = content.clone();
             if (liveFormPanel) {
                 this.liveEditModel.setContent(persistedItem);
                 if (this.reloadPageEditorOnSave && this.pageEditorUpdatedDuringSave) {
                     this.updateLiveForm(persistedItem);
+
```
alansemenov: I might be wrong, but I don't think we need an extra boolean variable here... Based on the logic, it should be enough with this:

```
if (!this.wizardFormUpdatedDuringSave && persistedItem.isSite()) {
  this.updateWizardStepForms(persistedItem, false);
}
```

I also noticed that `this.wizardFormUpdatedDuringSave` might be _undefined_ (f.ex. when content is saved after page controller selected). We should try to avoid ambiguity for a boolean variable.

ashklianko: Agree, site presumes that content is renderable by default and live form will present on a page

PR #1228, opened 2019-11-22
Incorrect confirmation behaviour in site config with an item-set #1226

ashklianko:
-Masking SiteConfiguratorDialog when any dialog is being opened while SiteConfiguratorDialog is open, that will make modal dialog focus handlers work correctly and solve described issues
-Updated dialogs to use isOpen() instead of isVisible()


ashklianko:
Depends on https://github.com/enonic/lib-admin-ui/pull/1156

PR #1244, opened 2019-11-27
Lokalise: Translations update

jsi:
This pull request was initiated by Lokalise (user Jørgen Sivesind) at 2019-11-27 13:20:07


jsi:
Just testing lokalise.  This is not good enough.

Issue #1080, opened 2019-10-09
Version widget - Incorrect state is displayed on the widget after the published content has been Marked As Deleted 


sgauruseu:
1. Select a folder and publish it. Open Versions widget
2. Open Delete Content and click on Mark As Deleted menu item.

Bug: "Published" status is displayed on the widget.
Expected :  Delete status should be displayed. 

![image](https://user-images.githubusercontent.com/3728712/66464050-f5b04d00-ea86-11e9-8767-a2acc9b1b253.png)

PR #1260, opened 2019-11-29
Version widget - Incorrect state is displayed on the widget after the…

ashklianko:
… published content has been Marked As Deleted #1080

After recent backend changes:
1. Draft and master branches might refer to the same content version
2. Marking content as pending delete only updates state property in master branch object, new content version is not created

Thus now we might have contents with different types of pending delete items, where one content has pending delete represented in separate content version and another one uses master branch version that is marked as PENDING_DELETE.

To represent it correctly in version history as a separate entry:
1. Added special type of versions list item - ContentVersionListItem - that will hold ContentVersion and workspace(branch) props, thus single ContentVersion might produce several ContentVersionListItem for separate workspaces (branches) when it is needed
2. For pending delete contents that doesn't have separate version for pending_delete state creating additional version list item entry


ashklianko:
https://github.com/enonic/app-contentstudio/issues/1261

ashklianko:
Updated

Issue #1267, opened 2019-12-02
Fix layout of the issue list dialog in mobile resolution


alansemenov:
![image](https://user-images.githubusercontent.com/11612490/69960188-2651b780-1509-11ea-9590-c1d28cbf3311.png)

1) Hide "Filter" text (only on low resolution)
2) Smaller font inside the dropdown
3) Smaller font on the two tabs so that they fit
4) More padding

PR #1270, opened 2019-12-02
Fix layout of the issue list dialog in mobile resolution #1267

ashklianko:


ashklianko:
Updated it to look like this:
photo_2019-12-02_22-55-53

Issue #1285, opened 2019-12-04
Changing content permissions triggers fetch of all modified child content items


GlennRicaud:
Calls to resolveByIds should be batched (But check first it is really necessary to retrieve all of them).

Example: Change permissions on the parent of a lot of contents
ContentStudio will call "resolveByIds" with the list of all modified contents

<img width="910" alt="Screenshot 2019-12-04 at 16 14 34" src="https://user-images.githubusercontent.com/12030113/70154999-c21e2780-16b1-11ea-8fae-1eabf039b758.png">

Fix also in 2.2.4

Check for similar situations

sigdestad:
We also need a circut breaker for this stuff in XP! Can never trust a client!

GlennRicaud:
> We also need a circut breaker for this stuff in XP! Can never trust a client!

https://github.com/enonic/app-contentstudio/issues/1288

GlennRicaud:
> We also need a circut breaker for this stuff in XP! Can never trust a client!

https://github.com/enonic/xp/issues/7693

runarmyklebust:
This will be always be a problem if running with several content studios open and someone doing changes to a lot of content?

ashklianko:
Considering UI part of task: I'll optimize those calls on UI so we won't fetch all ids, only needed ones

PR #1296, opened 2019-12-06
Batched resolvedByIds #1285

ashklianko:
-Won't fetch all content summaries after permissions updated, only those that are required


ashklianko:
@edloidas please check that app builds ok, couldn't check next locally

sigdestad:
Sweet.. Do we get this patch in to CS 2.2. branch as well?

Issue #994, opened 2019-09-19
Incorrect behaviour when uploading more attachments than allowed by content type


sgauruseu:
1. Open new  attachment content (min 2, max 4).
2. Select multiple files (5-7) in the uploader:

![image](https://user-images.githubusercontent.com/3728712/65236693-cdfe4280-dae1-11e9-9c7a-5335497b8a32.png)

**BUG**: 
1. Notification message appeared 7 times : "file uploaded" But actually - 4 files were uploaded.
2. Incorrect validation - red icon is displayed in the grid and details panel.
3. Items are displayed in the  `Attachments Widget`
 
![image](https://user-images.githubusercontent.com/3728712/65236007-58de3d80-dae0-11e9-80e1-a87465fe65e5.png)

![image](https://user-images.githubusercontent.com/3728712/65237027-7b715600-dae2-11e9-8eae-06d6399315e6.png)

![image](https://user-images.githubusercontent.com/3728712/65237156-bc696a80-dae2-11e9-87fc-086c1223f6f1.png)

PR #1311, opened 2019-12-10
Incorrect behaviour when uploading more attachments than allowed by c…

ashklianko:
…ontent type #994

-Input types can limit occurrences of items, however uploader is not checking how many items we're uploading and we might upload more items than input allows
-Added mechanism for uploader that checks on demand if item to be uploaded (have to return 'false' from FineUploader's sumbitCallback to reject)
-Adjusted CS files to correspond to latest uploaderEL changes


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1176

Issue #1316, opened 2019-12-11
Replace Content Tree icon with burger icon


alansemenov:
* Replace the tree icon in the top left corner with the burger icon, same as in the Launcher panel, with the same transition on click.

* Replace "CONTENT STUDIO" title to the right of the icon with hardcoded (for now) "Project: Default"

* Keep the tree icon in the Content Wizard toolbar

![image](https://user-images.githubusercontent.com/11612490/70609764-a3b0b280-1c02-11ea-8c9b-2b6e22790f62.png)

Issue #1318, opened 2019-12-11
Different layout of the menu panel in mobile mode


alansemenov:
* Use the same width as the Launcher panel 
* App name should be shown to the right of the "X" icon
* Menu item titles should be shown next to the icons
* A tap anywhere outside of the menu should collapse it

![image](https://user-images.githubusercontent.com/11612490/70610834-8da3f180-1c04-11ea-8eea-3b39a36221d9.png)

ashklianko:
Pushed to https://github.com/enonic/app-contentstudio/pull/1320

PR #1320, opened 2019-12-11
Replace Content Tree icon with burger icon #1316

ashklianko:


ashklianko:
Also added #1318, but without click handler (will add later), so ui could be evaluated

ashklianko:
Updated

Issue #1331, opened 2019-12-13
Version Comparison


alansemenov:
[_epic-version-comparison_](https://github.com/enonic/app-contentstudio/tree/epic-version-comparison)

![image](https://user-images.githubusercontent.com/11612490/70890704-f9f86980-1fe5-11ea-8107-5214331434cb.png)

![image](https://user-images.githubusercontent.com/11612490/70890718-ffee4a80-1fe5-11ea-9cd3-049953aa2653.png)

![image](https://user-images.githubusercontent.com/11612490/70890725-07155880-1fe6-11ea-8158-249a2589d569.png)

PR #1405, opened 2020-01-10
Version Comparison #1331

pmi:


sigdestad:
For the mobile version, maybe we should place older/newer to the left, and the dots to the right?

pmi:
@alansemenov regarding point 6, I personally like @sigdestad suggestion better because placing labels on top will ruin the neat design of stacked combos.

alansemenov:
> @alansemenov regarding point 6, I personally like @sigdestad suggestion better because placing labels on top will ruin the neat design of stacked combos.

@pmi fine by me, as long as you set fixed width on the containers and manage text overflow nicely. Remember about localisation - these captions might get long...

Issue #1428, opened 2020-01-21
Apply button won't save changes to part config after the content is published or marked as ready


poi33:
Community Ref: https://enonic-community.slack.com/archives/C73JLM8E7/p1579609641021400

Reported in 7.1.2 and 7.2.0
Content studio version 2.3.0 

Pressing the apply button does not update the data in the right column, and it goes back to previus data. 
Work around: refresh page and pressing apply again. 

To reproduce:
1. Select any part on a page that has config
2. Mark the content as ready or publish
3. Click the part that has the config
4. Make changes
5. Press Apply button

The panel is refreshed, changes are not saved.

PR #1441, opened 2020-01-27
Apply button won't save changes to part config after the content is p…

ashklianko:
…ublished or marked as ready #1428

-Issue with comparing components


ashklianko:
Also fixes https://discuss.enonic.com/t/replacing-files-in-attachmentuploader/2103

Issue #1446, opened 2020-01-29
"Source" icon is not available for HTML Area inside item-set in part config


alansemenov:
https://discuss.enonic.com/t/source-action-is-not-available-in-htmlarea-when-placed-inside-item-set/2106/2

Part config:
```
<part>
    <display-name>Team contact</display-name>
    <description>Team contact</description>
    <form>
        <input name="teamContactTitle" type="TextLine">
            <label>Title</label>
            <occurrences minimum="1" maximum="1"/>
            <default>Team contact title</default>
        </input>

        <input name="preface" type="HtmlArea">
            <label>Text</label>
            <occurrences minimum="1" maximum="1"/>
            <default>This is giving me a Source option</default>
            <config>
                <!-- <exclude>*</exclude>
                <include>AlignLeft AlignRight | Bold Italic</include> -->
            </config>
        </input>

        <item-set name="columns">
            <label>New team member</label>
            <help-text>Use this to add multiple team member's</help-text>
            <occurrences minimum="1" maximum="0"/>
            <items>

                <input name="contactName" type="TextLine">
                    <label>Name</label>
                    <occurrences minimum="1" maximum="1"/>
                    <default>Team member name</default>
                </input>

                <input name="aboutContact" type="HtmlArea">
                    <label>This is not giving me a Source option</label>
                    <occurrences minimum="1" maximum="1"/>
                    <default>team.member@gjensidige.no</default>
                </input>

                <mixin name="image"></mixin>
            </items>

        </item-set>
    </form>
</part>

```

sgauruseu:
Verified : 
![image](https://user-images.githubusercontent.com/3728712/81271270-9209f480-9054-11ea-9a94-f74fbb73e642.png)

PR #1449, opened 2020-01-29
"Source" icon is not available for HTML Area inside item-set in part …

ashklianko:
…config #1446

-Adding editor to a list of managed editors only after it is fully initialized, otherwise it might be triggered to reinitialize (while initializing itself) during parent element rendering

### DIFF
--- a/src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
+++ b/src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
@@ -110,17 +110,11 @@ export class HtmlArea
         const textAreaWrapper = new DivEl();
 
         textAreaEl.onRendered(() => {
-            if (this.authRequest.isFulfilled()) {
+            this.authRequest.then(() => {
                 this.initEditor(editorId, property, textAreaWrapper).then(() => {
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
```diff
@@ -109,14 +109,16 @@ export class HtmlArea
 
         const textAreaWrapper = new DivEl();
 
-        this.editors.push({id: editorId, textAreaWrapper, textAreaEl, property, hasStickyToolbar: false});
-
         textAreaEl.onRendered(() => {
             if (this.authRequest.isFulfilled()) {
-                this.initEditor(editorId, property, textAreaWrapper);
+                this.initEditor(editorId, property, textAreaWrapper).then(() => {
```
alansemenov:
```
this.initEditor(editorId, property, textAreaWrapper).then(() => {
  this.editors.push({id: editorId, textAreaWrapper, textAreaEl, property, hasStickyToolbar: false});
});
```

This should be extracted into a separate method to avoid code duplication. Actually, do we need a check for `if (this.authRequest.isFulfilled())` at all? If a promise is already resolved, `then` handler will be executed right away.

ashklianko:
Updated

Issue #941, opened 2019-09-10
Incorrect status in version history for content with scheduled publishing


jsi:
After unpublishing a content, the scheduling it for republishing at a later time, the status in version history is out of synch with the actual status of the content.  It shows "Published", when it is actually "Published (pending)":
![Screenshot 2019-09-10 14 34 28](https://user-images.githubusercontent.com/477453/64615983-db2d7a00-d3db-11e9-9067-8833088191ff.png)

The same problem is there if I actually edit the content before republishing it:
![Screenshot 2019-09-10 14 54 31](https://user-images.githubusercontent.com/477453/64616021-ee404a00-d3db-11e9-9916-36fe1ae01dce.png)

In both these situations, the version history should show a yellow box with "Published (Pending)" as the status text.

PR #1491, opened 2020-02-10
Incorrect status in version history when republishing content. #941

ashklianko:
-Updated status Published (Pending) to be displayed with other color and right caption

image

### DIFF
--- a/src/main/resources/assets/js/app/versioning/Branch.ts
+++ b/src/main/resources/assets/js/app/versioning/Branch.ts
@@ -1,4 +1,4 @@
 export enum Branch {
-    DRAFT,
-    MASTER
+    DRAFT = 'draft',
+    MASTER = 'master'
 }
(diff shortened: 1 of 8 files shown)


src/main/resources/assets/js/app/versioning/Branch.ts
```diff
@@ -1,4 +1,4 @@
 export enum Branch {
-    DRAFT,
-    MASTER
+    DRAFT = 'draft',
+    MASTER = 'master'
```
alansemenov: It started crashing quite badly for me after this change (`Branch[branch] `returns `undefined` in `UriHelper.getPortalUri()`). Whenever I select an item in the Grid there's an endless spinner in the Preview panel.

src/main/resources/assets/js/app/view/context/widget/version/VersionsView.ts
```diff
@@ -88,136 +95,217 @@ export class VersionsView
...
+    private getStatusClass(): string {
+        if (!this.contentVersion.isInMaster()) {
+            return CompareStatusFormatter.formatStatusClassFromContent(this.content).toLowerCase();
+        }
+
+        if (this.isPublishPending()) {
+            return `${CompareStatusFormatter.formatStatus(CompareStatus.EQUAL, null, true)} ${PublishStatus.PENDING}`.toLowerCase();
+        }
+
+        return CompareStatusFormatter.formatStatus(CompareStatus.EQUAL).toLowerCase();
```
alansemenov: This should be 

`CompareStatusFormatter.formatStatus(CompareStatus.EQUAL, null, true).toLowerCase()
`

CSS class is incorrect now:
 
image

Issue #1511, opened 2020-02-14
Refactor ProjectAccessControlComboBox


ashklianko:
Same code is used in IdProviderAccessControlComboBox (also PrincipalComboBox ). Move common code to lib-admin-ui

PR #1521, opened 2020-02-21
Projects: refactor ProjectAccessControlComboBox to avoid code duplica…

ashklianko:
…tion #1511


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1209

Issue #1529, opened 2020-02-24
Context is set to Default project when user has no permissions for any projects


alansemenov:
If you log in with a user who doesn't have permissions to ANY projects, context still seems to be set to "Default": content in the grid is fetched from default repo, you can close the notification dialog and start changing content. 

![image](https://user-images.githubusercontent.com/11612490/75156883-59c06e80-5713-11ea-9c9e-2a64da42292f.png)

This should not happen. The context should not be set unless there's a concrete project determined by user permissions. If user has no access to any projects, no requests to fetch the content should be made and all controls must be disabled.

![image](https://user-images.githubusercontent.com/11612490/75157581-b5d7c280-5714-11ea-80d3-cae75ec12b0b.png)

1. Disable search icon, New button, Create Task and cog icon in the toolbar (or implement readonly mode for the entire toolbar)
2. Hide issues button in the header
3. Replace Context selector with "_< No projects found >_"
4. Replace text in the preview panel with "_Contact your administrator to set up project permissions!_"

PR #1541, opened 2020-02-26
Context is set to Default project when user has no permissions for an…

ashklianko:
…y projects #1529

-disabling required elements when there are no available projects


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1212

Issue #1585, opened 2020-03-04
Project Editor and Owner should have access to HTML Area/Text component source


alansemenov:
We now show "Source" button in HTML Area and Text component to users with `cms.expert` role (or `sys.admin` or `cms.admin`). We should also allow access to users with "`Editor`" or "`Owner`" permissions for the current project, even if they don't have any of the aforementioned roles.

PR #1607, opened 2020-03-11
Project Editor and Owner should have access to HTML Area/Text compone…

ashklianko:
…nt source #1585

### DIFF
--- a/src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
+++ b/src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
@@ -33,6 +33,10 @@ import {SelectorOnBlurEvent} from 'lib-admin-ui/ui/selector/SelectorOnBlurEvent'
 import {BrowserHelper} from 'lib-admin-ui/BrowserHelper';
 import {FormEl} from 'lib-admin-ui/dom/FormEl';
 import {ArrayHelper} from 'lib-admin-ui/util/ArrayHelper';
+import {ProjectGetRequest} from '../../settings/resource/ProjectGetRequest';
+import {ProjectContext} from '../../project/ProjectContext';
+import {Project} from '../../settings/data/project/Project';
+import {PrincipalKey} from 'lib-admin-ui/security/PrincipalKey';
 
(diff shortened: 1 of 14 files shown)


src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
```diff
@@ -64,13 +68,38 @@ export class HtmlArea
         this.inputConfig = config.inputConfig;
 
         this.authRequest =
-            new IsAuthenticatedRequest().sendAndParse().then((loginResult: LoginResult) => {
-                this.editableSourceCode = loginResult.isContentExpert();
-            });
+            Q.all([new IsAuthenticatedRequest().sendAndParse(), new ProjectGetRequest(ProjectContext.get().getProject()).sendAndParse()])
```
alansemenov: This has to be changed so that `ProjectGetRequest` is sent only if current user is not either an admin or cms.expert.

ashklianko: Ok

src/main/resources/assets/js/app/inputtype/ui/text/HTMLAreaHelper.ts
```diff
@@ -97,4 +105,37 @@ export class HTMLAreaHelper {
...
+        if (isExpert) {
+            return true;
+        }
+
+        const isOwner: boolean = project.getPermissions().getOwners().some((owner: PrincipalKey) => {
+            return userPrincipals.some((userPrincipal: PrincipalKey) => userPrincipal.equals(owner));
+        });
+
+        return isOwner;
+    }
```
alansemenov: One pass of the `userPrincipals` array should be enough to find whether it's an editor or an owner:

```
const permissions = project.getPermissions();
const owners = permissions.getOwners();
const editors = permissions.getEditors();

return userPrincipals.some((userPrincipal: PrincipalKey) => 
  owners.indexOf(userPrincipal) > -1 || editors.indexOf(userPrincipal) > -1);
```

PR #1618, opened 2020-03-18
Remove "Project access" for the default project

ashklianko:


ashklianko:
update request for default project sends null for readAccess and permissions

Issue #1597, opened 2020-03-06
Filter out project roles from Principal selector


alansemenov:
We should filter out [project roles](https://github.com/enonic/xp/issues/7932) (starting with `cms.project.*`) in all principal selectors in Content Studio and the Users app.

PR #1621, opened 2020-03-18
Filter out project roles from Principal selector #1597

ashklianko:
-using extended principal loader to handle required filtering of project roles


src/main/resources/assets/js/app/security/FilterablePrincipalLoader.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class FilterablePrincipalLoader extends PrincipalLoader {
+
+    private static PROJECT_ROLE_PREFIX: string = 'cms.project.';
+
+    private forbiddenPrincipals: { [key: string]: PrincipalKey; } = {};
+
+    constructor() {
+        super();
+    }
```
alansemenov: No need for the constructor here.

src/main/resources/assets/js/app/security/FilterablePrincipalLoader.ts
```diff
@@ -0,0 +1,56 @@
...
+    resetForbiddenPrincipals(): FilterablePrincipalLoader {
+        this.forbiddenPrincipals = {};
+
+        return this;
+    }
+
+    skipPrincipals(principalKeys: PrincipalKey[]): FilterablePrincipalLoader {
+        principalKeys.forEach((principalKey: PrincipalKey) => {
+            this.forbiddenPrincipals[principalKey.toString()] = principalKey;
+        });
```
alansemenov: change this to 

`principalKeys.forEach((principalKey: PrincipalKey) => this.skipPrincipal(principalKey)); `

src/main/resources/assets/js/app/security/FilterablePrincipalLoader.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    private isAllowedPrincipal(principal: Principal): boolean {
+        const principalKey: PrincipalKey = principal.getKey();
+
+        if (this.forbiddenPrincipals[principalKey.toString()]) {
+            return false;
+        }
+
+        if (principalKey.isRole() && principalKey.getId().indexOf(FilterablePrincipalLoader.PROJECT_ROLE_PREFIX) === 0) {
+            return false;
```
alansemenov: I suggest implementing more flexible check for forbidden patterns than simply hardcoding project role prefix. This will help us easily expand the exclusion rules, should we need this in the future.

Here's an idea:

```
interface PrincipalPattern {
    type?: PrincipalType;
    id: string;
}

export class FilterablePrincipalLoader extends PrincipalLoader {

    private static PROJECT_ROLE_PREFIX: string = 'com.enonic.cms';

    private static forbiddenPrincipalPatterns: PrincipalPattern[] = [{
        type: PrincipalType.ROLE,
        id: `^${FilterablePrincipalLoader.PROJECT_ROLE_PREFIX}`
    }];

    ....

    private matchesForbiddenPattern(principalKey: PrincipalKey): boolean {
        return FilterablePrincipalLoader.forbiddenPrincipalPatterns.some((principalPattern: PrincipalPattern) =>
            (!principalPattern.type || principalPattern.type === principalKey.getType()) && principalKey.getId().match(principalPattern.id)
        );
    }

    private isAllowedPrincipal(principal: Principal): boolean {
        const principalKey: PrincipalKey = principal.getKey();

        if (this.forbiddenPrincipals[principalKey.toString()]) {
            return false;
        }

        return !this.matchesForbiddenPattern(principalKey);
    }
}

```

ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1220
Part of logic was moved to original PrincipalLoader since it already filters principals, now it can be overridden easier. 
In CS loader added possibility to add filter patterns

Issue #1646, opened 2020-03-27
Refactor usage of save confirmation dialog


alansemenov:
See https://github.com/enonic/lib-admin-ui/issues/1229

In `app-contentstudio` it will be mainly texts cleanup.

PR #1649, opened 2020-03-30
Refactor usage of save confirmation dialog #1646

ashklianko:
-If trying to close project wizard with invalid data then validating form items and showing errors


src/main/resources/assets/js/app/settings/wizard/ProjectItemNameWizardStepForm.ts
```diff
@@ -63,6 +64,12 @@ export class ProjectItemNameWizardStepForm
         });
     }
 
+    public validate(_silent?: boolean): ValidationRecording {
```
alansemenov: Not directly related to this PR, but from what I see parameter `_silent` is never used (correct me if I'm wrong). So we can remove it from here and its parent `WizardStepFrom` class.

Issue #1655, opened 2020-03-31
Allow site creation for Editor and Owner roles


alansemenov:
When you log in as `system.admin`:
![image](https://user-images.githubusercontent.com/11612490/78005790-5d5f9a80-733c-11ea-8a9a-60d8094d733b.png)

When you log in as a user who has Editor or Owner permissions on a project (and no `cms.cm.app` role) you can no longer select Site in the New Content dialog:
![image](https://user-images.githubusercontent.com/11612490/78005990-aadc0780-733c-11ea-806e-5d3afccc56c0.png)

PR #1662, opened 2020-03-31
Allow site creation for Editor and Owner roles #1655

ashklianko:


sigdestad:
Hmm... Editors should not be able to create site IMHO? Currently, only CMS admin's can create sites right?

sigdestad:
This is related to the fact that only "Site admins" can currently add/remove apps I believe? Project owners should now have the same permission I guess?

alansemenov:
@sigdestad ok, so - Owners should be able to create sites, but not Editors?

sigdestad:
Yes

src/main/resources/assets/js/app/create/NewContentDialog.ts
```diff
@@ -318,11 +319,35 @@ export class NewContentDialog
...
+        if (isContentAdmin) {
+            return Q(this.doFilterContentTypes(contentTypes));
+        }
+
+        return ProjectHelper.isUserProjectOwnerOrEditor(loginResult).then((isOwnerOrEditor: boolean) => {
+            if (isOwnerOrEditor) {
+                return Q(this.doFilterContentTypes(contentTypes));
+            }
+
+            return Q(this.doFilterContentTypes(contentTypes, false));
```
alansemenov, line 334: This can be just 

`return Q(this.doFilterContentTypes(contentTypes, isOwnerOrEditor));`

ashklianko:
Updated UI part, but backend refuses to update site for owner if he doesn't have admin or cms.admin roles, so backend to be updated also.

alansemenov:
Backend task: https://github.com/enonic/xp/issues/8005

PR #1690, opened 2020-04-07
Refactor request classes to support handling of non-JSON responses

ashklianko:
1. Raw response from backend is not always json, added param 'isJsonResponse' to manage if raw response to be wrapped into JsonResponse class
2. Removed unwanted generic <JSON> param from all requests and loaders since it is a) Complicates reading request classes b) not always used
3. Moved request path resolving to ResourceRequest to simplify adding path elements and to reuse duplicated code


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1237

PR #1703, opened 2020-04-13
Separate API call to update project language

ashklianko:
Depends on https://github.com/enonic/lib-admin-ui/pull/1241


src/main/resources/assets/js/app/settings/wizard/ProjectReadAccessWizardStepForm.ts
```diff
@@ -14,47 +14,112 @@ import {ProjectReadAccess, ProjectReadAccessType} from '../data/project/ProjectR
...
+        layoutPromises.push(this.layoutLanguage(item.getLanguage()));
+
+        if (item.isDefaultProject()) {
+            return Q.all(layoutPromises).spread<void>(() => Q<void>(null));
+        }
+
+        layoutPromises.push(this.layoutReadAccess(item.getReadAccess(), item.getPermissions()));
 
-        this.updateFilteredPrincipalsByPermissions(item.getPermissions());
+        return Q.all(layoutPromises).spread<void>(() => Q<void>(null));
```
alansemenov, line 53: This would probably be more readable with one `return`:

```
if (!item.isDefaultProject()) {
  layoutPromises.push(this.layoutReadAccess(item.getReadAccess(), item.getPermissions()));
}        

return Q.all(layoutPromises).spread<void>(() => Q<void>(null));

```

src/main/resources/assets/js/app/settings/wizard/ProjectReadAccessWizardStepForm.ts
```diff
@@ -14,47 +14,112 @@ import {ProjectReadAccess, ProjectReadAccessType} from '../data/project/ProjectR
...
+        locales.some((locale: Locale) => {
+            if (locale.getId() === language) {
+                selectedLocale = locale;
+                return true;
+            }
+
+            return false;
+        });
+
+        return selectedLocale;
```
alansemenov, line 91: this can be replaced with

`return locales.find((locale: Locale) => locale.getId() === language);`

PR #1735, opened 2020-04-20
Keyboard navigation in the Content grid stops working after project s…

ashklianko:
…elected #1733

-Keybindings are being shelved/unshelved in wrong order when project selection dialogs is open, thus closing it doesn't bring treegrid's key events back
-Now binding treegrid's key events AFTER project selection dialog is open


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1253

Issue #1739, opened 2020-04-21
Remove Access step from the Content Wizard


alansemenov:
* Remove `SecurityWizardStepForm` step
* Change behaviour of the lock icon in the Wizard step navigator to open Edit Permissions dialog

PR #1755, opened 2020-04-24
Remove Access step from the Content Wizard #1739

ashklianko:
-Edit permissions button will reside to the right of step navigator buttons


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1259

Issue #1762, opened 2020-04-29
Incorrect layout in Wizard Step Navigator Toolbar


sgauruseu:
1. Open `Large tree`

BUG:   Incorrect layout.

![image](https://user-images.githubusercontent.com/3728712/80588128-6a95a500-8a20-11ea-8ffa-aa0da4724f0b.png)

PR #1765, opened 2020-04-30
Incorrect layout in Wizard Step Navigator Toolbar #1762

ashklianko:
-Wizard toolbar styles to be applied without live edit also, thus taking them to the right place


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1268

Issue #1314, opened 2019-12-10
Don't use com.enonic.xp.util.StringTemplate


rymsha:
com.enonic.xp.util.StringTemplate is going to be removed. https://github.com/enonic/xp/issues/7715
Applications should use template engines (mustache, thymeleaf) for html/json or TS string interpolation for simple cases.

PR #1775, opened 2020-05-05
Don't use com.enonic.xp.util.StringTemplate #1314

ashklianko:
-Using Apache's StringSubstitutor instead (fallback solution, dev jar size increased 6.1M -> 6.7M due to Apache dependencies)

Live edit is rendered on backend via iframe's url; Switching it to js controller is possible but will require heavy rework of how it is done and probably will be overkill for now, but let's discuss.
I see it this way:
1. Extract rendering function to a portal lib (most complicated part)
2. Create service with controller in app-cs to receive live edit render call 
3. Service invokes portal lib's render function that returns html
4. Process html with template engine
5. Add required live edit injections (that are done now in LiveEditInjection.java)
6. Return response from a controller


alansemenov:
Maybe it should be in `lib-admin` instead, since LiveEdit is admin-only feature? No need to pollute `lib-portal` with this...

ashklianko:
Page/Live edit rendering functionality is heavily integrated in portal api, need to investigate is it possible at all extract it from there to other place.

Issue #1383, opened 2020-01-09
Code tag cannot be removed in HTML Editor


vot-enonic:
Working in htmlArea, if you mark text and apply "Code" as paragraph format, the formatting will not be replaced if you apply e.g. "Normal". You need to go into "Source" to remove the code tags. Shouldn't it be possible to remove these tags from the WYSIWYG editor by applying another paragraph format?

alansemenov:
It's actually worse than that - after applying "Code" formatting the format dropdown won't display correct selection (see screenshot below where it says "Normal" even though the cursor is in the middle of `<code>` tag)

![image](https://user-images.githubusercontent.com/11612490/78251076-1961c680-74f1-11ea-889e-dd398775bbd6.png)
 
It seems to work fine in [CKE 4 demo](https://ckeditor.com/ckeditor-4/demo/)

ashklianko:
The Format plugin we use is intended for block level elements, and code tag is inline so it doesn't work well.
I will add custom plugin with button on toolbar to wrap/unwrap selection into code tag

alansemenov:
@ashklianko The thing is we don't want any additional buttons in the toolbar... Can you research why it works fine in CKE 4 demo? Select "Computer code" from the "Block styles" dropdown.

ashklianko:
That is also a separate plugin called  "Styles Combo"  that can handle inline elements

PR #1847, opened 2020-05-22
Code tags will not be removed in WYSIWYG #1383

ashklianko:
-Format plugin used now is intended for block elements only
-Replaced format plugin with Styles plugin that has more options to customize and allows inline elements


src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
```diff
@@ -1015,6 +1010,61 @@ class HtmlEditorConfigBuilder {
...
+            return {name: 'Heading 4', element: 'h4'};
+        }
+
+        if (heading === 'h5') {
+            return {name: 'Heading 5', element: 'h5'};
+        }
+
+        if (heading === 'h6') {
+            return {name: 'Heading 6', element: 'h6'};
+        }
```
alansemenov, line 1063: Very important that we localise all the options (also the ones created in `initCustomStyleSet`), otherwise we'll lose translations:

Current (norwegian language):
image

New version:
image

**phrases.properties:**
```
  text.htmlEditor.styles.heading = Heading {0}
```
```
private headingToStyleObject(heading: string): any {
  const headingIndex = heading.match(/\d/g)[0];
  return {name: i18n('text.htmlEditor.styles.heading', headingIndex), element: heading};
}
```

ashklianko: Ok!

Issue #1856, opened 2020-05-25
Grid toolbar and Selection toggler are not updated after deleting a project in collapsed Projects -folder


sgauruseu:
**In content grid**
1. Expand a parent folder , click on a checkbox and select a child content.
2. Collapse the parent folder.
3. Click on Delete button and confirm the deleting 

![image](https://user-images.githubusercontent.com/3728712/82895305-ad359900-9f5c-11ea-9f53-ac20c6c4ee3d.png)
 Bug: - toolbar is not updated

1. Add new project 
2. Select the project and collapse 'Projects' folder.
3. Click on Delete button in the toolbar and delete this project 

**BUG**: The project is deleted , but Delete button is enabled and Selection toggler shows that a project is selected  

![image](https://user-images.githubusercontent.com/3728712/82824347-d6e5b600-9eb1-11ea-8d6d-5bdb1a2003fd.png)

**Expand Projects folder** :   no projects are selected

![image](https://user-images.githubusercontent.com/3728712/82824400-ee24a380-9eb1-11ea-8fe7-bb6cf12a700f.png)

PR #1860, opened 2020-05-26
Grid toolbar and Selection toggler are not updated after deleting a p…

ashklianko:
…roject in collapsed Projects -folder #1856


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1304

Issue #1857, opened 2020-05-26
[IE 11] Issues with inserting components in Page Component View


sgauruseu:
This bug can be linked with https://github.com/enonic/app-contentstudio/issues/1484

1. Open Features in Edge browser.
2. Open Page Component View and insert an image

**Issue 1** : Display name of the image is not displayed in the view 

![image](https://user-images.githubusercontent.com/3728712/82883225-d3523d80-9f4a-11ea-9858-61fadf80aa93.png)

3. Insert a part then swap components using DnD 

**Issue 2**  Error appears after inserting a part or swapping components

![image](https://user-images.githubusercontent.com/3728712/82883503-3348e400-9f4b-11ea-9668-a784140d2f83.png)

ashklianko:
Issue is not actual in new Edge on Chromium, but I am currently fixing IE11 bug with Page Component View - it is not clickable at all, so after that fixed I'll check both cases.
Issue 2 seems to be fixed in #1484

PR #1874, opened 2020-05-28
[IE 11] Issues with inserting components in Page Component View #1857

ashklianko:
-Helping IE receive click events


ashklianko:
Also requires https://github.com/enonic/lib-admin-ui/pull/1309

PR #1903, opened 2020-06-04
Lokalise: Translations update

jsi:
This pull request was initiated by Lokalise (user Jørgen Sivesind) at 2020-06-04 09:46:03


jsi:
Not correct files.

Issue #1953, opened 2020-06-12
Implement tree structure in the project selector list


alansemenov:
Change project selector in the Content Studio header and flat list in the "Select context" modal dialog to be tree-structured, based on the new parent project Id field in the `list` API.

PR #2002, opened 2020-06-26
Implement tree structure in the project selector list #1953

ashklianko:


src/main/resources/assets/js/app/project/list/ProjectList.ts
```diff
@@ -1,16 +1,90 @@
...
+            result.push(...this.unwrapProjectChildren(defaultProject));
+        }
+
+        const level0projects: Project[] = items.filter(
+            (project: Project) => !ProjectHelper.isDefault(project) && this.projectLevel.get(project.getName()) === 0);
+        level0projects.forEach((level0project: Project) => {
+            result.push(level0project);
+            result.push(...this.unwrapProjectChildren(level0project));
+        });
+
```
alansemenov, line 75: Hmm... Why do we have a special check for `defaultProject` and exclude if from `level0projects`  if we do the same operation (`result.push`, `this.unwrapProjectChildren`) for both?

src/main/resources/assets/js/app/project/list/ProjectList.ts
```diff
@@ -1,16 +1,90 @@
...
+        return result;
+    }
+
+    private unwrapProjectChildren(project: Project): Project[] {
+        const result: Project[] = [];
+
+        this.projectChildren.get(project.getName()).forEach((child: Project) => {
+            result.push(child);
+            result.push(...this.unwrapProjectChildren(child));
```
alansemenov, line 84: This piece of code is repeated three times now, one here and two in `getDirectProjectChildren` (should be once there, as I wrote above). But it seems like it deserves its own method.

src/main/resources/assets/js/app/project/list/ProjectList.ts
ashklianko: To ensure that Default project will always be first on the projects list

alansemenov: Ok, then we can just sort `result` before returning it, to make sure the default project is first.

ashklianko: We need to have default's chidren projects (with their children etc.) to be in the list right after default entry, so it is better this way - first process default with children, then the rest

ashklianko:
Some issue with github, had to create new PR https://github.com/enonic/app-contentstudio/pull/2007

PR #2121, opened 2020-07-30
Misc UI fixes #2110

### LINKED ISSUE #2110: Misc UI fixes
**Context selector:**

1. Project selector in the header should have a tooltip: "Select project" (same as the modal dialog's title)
2. There should be 10px side paddings inside ProjectListItem elements so that the icon has some air around it
3. Selected row must be in the same blue color that we use in the content grid etc.
image

**Widget:**

4. We should have more vivid highlighting of the current layer's item in the widget.
Here's how it looks now:
image

Here's the suggested design:
image

Note the thicker border and shallow drop-shadow (like in the Version History widget).
I'm still very confused to why we need this side-arrow. It should not be necessary if we have decently vivid highlighting, but let's keep it for now.

5. Current layer's box should be auto-expanded
6. Side-arrow now jumps up and down on expand/collapse. I suggest moving it up to point to the upmost border, then it won't jump.

7. For an inherited item we should apply opacity to the entire box, not only to the item viewer. Now it's very difficult to spot the difference. Here we are entirely missing highlighting of the current layer. We should not have any highlighting on click/focus, current layer should always be highlighted.

image

8. I don't think we should show default language of the layer if the content is inherited. In the screenshot above the bottom item is inherited and language of the content is still English, but we show Russian flag (because default language of the layer is Russian). This is very confusing. We should show the language flag only if the content item is local/not inherited AND has language set. Which takes us to the next item...

9. If there's no language flag to show (when the item is inherited or local with no language set), show content icon. This will be another visual sign that the item is not localised.

10. Localise button should be shown only inside the current layer's box. For all others it should be "Open".

11. "Open" button inside a parent layer's box should open Content Wizard in THAT layer's context (parent project's name in the URL). Now it opens content wizard in the current layer. 

12. "Layers" widget should not be available if current project doesn't have a parent project. Remember that not everyone will use this feature, so they should not care about this widget.

ashklianko:


alansemenov:
@ashklianko We should add 10px bottom padding to the layers widget so that drop-shadow is not cut off. As of now the widget doesn't have a dedicated CSS style, so we need to add that as well.

Issue #2123, opened 2020-07-30
Browse toolbar and Layers widget is not updated after changing a language in subproject


sgauruseu:
1. Create a subproject  with a language in Default.
2. Go to the subproject context, select a content - Localize button gets visible in the toolbar, expected language is displayed in the widget (OK)
3. Switch to settings and  remove the selected language in wizard, save it .
4. Switch to content again.

BUG: Localize button should not be present in the toolbar, widget should be updated: 
 
![image](https://user-images.githubusercontent.com/3728712/88926036-1350dc00-d27e-11ea-9fc8-2a0fd7bc7a84.png)

Issue #2196, opened 2020-08-18
Always replace "Edit" with "Localised" for inherited item


alansemenov:
When current layer doesn't have default language set, **Edit** button won't be replaced with **Localise** in the grid toolbar. We should always do that for an inherited item, even when there's no default language, because in this context "**Localise**" simply means "_make local_", not necessarily "_translate_".

![image](https://user-images.githubusercontent.com/11612490/90507162-bac27f80-e155-11ea-870b-e40ca9feda46.png)

PR #2198, opened 2020-08-19
Browse toolbar and Layers widget is not updated after changing a lang…

ashklianko:
…uage in subproject #2123


ashklianko:
Browse toolbar will be fixed in #2196

PR #2207, opened 2020-08-25
Always replace "Edit" with "Localise" for inherited item #2196

ashklianko:


sigdestad:
Did not check code, but phrase should state "localize" imho

alansemenov (approved):

Issue #2211, opened 2020-08-26
Consistent logic for displaying project/layer icons


alansemenov:
The logic for displaying project/layer icons is currently inconsistent. The closest to how it should be is the logic used inside the Layers widget, except there we don't show custom project/layer icon (which actually has highest priority and should override fallback/language icon).

The rules for displaying correct icon are as follows:
1. Use custom icon if it's set
else
2. Use flag icon based on default language of the project/layer, if it's set
else
3. Use fallback (static) project/layer icon

This logic should be consistent and applied in:

1. Current project/layer displayed in the header
2. "Select Project" modal dialog
3. Layers widget
4. Settings grid
5. Project wizard form (note that here we should also dynamically toggle between fallback icon and flag icon on change of language selector value)

PR #2236, opened 2020-08-31
Consistent logic for displaying project/layer icons #2211

ashklianko:


ashklianko:
Also small issue was resolved in https://github.com/enonic/lib-admin-ui/pull/1443, try switching from default project to a layer with a language set and look at project selector in app bar.

Issue #2215, opened 2020-08-26
Show inheritance option in the Sorting dialog


alansemenov:
If `inherit` field in ContentSummary contains `SORT`, it means that the item is inheriting sorting order of its parent. In this case we should do the following changes to the Sorting dialog:
1. Disable the list of items
2. Change subtext to "_Select sorting type from the dropdown above_".
3. Fetch sorting setting of the parent (using `GetContentByIdRequest` request in the context of the parent project) and display it as selection of the dropdown as "_Inherited_: " + sorting of the parent.

![image](https://user-images.githubusercontent.com/11612490/91316818-91d76580-e7b9-11ea-8862-1dff5a00cbf8.png)

4. It should also be present and highlighted in the dropdown list:

![image](https://user-images.githubusercontent.com/11612490/91318312-605f9980-e7bb-11ea-8e1c-0bfe4996c62f.png)

Issue #2218, opened 2020-08-26
Update inherit settings of the content when toggling between inherited and non-inherited sorting options


alansemenov:
If user switches sorting settings of a content item from "Inherited" to one of the regular options, or the other way around, we should do the following:
1. Toggle header subtitle (see https://github.com/enonic/app-contentstudio/issues/2215)
2. Enable or disable the list of child items
3. Enable the Save button  

Once the Save button is clicked, we should not only save the new sorting but also update inherit settings of the item via [API](https://github.com/enonic/xp/issues/8303) by adding or removing SORT option.

ashklianko:
Pushed into https://github.com/enonic/app-contentstudio/pull/2269

PR #2269, opened 2020-09-08
Show inheritance option in the Sorting dialog #2215

ashklianko:


ashklianko:
https://github.com/enonic/lib-admin-ui/pull/1456 also required

alansemenov (changes requested):
Tested it now, and there are issues when item has "local" sorting and I want to reset to "inherited":

1. If "local" sorting is the same as "inherited" sorting, Save button remain disabled when I select "Inherited: " option.

Screenshot 2020-09-15 at 16 16 20

2. If "local" and "inherited" sorting are different, I can select "Inherited" option and Save but it won't call the API to "restore" the sorting inheritance. When I reopen the dialog it shows the new sorting type, but as "local", not "inherited":

image

alansemenov:
So basically https://github.com/enonic/app-contentstudio/issues/2218 is not completely done yet

alansemenov:
`restoreInherit` API is [ready](https://github.com/enonic/xp/issues/8303)

ashklianko:
Ok, will update

ashklianko:
Updated

alansemenov (changes requested):
One last bug: it's not possible to switch between inherited and non-inherited for the same sort option, for example from "_Display name: A->Z_" to "_Inherited: Display name: A->Z_" and vice versa.

ashklianko:
> One last bug: it's not possible to switch between inherited and non-inherited for the same sort option, for example from "_Display name: A->Z_" to "_Inherited: Display name: A->Z_" and vice versa.

Done

Issue #2298, opened 2020-09-18
Localisation should work only in the current layer


sgauruseu:
1. Create 2 layers in Default project, Select Norsk and En languages in these layers.
2. Switch to the second layer , open Layers widget and click on `Open` button in the second item:
![image](https://user-images.githubusercontent.com/3728712/93598102-dfb04980-f9c4-11ea-99ef-6fa6b784b1ab.png)

**BUG**:  Language has been set and this content gets localized after opening it from another layer: 

![image](https://user-images.githubusercontent.com/3728712/93599054-379b8000-f9c6-11ea-8a85-80deb8639862.png)

PR #2331, opened 2020-09-30
Localisation should work only in the current layer #2298

ashklianko:
-Added 'localize' url action
-When opening inherited item in current layer from browse panel url will be constructed with 'localize'
-Layer language for inherited item in wizard panel will only be set only if 'localize' presents in url
-'localize' in url will be replaced with 'edit' after item is saved or if item is initially not inherited


alansemenov (approved):
Works like a charm, great job :)

Issue #2362, opened 2020-10-12
Custom config breaks layout of HTML Area toolbar


alansemenov:
This config 
`<include>Format | Bold Italic | BulletedList NumberedList | Outdent Indent | SpecialChar Anchor Image Macro Link Unlink Table | CopyFormatting RemoveFormat | Subscript Superscript</include>`

make the toolbar look like this on a narrow screen:

![image](https://user-images.githubusercontent.com/11612490/95722341-16038080-0c74-11eb-9eaf-ff4a80b6d8ea.png)

PR #2365, opened 2020-10-12
Fixed wrapping of customised HTML Area toolbar (#2362)

alansemenov:


ashklianko (approved):
There are still issues with single icons floating at the end of toolbar, but that for a separate task

PR #2373, opened 2020-10-13
Trailing space in content path breaks child resolution

ashklianko:


sigdestad:
What does this change do? The problem is strictly speaking a backend issue?

ashklianko:
In addition to backend fix we still need to filter invalid input before sending request to backend

alansemenov:
@sigdestad Backend fix is a known, potentially breaking change, see [here](https://github.com/enonic/xp/issues/7243). In this one we are trimming spaces on the client before sending data to the server which will quickly solve the Collicare's problem.

Issue #2370, opened 2020-10-13
Dynamic validation of content path


alansemenov:
Current behaviour when setting or changing a display name is that we dynamically generate path according to display name, unless the content is published. Content path can also be changed manually (even after publishing). On content save, we run server-side validation and give a notification if content exists. 

We should make this validation a bit more pre-emptive and check if path exists once the current value is modified (either manually or via display name). Also, the message should be more visible (now it pops up at the bottom of the screen).

It should work like this (very similar to email validation in the User app):
1. Whenever path is changed, either manually or via display name field, have a very short (start with 1 second) delay to make sure user stopped typing and then make a request to `contentsExistByPath` API to check whether such path already exists.

2. If it does, apply error styling to the path input and show span with text "_Not available_" (also in red). If it doesn't, remove error styling and the text.

3. If path is modified and such path already exists, the content should be considered invalid: Save button should be disabled and there should be invalid icon over the thumbnail. Changing path to a non-existing value should make the content valid again.

![image](https://user-images.githubusercontent.com/11612490/95869966-3dd60f80-0d6c-11eb-856b-94f49d7d6740.png)

![image](https://user-images.githubusercontent.com/11612490/95869731-fbacce00-0d6b-11eb-88a8-f78c3f8c5e00.png)

PR #2375, opened 2020-10-14
Dynamic validation of content path #2370

ashklianko:


alansemenov (changes requested):
Hooking into `notifyPropertyChanged` means making a request every 100 ms and this is too fast. We need to have a one second delay, at least. But don't change delay of `notifyPropertyChanged` because that we need to be almost instant.

alansemenov (approved):

PR #2482, opened 2020-11-06
Changing part config inside a content with nested option-set causes refresh of the entire page editor

alansemenov:


ashklianko (approved):
Hope no side effects will happen

Issue #1939, opened 2020-06-09
Form fields in read-only project wizard get focus on tab


sgauruseu:
1. Login in with contributor user.
2. Open existing project(double click on a row with project)
3. Project wizard is loaded. Press the Tab key

BUG - Inputs in the wizard get enabled

The same issue in Content Wizard

PR #2511, opened 2020-11-12
Project wizard gets enabled after a contributor pressing Tab key #1939

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1576

alansemenov (changes requested):
Let's try and solve this in a more generic way, in the base `WizardPanel` class, with protected classes responsible for sending authentication request, checking permissions and enabling/disabling controls based on permissions.

Issue #2530, opened 2020-11-16
Enable marking all items as ready in the Publishing Wizard


alansemenov:
If the list of items to be published contains at least one item in "_In progress_" state (use the same condition we use to show "**Work in progress!**" warning at the top), the "_Publish_" button should be converted into menu button with "_Mark as ready_" option in the dropdown. 

The button label should show the number of items in the brackets "_Mark as Ready (X)_" . Only selected items (with checked checkbox) should be counted, so the number should change dynamically whenever "In progress" items are selected or unselected in the list. If all in progress items are unselected, the dropdown of the "Publish" menu button should be disabled.

Once "_Mark as Ready_" is clicked, we should build the list of selected items in "_In progress_" state and send their ids to the _markAsReady_ API. After that the dialog should be refreshed, with no warning at the top and no menu for the "Publish" button.

![image](https://user-images.githubusercontent.com/11612490/99278506-6cde1600-282f-11eb-8e3f-0fa58f031830.png)
![image](https://user-images.githubusercontent.com/11612490/99278521-723b6080-282f-11eb-97a0-07347040fcbe.png)

PR #2552, opened 2020-11-23
Enable marking all items in "In progress" state as "Ready" #2530

ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1588


alansemenov:
This will have to be reworked once https://github.com/enonic/xp/issues/8485 is done

alansemenov (changes requested):
Fantastic job, Sasha! One small thing to fix: required items (those with ids in the `requiredContents` field of the API response) should be excluded from the number in "`Exclude all (X)`" link, and shouldn't be added to `excludedIds` field when sending the response.

In the example below there are 2 in progress items, but one of them is required for publishing. In this case the link should say "_Exclude all (1)_", and we should not attempt to exclude the second item from publishing when the link is clicked.

image

ashklianko:
Updated!

sigdestad:
Hmm.. Alan, this is positively wrong IMHO. 
In this case (where there are 2 in progress items) we should exactly say "exclude all (2)"

You cannot really know how many items are "required" or not until you try removing all "in progress".

alansemenov:
What do you mean? Of course we know how many are required or not in the current list.

sigdestad:
I mean, we will attempt to remove them all, regardless of if the current state says some of them are required

sigdestad:
This is the whole point of the task..

sigdestad:
Also, what I mean is that when you attempt to remove all, it might happen that some end up being required anyway (i.e. parent of selected)

alansemenov:
Why "attempt" anything there? We know exactly how many we can exclude, that's how we know which "x" icons to disable. Obviously after each exclusion the number can change, but we always know how many we can exclude at any given moment. If it says "Exclude all (1)" for one invalid item that cannot be excluded and nothing happens when you click the link, this is obviously a bad UX. 

Another alternative is to always show the total number of items to exclude but disable the link if none of the items can be excluded.

sigdestad:
So, how many times do you want me to click “exclude all“ then?

alansemenov:
Just once. There will never be a case when you will get **more** items to exclude than you originally had. Let's try it out and if we encounter some side issues then we'll consider other alternatives.

alansemenov:
Remember that we refresh these numbers after each exclusion. So if for example some "in progress" items were required by some "invalid" items, then the number in "Exclude all (X)" for the "in progress" items will be recalculated after you excluded the invalid ones and vice versa.

sigdestad:
The idea is that you press the button _once_. If there are any items left after this, they cannot be removed (parent items of Main items that have never been published)

alansemenov:
> The idea is that you press the button once. If there are any items left after this, they cannot be removed (parent items of Main items that have never been published)

That is exactly what I wrote in the original comment. Items that cannot be excluded should not be a part of X in the "_Exclude all (X)_" link. Because they cannot be excluded.

sigdestad:
We'll have to take this F2F

alansemenov:
To sum up the changes:

1. Show total number of "in progress"/"invalid" items in the corresponding links
2. Always add all corresponding ids to `excludedIds` field of the API request when the link is clicked
3. Check if there are any items that were **not** excluded after receiving API response. If there's at least one item that was not excluded, give a warning (yellow toast popup) saying "Some of the items are required for publishing and could not be excluded" and hide the corresponding link.

ashklianko:
Updated

alansemenov (approved):

Issue #2556, opened 2020-11-23
Add ui tests for Image Selector with deleted image


sgauruseu:
Image Selector with deleted image is not properly validated #2555

PR #2557, opened 2020-11-24
ui tests for Image Selector with deleted image #2556

sgauruseu:


jsi (approved):
Looks good.  Except second test has a bad name.  It can be fixed for the next PR.

Issue #1995, opened 2020-06-23
Image Wizard - Save button is enabled after reopening a zoomed image


sgauruseu:
[image_spumans01.zip](https://github.com/enonic/app-contentstudio/files/4818345/image_spumans01.zip)

1. Upload the image in Root directory 
2. Open the image and zoom it .
3. Reopen the image 

![image](https://user-images.githubusercontent.com/3728712/85384845-3d5e0280-b54a-11ea-98d8-6e4042ba52ed.png)

BUG Save is enabled after reopening this image

This issue can be linked with https://github.com/enonic/app-contentstudio/issues/1363

alansemenov:
Reappeared after the fix of https://github.com/enonic/app-contentstudio/issues/1971

PR #2571, opened 2020-11-26
Image Wizard - Save button is enabled after reopening a zoomed image …

ashklianko:
…#1995


alansemenov (changes requested):
With this change you no longer see that the image has been modified - there's no "Reset filters" link when opening a zoomed image.

Before the change:
image

ashklianko:
Updated

alansemenov (approved):
Good stuff

PR #2579, opened 2020-11-30
Form fields in read-only project wizard get focus on tab

ashklianko:


ashklianko:
Requires enonic/lib-admin-ui#1584

alansemenov (approved):

PR #2619, opened 2020-12-07
Move Layers widget into Content Studio+

ashklianko:


src/main/resources/i18n/phrases.properties
```diff
@@ -297,9 +297,6 @@ widget.pagetemplate.default=Page Template is not used
 widget.pagetemplate.notfound=Not found
 widget.pagetemplate.notused=Not used by any content
 widget.useraccess.restricted=Restricted access to item
-widget.layers.showmore=Ascendants ({0})
-widget.layers.more=Descendants ({0})
-widget.layers.showall=Show all ({0})
```
alansemenov, line 302: Can we keep the texts for now? It's tempting to get CS+ to have its own text bundle - and we will definitely do that in the future - but for now there are only a few phrases, so let's for now avoid asking our external contributors to translate yet another repository.

Issue #2627, opened 2020-12-09
Project list is fetched on every change in Project Wizard form


alansemenov:
We don't auto-save the form, so fetching the list of projects for every change doesn't seem right? Even if we did, the parent project selector is disabled for an existing project, so I'm not sure why we ever do this here. 

![layer-list](https://user-images.githubusercontent.com/11612490/101651372-4893e480-3a3d-11eb-9c71-b70f2b9ab476.gif)

PR #2632, opened 2020-12-13
Project list is fetched on every change in Project Wizard form #2627

ashklianko:
-Creating app containers (main or settings) only on demand
-More generic app containers
-Skipping redundant server requests


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1640

alansemenov (approved):

Issue #2293, opened 2020-09-17
Broken tree in the Project Selection dialog when user doesn't have access to all layers


sgauruseu:
1. Create 4 nested layers in  Default project: 
![image](https://user-images.githubusercontent.com/3728712/93455952-b2d93500-f8e5-11ea-8e8e-2b0be070e894.png)

2. Create an user with Owner role for the first and fourth layers.
3. Do login with the user. `Select project dialog` loads.

**Issue 1:**   Еhe project tree is not shown correctly.

![image](https://user-images.githubusercontent.com/3728712/93457391-95a56600-f8e7-11ea-8dbc-ed37a6d106a1.png)

Expected :  All projects should be displayed in this tree. But layers for which the user has no suitable roles should be displayed and disabled. Default project should be displayed and disabled as well.

**Issue 2**  
1. Switch to content mode and select 4th layer.
2. Open Layers Widget.

**Expected**: All layers should be displayed in the widget. But layer-item for which the user has no suitable roles should be displayed and disabled.

**Actual**:    Only one layer is present in the widget:

![image](https://user-images.githubusercontent.com/3728712/93458337-da7dcc80-f8e8-11ea-9341-f33cc2cceca6.png)

The same issue in  `Layers Content Tree` dialog : 

![image](https://user-images.githubusercontent.com/3728712/93458583-2761a300-f8e9-11ea-9a5e-96c5a9612142.png)

**Expected tree:**  - All layers should be displayed in the dialog, but only 2 items should be enabled

![image](https://user-images.githubusercontent.com/3728712/93458671-495b2580-f8e9-11ea-847c-fa463c2ff82e.png)

PR #2662, opened 2020-12-18
Broken tree in the Project Selection dialog when user doesn't have ac…

ashklianko:
…cess to all layers #2293

-Central part of a task is ProjectsTreeBuilder.ts that is responsible for fetching missing parts of a project inheritance tree. Building tree only to the top, not showing child projects if they are not available for user


ashklianko:
Styling of a non-availabe items to be changed on demand in next task

alansemenov (approved):
Great job

Issue #2712, opened 2021-01-05
BrowsePanel: unify handling of TreeGrid data change event throughout all BrowsePanel inheritors


ashklianko:
@ashklianko commented on [Tue Dec 08 2020](https://github.com/enonic/lib-admin-ui/issues/1627)

Currently same logic with slight deviations is spread and diplicated in ContentBrowsePanel, ApplicationsBrowsePanel and BrowsePanel itself, 
need to toggle preview item and update browse actions properly

PR #2713, opened 2021-01-05
BrowsePanel: unify handling of TreeGrid data change event throughout …

ashklianko:
…all BrowsePanel inheritors #2712

-handling selection change and highlighting change in one place in BrowsePanel
-Using unified ViewItem interface throughout BrowsePanel and it's grid and other classes to wrap actual data so inheritors could easily extend it
-removed unused and redundant classes
-moved some Content classes out of lib-admin-ui (ContentIconUrlResolver, ContentSummaryViewer)
-removed creation of header in ItemStatisticsPanel so it to be used on demand
-moved check of item to be renderable out of lib-admin-ui since it is only relevant in content studio where we render content on preview
-refactored check of content to be renderable: now doing check when fetching ContentSummaryAndCompareStatus to
a) avoid spaghetti code actions check and item preview logic
b) skip fetching isRenderable request for same item when navigating in TreeGrid, we currentlyload items in ContentTreeGrid by 10, so won't be a problem to fetch it immediately since request is lightweight


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1681

alansemenov (approved):

Issue #1416, opened 2020-01-15
LiveEditPageProxy loads component in the unsafe way


edloidas:
`LiveEditPageProxy.loadComponent` sends a requests for the component and transforms the `string` to an HTML element, which is unsafe operation. Find a solution to completely remove this part of code or sanitize and validate the received data.

PR #2818, opened 2021-01-25
LiveEditPageProxy loads component in the unsafe way #1416

ashklianko:
-using DOMPurify to sanitize loaded html


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/1754

alansemenov (approved):

PR #2884, opened 2021-02-17
Responsive header in the Content Wizard

ashklianko:
To avoid overcomplicated logic of wrapping/unwrapping static path part and calculating it's length on value change/resize:
1. Using flex styling and showing static part with title 
2. Removed redundant elements
3. Letting dynamic input grow as much as needed and static path shrink down to zero

@alansemenov please check if this approach is appropriate

Requires https://github.com/enonic/lib-admin-ui/pull/1802

### DIFF
--- a/src/main/resources/assets/styles/wizard/content-wizard-panel.less
+++ b/src/main/resources/assets/styles/wizard/content-wizard-panel.less
@@ -7,19 +7,22 @@
       display: none;
     }
 
-    input:last-of-type {
-      display: inline-block;
-      margin-right: 10px;
+    .bottom {
+      input {
+        display: inline-block;
+      }
     }
 
     &.path-exists {
-      .path-error {
-        display: inline-block;
-      }
+      .bottom {
+        .path-error {
+          display: inline-block;
+        }
 
-      input:last-of-type {
-        border: 1px solid @admin-input-red;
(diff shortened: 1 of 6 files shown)


ashklianko:
Fixed merge conflicts

src/main/resources/assets/styles/wizard/content-wizard-panel.less
```diff
@@ -62,9 +55,6 @@
     &._0-240,
     &._240-360,
     &._360-540 {
-      .wizard-header {
-        max-width: calc(100% - 105px);
-      }
```
alansemenov, line 67: This piece of code used to fix this issue:

image

alansemenov (changes requested):
image

The icon should be vertically centred

alansemenov:
Plenty of space here to fit the entire path, but both display name and path are truncated:

image

ashklianko:
> Plenty of space here to fit the entire path, but both display name and path are truncated:
> 
> image

I've limited length to 660 px to be same as max length of wizard step forms

alansemenov:
> I've limited length to 660 px to be same as max length of wizard step forms

Ah, that's why... Ok, so this one is correct then :)

ashklianko:
Updated, also moved minimize arrrow button to a new place in a toolbar

alansemenov (approved):

Issue #2896, opened 2021-02-23
Content Manager Expert should not have access to content permissions


alansemenov:
Users with this role can open the Edit Permissions dialog inside the Content Wizard and change permissions there. The change will be blocked on the server, but we shouldn't show the UI controls in the first place.

"Edit" link is correctly hidden for Content Manager Expert in the "Details" widget, and we should use the same conditions to hide the lock icon in the Content Wizard.

PR #2901, opened 2021-02-24
Content Manager Expert should not have access to content permissions …

ashklianko:
…#2896


src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1823,41 +1824,48 @@ export class ContentWizardPanel
...
         if (hasAdminPermissions) {
-            this.toggleElementsVisibility(true);
+            this.toggleSettingsElementsVisibility(true);
+            this.editPermissionsToolbarButton.setVisible(true);
         } else {
-            ProjectHelper.isUserProjectOwner(loginResult).then((isOwner: boolean) => this.toggleElementsVisibility(isOwner));
+            ProjectHelper.isUserProjectOwner(this.loginResult).then((isOwner: boolean) => {
+                const isContentExpert: boolean = this.loginResult.isContentExpert();
+
+                this.toggleSettingsElementsVisibility(isContentExpert && isOwner);
```
alansemenov: This is incorrect. A project owner doesn't need Content Expert role in order to see Settings. Should be `||` instead of `&&`.

alansemenov (changes requested):

alansemenov (approved):

Issue #2949, opened 2021-03-05
Responsive Image Selector


ashklianko:
@alansemenov commented on [Thu Mar 04 2021](https://github.com/enonic/lib-admin-ui/issues/1839)

In the gallery mode Image Selector always shows 3 images per row which responsively shrink along with selector's width. When the width is small, images are sometimes shrunk so much that it's hard to see them.

![image](https://user-images.githubusercontent.com/11612490/109961412-3fd44700-7cea-11eb-9c1d-329343df3c8c.png)

Fix it so that in this mode dropdown switches from 3 to 2 to 1 image per row depending on available width of the OptionsTreeGrid container.

_0-240, _240-360: 1 image per row (100% width per image):
![image](https://user-images.githubusercontent.com/11612490/109963460-ba05cb00-7cec-11eb-89c3-ea079c716555.png)

_360-540: 2 images per row (50% width per image):
![image](https://user-images.githubusercontent.com/11612490/109964784-6a280380-7cee-11eb-9c01-b7cff8b01912.png)

540 and higher: 3 images per row (33.3% width per image)
![image](https://user-images.githubusercontent.com/11612490/109963974-58922c00-7ced-11eb-888d-e6792599c6d3.png)

PR #2950, opened 2021-03-05
Responsive Image Selector #2949

ashklianko:


alansemenov (changes requested):
It seems a bit buggy when you resize the container while the image selector's dropdown is open.

For example I get this when expanding the Inspect panel with the dropdown open:

image

and this when I'm collapsing it:

Screenshot 2021-03-09 at 14 11 59

It might be that responsive range class is not being set on the container correctly, but we didn't notice it before.

We should also make sure that when container width changes we resize the images accordingly (change `size` parameter on image URL).

ashklianko:
Updated

alansemenov (approved):

PR #2977, opened 2021-03-16
Attachment is removed from content while it's still in use

ashklianko:
-Not deleting attachment if it is being used in other page components, in this case just cleaning property and saving; traversing page components configs to check attachment usage


sigdestad:
Just curious about this issue? Was it ever a real problem, and how can we actually prevent it from happening?

alansemenov:
@sigdestad yes, it's a real problem coming from a [support ticket](https://enonic.zendesk.com/agent/tickets/4454). we will prevent this by checking if the attachment is still in use by other components before trying to remove it from the content.

sigdestad:
Ok, so are people removing attachments via API then? Also, does that mean you somehow verify usage based on content type? AFAIK attachment API is not a content specific api, but designed to work at node level?

alansemenov:
Not API, Content Studio. It's about removing a reference to an attachment from a config of one of multiple parts referring to the same attachment which will then remove the attachment itself from the content. It will then become inaccessible in all other parts that had a reference to the same attachment.

alansemenov (approved):

sigdestad:
Ah.. I get it. Thanks for the clarification :-)

Issue #1763, opened 2020-04-29
Better feedback when applications used by a site get stopped or uninstalled


espen42:
If a site has an app missing/stopped/incompatible, in CS edit mode the missing app(s) are marked in red, indicating that they're missing. This is the only indication of why the preview fails.

The preview panel only says "Can't render the preview, check the logs", but there is no error message in neither server log nor browser console (at least in several cases lately, XP 7.2.0, CS 2.3.2).

And if editing a different page (i.e. not at the site root level), you only get the failed preview panel message, with no information anywhere.

alansemenov:
We should not force preview to fail when at least one app is missing, like we do now. The missing apps might not be relevant for rendering of preview at all.

alansemenov:
In the case of the missing/stopped apps, change the **second** part of the message in the preview panel to the following: _Some applications required for rendering are not available_

![image](https://user-images.githubusercontent.com/11612490/108051415-288b2d80-704b-11eb-9c00-97823b07f285.png)

PR #2978, opened 2021-03-16
Preview fails with little information, in CS edit mode on sites with …

ashklianko:
…apps missing #1763


alansemenov (changes requested):
This task is a bit more complicated than just changing the texts :)

There are two cases when we now fail content preview inside the Content Wizard:
1. When the preview fails to render because there's something with the application responsible for rendering.
2. When one of the applications assigned to the site is stopped/missing.

We now give the same error message ("_Please check logs for errors or open preview in a new window_") in both cases in the Content Wizard's preview panel and this is confusing. (Note that in the 2nd case preview actually works in the Content Grid and that is correct.)

What we should do is keep the old message for case 1 and show a different message (see task description) in case 2.

To test case 1, you can for example remove a [property](https://github.com/enonic/app-features/blob/master/src/main/resources/site/pages/main/main.js#L16) inside the main controller of the Features site. Deploying this will cause the site renderer to fail. Here we should still show the old message.

To test case 2, assign any application from the Market to the Features site, save, then uninstall the app in the Applications module. In this case we should show the new message in the preview panel:

image

ashklianko:
Updated

Issue #2973, opened 2021-03-15
Display Name input doesn't get focus in the Wizard


sgauruseu:
1. Select a project and open new `layer wizard`.
**Expected** - Display Name input in the should get focus.
**Actual** - language option filter input is focused.
![image](https://user-images.githubusercontent.com/3728712/111126700-0ae6b080-8584-11eb-9429-6c579fa175ca.png)

**BUT** - there is correct behavior if you open layer wizard without first selecting a project.

ashklianko:
Issue moved to [enonic/lib-admin-ui #1859](https://github.com/enonic/lib-admin-ui/issues/1859) via [**ZenHub**](https://www.zenhub.com/)

PR #3000, opened 2021-03-19
Display Name input should get focus in the Layer wizard #2973

ashklianko:


modules/lib/src/main/resources/assets/js/app/settings/wizard/panel/ProjectWizardPanel.ts
```diff
@@ -108,6 +108,12 @@ export class ProjectWizardPanel
         }
     }
 
+    giveInitialFocus(): void {
+       this.whenRendered(() => {
```
alansemenov, line 112: Maybe we should move this into the [parent class](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/js/app/wizard/WizardPanel.ts#L706), so that the display name input gets focus after render in all of the wizards? Makes total sense to me.

Issue #2063, opened 2020-07-14
Issue Details dialog blinks after items in an issue have been changed


sgauruseu:
1. Open Task Details dialog or Request details dialog.
2. Add new item (remove an item)

BUG: Dialog blinks 4 times after the updating.

![image](https://user-images.githubusercontent.com/3728712/87401894-71659a00-c5c3-11ea-90d8-de5f5aeedce1.png)

ashklianko:
ListIssuesRequest - is invoked up to 10 -14 times on adding/removing item!!!

PR #3006, opened 2021-03-23
Issue Details dialog blinks several times (4-5) after adding or remov…

ashklianko:
…ing items #2063

-Blinking was due to reload of masked issue list dialog. Fixed so that issue list dialog is reloaded only when switching back to it.


alansemenov (changes requested):
Nice fix but we used to have a number of items in brackets inside the "Items" tab. Now it's gone and appears only when you add or remove the items. Same with the "Publish" button on that tab.

image

image

ashklianko:
Updated

PR #3017, opened 2021-03-26
Form inside an option-set's unselected option is not cleared on save

ashklianko:
-Using form view clean() method to remove unwanted values


alansemenov (changes requested):
The purpose of the code you deleted (or at least some part of it) was to cleanup empty property sets for non-selected options inside an option-set from the data object before save. As you can see in the screenshot below, there are two options selected in the option-set but empty objects are stored for all 4 options:

image

alansemenov (approved):
Nice cleanup!

PR #3116, opened 2021-05-03
Move content-related classes from lib-admin-ui

alansemenov:
Requires https://github.com/enonic/lib-admin-ui/pull/1935


ashklianko (approved):
Seems to be working fine after move

Issue #2078, opened 2020-07-20
Upload icon remains visible after uploading more items than allowed by AttachmentUploader config


sgauruseu:
1. Open new wizard for Attachments content (min 2. max=4, for example)
2. Click on Upload button and select 5 images in the modal dialog.

**BUG: Upload button should not be displayed, because max number of item is 4:** 

 
![image](https://user-images.githubusercontent.com/3728712/87936685-eccfbb80-ca9b-11ea-82f1-460620f04cb8.png)

**Refresh the wizard in browser page. The icon gets not visible.**

**Case 2:** 
- reopen content that was used in the first case.
- click on remove icon, remove an image and upload another image .

BUG: only one item remains in the wizard after replacing an image

![image](https://user-images.githubusercontent.com/3728712/87937167-e1c95b00-ca9c-11ea-83fa-91787fe77fc2.png)

ashklianko:
Reproduced very hard from time to time, issue in async update events coming before item is saved and rewriting attachemnts property array

PR #3120, opened 2021-05-04
Attachments wizard - upload icon remains visible after uploading item…

ashklianko:
…s more then allowed in configuration #2078

-Delaying adding/removing items to AttachmentUploader's property array until attachment is created/removed, otherwise intermediate update event from it might overwrite property array with stale values
-Removed redundant code skipping update


alansemenov (approved):
Nicely solved

Issue #2151, opened 2020-08-06
Apply button in the Inspect panel shouldn't save content when no changes have been made


sgauruseu:
1. Open a published(ready to publish) site with a text component. Open Page Component View, click on the text component: 
![image](https://user-images.githubusercontent.com/3728712/89524429-3c78eb80-d7ed-11ea-8aa5-28eac39d97ca.png)

2.  InspectionsPanel is opened, click on Apply button:
![image](https://user-images.githubusercontent.com/3728712/89524346-10f60100-d7ed-11ea-8dae-bd5826dd0656.png)

**Bug**: 
1) Incorrect notification message appears after  clicking on Apply  button :`Item is saved`
2) New content version is created.
3) Published content gets `modified` (work in progress)

**Expected**: 
-This message should appear :  `No changes to apply`
- New version should not be created.
- content;s status should not be changed

ashklianko:
Issue not reproduced (we check for changes before saving), I've made only small refactoring. However some improvements could be made: apply button should be enabled only when there are changed in inspected component. That will take time to implement so only refactoring for now. @sgauruseu please confirm that not reproduced

PR #3122, opened 2021-05-05
Apply button in the Inspect panel shouldn't save content when no chan…

ashklianko:
…ges have been made #2151

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -243,7 +243,6 @@ export class ContentWizardPanel
         }
 
         this.contentParams = params;
-
         this.loadData();
 
         this.isContentFormValid = false;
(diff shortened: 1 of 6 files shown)


alansemenov (changes requested):
It can be reproduced quite easily, actually. Just keep selecting the text component and clicking Apply. (Why we have Apply there at all is another issue.) Good thing that it doesn't create a new version because I guess there's some smartness on the server that checks whether content has changed, but we shouldn't try to save it on the client either.

It also loses component selection after save, which is another bug...

https://user-images.githubusercontent.com/11612490/117281254-bf55d200-ae63-11eb-9500-7aed7cd6a091.mov

ashklianko:
Updated

alansemenov (approved):

sigdestad:
Did we manage to remove/disable apply button if nothing was changed?

ashklianko:
Not yet, that would be a separate issue, however we've fixed some general issue with saving content

Issue #3091, opened 2021-04-28
Page Template wizard - Save button remains enabled after selecting a controller in unnamed template


sgauruseu:
1. Open new page template wizard(in Features for example)
2. Do not fill the name input, just select a controller - `main region`
3. 
3. **BUG** - the template is saved but Save button remains  enabled: 
![image](https://user-images.githubusercontent.com/3728712/116375639-2518dd00-a818-11eb-8d1f-188aac33492c.png)

PR #3127, opened 2021-05-05
Page Template wizard - Save button remains enabled after selecting a …

ashklianko:
…controller in unnamed template #3091

-When rendering form items we might spawn default empty property arrays for them to store data. In a specific case with a new content where it's data is not saved yet with default/empty values we can trigger content's page create/update that will change page data only on a backend, after that we receive and process content update event with content that doesn't have default data values set yet; form items are not getting updated because on update event when comparing new data and current form data we treat empty property and null as equal and form data update is skipped, however nulls are not ignored in other checks and comparing persisted item data with current form data will see the change and enable save button
-Fixing this by handling null and empty property as not equal


ashklianko:
Potentially some tests may fail

alansemenov (approved):

Issue #3096, opened 2021-04-29
Insert macro dialog does not retain focus when switching tabs/windows


thomasheartman:
When inserting a macro (via the insert macro dialog), if I:
	1. Focus one of the input fields
	2. Change tabs or switch windows
	3. Switch back to the tab/window with the insert macro dialog

Then: the focus (on the input) field is not retained.

An additional oddity is that the HtmlArea (as visible in the background) loses its 'active state style' when changing windows, but regains it when I come back. Whether this is intentional or not, I don't know. It doesn't matter much, but it's interesting.

https://user-images.githubusercontent.com/17786332/116522435-7d61e480-a8d5-11eb-8f37-b33c87e8d17d.mp4

PR #3130, opened 2021-05-06
Insert macro dialog does not retain focus when switching tabs/windows…

ashklianko:
… #3096

-Most of HtmlArea dialogs are backed up by hidden original cke dialog that prevents losing focus, but MacroModalDialog is our original dialog so by some internal cke logic it's editor loses focus on clicking outside of dialog
-Fixed by using internal cke focusManager functionality that prevents losing focus


alansemenov (approved):
Niiiiiiice

Issue #1967, opened 2020-06-17
User is switched to default context after current context/project is deleted


sgauruseu:
1. Install `Superhero`
2. Create an user with the next roles: 

![image](https://user-images.githubusercontent.com/3728712/84881559-b6ff7780-b096-11ea-9291-5301494ea7a6.png)

3. Do log in with the user. `SuperHero Blog` context is loaded - OK
4. Switch  to browser with SU and delete  `SuperHero Blog`  project
5. Switch to browser with user with Author role

BUG: default context is loaded for the user with Author role

![image](https://user-images.githubusercontent.com/3728712/84883375-1bbbd180-b099-11ea-8750-30b34bc29a4b.png)

PR #3170, opened 2021-05-17
User is switched to default context after current context/project is …

ashklianko:
…deleted #1967

-Handling deleting of the only available project for a user:
1. Introduced new project context state NOT_AVAILABLE
2. Added listeners for a case when single project in use becomes unavailable
3. Cleaning TreeGrid in that case and disabling toolbars and buttons

### DIFF
--- a/modules/app/src/main/resources/assets/js/main.ts
+++ b/modules/app/src/main/resources/assets/js/main.ts
@@ -601,6 +601,10 @@ function initProjectContext(application: Application): Q.Promise<void> {
             return Q(null);
         }
 
+        if (projects.length === 0) {
+            ProjectContext.get().setNoProjects();
+        }
+
         ProjectSelectionDialog.get().open();
 
         return Q(null);
--- a/modules/app/src/main/resources/assets/js/main.ts
+++ b/modules/app/src/main/resources/assets/js/main.ts
@@ -602,7 +602,7 @@ function initProjectContext(application: Application): Q.Promise<void> {
         }
 
         if (projects.length === 0) {
-            ProjectContext.get().setNoProjects();
+            ProjectContext.get().setNotAvailable();
         }
 
         ProjectSelectionDialog.get().open();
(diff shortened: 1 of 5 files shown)


modules/lib/src/main/resources/assets/js/app/project/ProjectContext.ts
```diff
@@ -54,8 +63,24 @@ export class ProjectContext {
...
+    private notifyNoProjectsAvailable() {
+        this.noProjectsAvailableListeners.forEach((handler: () => void) => {
+            handler();
+        });
+    }
 }
 
 enum State {
-    INITIALIZED, NOT_INITIALIZED
+    INITIALIZED, NOT_INITIALIZED, NO_AVAILABLE
```
alansemenov: Let's rename it to "`NOT_AVAILABLE`" (as in "_context is not available_").

modules/lib/src/main/resources/assets/js/app/project/ProjectContext.ts
```diff
@@ -35,6 +37,13 @@ export class ProjectContext {
         this.notifyProjectChanged();
     }
 
+    setNoProjects() {
```
alansemenov: `setNotAvailable`

alansemenov (changes requested):

modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -601,6 +601,10 @@ function initProjectContext(application: Application): Q.Promise<void> {
             return Q(null);
         }
 
+        if (projects.length === 0) {
+            ProjectContext.get().setNoProjects();
```
alansemenov: This one should be renamed to `setNotAvailable` as well

ashklianko: Sorry, idea didn't show this invocation

alansemenov (approved):

Issue #3058, opened 2021-04-16
The test for contentStudioApp and contentStudioProjects generate the same artifacts.


jsi:
We need to separate these, so we can keep track of them since the latest always overwrite the older ones.

Basically, we are running the two tests:
`testContentStudioProjects`
`testContentStudioApp`

Both generate artefacts (zip files) with the same name and folder.  We need something that separates them.

In the future, we may want to split the tests into further parts, so the solution needs to be made so that each new test can have its own artifact.

jsi:
I tried some different solution to this, but I do not understand gradle well, so my solutions didn't work.  This must be assigned to a gradle expert.

jsi:
I can finally confirm that we have a working solution! :)

PR #3173, opened 2021-05-18
The test for contentStudioApp and contentStudioProjects generate the …

edloidas:
…same artifacts. #3058

Replacing session variable with command line property `-PartifactAppendix=...`.


jsi (approved):
This looks promising.

Issue #3178, opened 2021-05-19
Error message should not be shown before new X-data form is validated


sgauruseu:
1. Open new wizard folder under "Features" site
2. Fill in the name and click on Save
3 Click + icon next to "X-data (folders only)" inside the Content Wizard form
**BUG** - This field is required should not be displayed before the saving 

![image](https://user-images.githubusercontent.com/3728712/118797996-d35af400-b8a5-11eb-8ef2-8c76de1cfb97.png)

PR #3197, opened 2021-05-26
Error message should not be shown before new X-data form is validated…

ashklianko:
… #3178

-Controlling x-data validation errors displaying: showing only when content is saved and for optional x-data only when x-data present in persisted item data
-Instantiating form context once
-Setting cloned persisted content to form context instead of same object that is set to content data forms, otherwise this object's data and x-data is modified in wizard data forms

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -620,7 +620,6 @@ export class ContentWizardPanel
         return super.saveChanges().then((content: Content) => {
             const persistedItem = content.clone();
             if (liveFormPanel) {
-                this.initFormContext(persistedItem);
                 this.liveEditModel.setContent(persistedItem);
 
                 if (this.pageEditorUpdatedDuringSave) {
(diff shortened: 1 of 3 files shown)


modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -2038,17 +2037,19 @@ export class ContentWizardPanel
 
             this.getWizardHeader().resetBaseValues();
 
-            return content;
-        }).then((content: Content) => {
-            this.formState.setIsNew(false);
-            this.contentWizardStepForm.validate();
-            this.xDataWizardStepForms.validate(false, true);
-            this.displayValidationErrors(!this.isValid());
-
```
alansemenov, line 2047: You have moved this to `postUpdatePersistedItem` but you are not calling it...

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
ashklianko: It is being called in base WizardPanel.ts

ashklianko: Also I'd mention that post save/update logic is spread into multiple methods/then clauses - one of the candidates to refactor and to be moved in some single place if possible

ashklianko:
Forgot to mention this dependency : https://github.com/enonic/lib-admin-ui/pull/1962

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
alansemenov: Right

alansemenov (approved):

Issue #3132, opened 2021-05-07
Hide "Reset" button in the Content Wizard if content item doesn't have a parent


alansemenov:
When current layer has a parent project/layer but current content item doesn't have a corresponding item in that parent project/layer (for example it was deleted), we should not show "Reset" button since there's nothing to reset to.

PR #3216, opened 2021-06-03
ui-tests issue#3148

sgauruseu:
Hide "Reset" button in the Content Wizard if content item doesn't have a parent #3132


jsi (approved):
Nice!

Issue #3121, opened 2021-05-05
Support disabling of content path transliteration


alansemenov:
We transliterate content path in the Content wizard, so that "`/øåæ`" becomes "`/oaae`", "`/рыба`" becomes "`/ryba`" and so on. 

We will keep this as a default behaviour but implement a config (in `com.enonic.app.contentstudio.cfg`) called `contentWizard.allowPathTransliteration` which, when set to `false`, will keep the path as it is.

We already [do this](https://github.com/enonic/app-contentstudio/blob/master/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts#L703) for media content, expand this to always do the same if the config is "on".

PR #3218, opened 2021-06-04
Support disabling of content path transliteration #3121

ashklianko:
-Added path.transliteration config param to control transliteration


alansemenov (changes requested):
Let's rename config property to `contentWizard.allowPathTransliteration`, and `CONFIG` object's property to `allowPathTransliteration`, to make it consistent with other config options.

ashklianko:
Updated

alansemenov (approved):

Issue #3215, opened 2021-06-03
Deprecate Image component


alansemenov:
1. Remove "Image" from the Insert panel
2. Remove "Image" from the context menus of the Page Editor and Page Components view
3. Create new literal resource `field.rich.text` and use it instead of `field.text` in `Insertables.ts`. In `phrases.properties` set it to "_Rich Text Editor_".
4. Create new literal resource `field.rich.text.help` and use it instead of "`field.text.help`" in `Insertables.ts`. In `phrases.properties` set it to "_Write text and embed images_".
5. Remove `field.text` and `field.text.help` from all the phrase files.

![image](https://user-images.githubusercontent.com/11612490/120656471-ee239000-c483-11eb-9e13-24bcbe10afce.png)

![image](https://user-images.githubusercontent.com/11612490/120656576-05627d80-c484-11eb-93e3-d5d2fa228e62.png)

Insert panel after the changes:
![image](https://user-images.githubusercontent.com/11612490/120657121-81f55c00-c484-11eb-9c62-4d46b0bec170.png)

PR #3219, opened 2021-06-04
Deprecate Image component #3215

ashklianko:
-Removed image from insert panel and context menus, but left in inspect panel to allow inspect existing pages with image component
-Updated text entries for html editor


alansemenov (changes requested):
Should probably get rid of the `field.image.*` text resources as well

ashklianko:
> field.image.

Used in ImagePlaceholder and ImageModalDialog

alansemenov (approved):

PR #3225, opened 2021-06-09
Use cached response of IsAuthenticatedRequest

ashklianko:
-Removing saved LoginResult values because now classes will be using auth request cache

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/security/UserAccessWidgetItemView.ts
+++ b/modules/lib/src/main/resources/assets/js/app/security/UserAccessWidgetItemView.ts
@@ -32,8 +32,6 @@ export class UserAccessWidgetItemView
 
     private bottomEl: AEl;
 
-    private loginResult: LoginResult;
-
     public static debug: boolean = false;
 
     constructor() {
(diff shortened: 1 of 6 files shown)


alansemenov (changes requested):
It's definitely a great start. The number of `isAuthenticated` requests got down from 4 to 1 in the Content Grid. 

Where we do still have a problem though is when several requests to `isAuthenticated` are sent _in parallel_ (almost simultaneously). In this case none of them has cached the response yet and as a result we are still sending X identical requests.

To test this: add several text components to a page, save the content and reload. `TextComponentView` makes a call to `HTMLAreaHelper.isSourceCodeEditable` which sends `isAuthenticated` request. Since these requests are sent pretty much simultaneously for each text component, we end up with as many requests as the text components.

To solve this: we need to "store" the request from the moment it's sent and until we have received the response, and don't send a new request if there's already one in progress. We have done something similar in [StylesRequest](https://github.com/enonic/app-contentstudio/blob/master/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/styles/StylesRequest.ts) but here's it's even simpler since we don't need to group requests by ContentId.

ashklianko:
Yes, thought about that, will implement then

ashklianko:
Updated in lib-admin-ui!

Issue #3226, opened 2021-06-09
Responsive Widget panel in docked mode


alansemenov:
When browser window is wide enough to open the Widget panel in docked mode, layout of the ContextSplitPanel is as follows:
* Widget Panel width is 280px (minus splitter)
* Remaining width is split between Content Grid and Preview Panel with "golden ratio" (38% / 62%)

Change the layout behaviour so that in docked mode the Widget Panel is responsive and is in the same "golden ratio" with the Preview Panel.

Example:
* Browser window is 2000px wide.
** Content Grid takes 760px (38% of 2000px).
** Remaining 1240px should be split between the Preview Panel and the Widget Panel with 62:38 ratio
*** The Preview Panel will take 768px (62% of 1360px) and the Widget Panel will take the remaining 472px
** If the Widget panel is collapsed, entire width (2000px) will be split between the Content Grid and the Preview Panel with 38:62 ratio (like now)

Keep 280x px as min width of the Widget panel.

![image](https://user-images.githubusercontent.com/11612490/121339242-902cf780-c91e-11eb-864e-0455778be0c9.png)

PR #3234, opened 2021-06-16
Responsive Widget panel in docked mode #3226

ashklianko:
-Docked panel put inside SplitPanel with live/statistics panel to occupy 68% of available width
-LivePanel related code moved to ContentWizardPanel


ashklianko:
Commited some fixes for wizard and not only

alansemenov (approved):
Works as a charm :)

Issue #3227, opened 2021-06-10
Implement managed thumbnail request


alansemenov:
When a folder with images is expanded in the Content Grid, browser sends simultaneous requests for all content icons that are added to the Grid. XP will start processing all of the original images simultaneously and if the originals are hi-res, then the server might struggle.

Let's try to solve this by sending a _managed_ request to the server instead of simply adding `<img>` element with url of an icon to the DOM.

Implement a new viewer class (deriving from `ContentSummaryAndCompareStatusViewer`) which will do the following:
* Originally create `namesAndIconView`'s .`iconImageEl` with a placeholder icon (this is already supported in the `ImgEl` class). This will most likely require refactoring of `NamesAndIconView` (or implementing a new class deriving from it) 
* Implement a promise in `resolveIconUrl` of the new viewer class which will send a GET request (implement a new class for this) to the image url
* When response arrives, check the status code. If it's 200, resolve the promise and set src of the viewer's `iconImageEl` element. If it's anything else, keep thumbnail in the image.
* We will later implement a queue to reload the failed loads.

PR #3243, opened 2021-06-21
Implement managed thumbnail request #3227

ashklianko:
-Sending XHR request to control image displaying in grid
-Caching image request for specified image id
-Showing image placeholder until 200 server response received
-Converting image received to a base64 string and setting to an Img tag src to prevent second image request (img src tag doesn't use previously received xhr response)
-In future failed request to be managed specifically


alansemenov (changes requested):
* Uploading an image directly via New Content dialog now fails with these errors:
 
image
 
* We need to add some kind of animation indicating that the image is being loaded. Right now there's not even an empty placeholder which would be replaced by the loaded image. I suggest showing a small spinner for each image that is being loaded, and making sure that the spinner container is of the same width/height as image thumbnail so that display-name/path won't move once the image is loaded.
Note how the the top 3 images (these are hi-res) in the expanded folder take their time to load with no indication whatsoever, then move the name/path block to the right:

chrome-capture

alansemenov:
Images I used for testing: https://drive.google.com/drive/folders/1DjvNbviDzPM5NOIr9CrOokFzrmytaU7U?usp=sharing

alansemenov (approved):
Nice

Issue #3248, opened 2021-06-22
Content Wizard - display name and path are cleared after opening a content


sgauruseu:
**Case 1**
- Open an existing folder
**BUG: Display name and path inputs are cleared** 

**Case 2:** 
- Open new site wizard 
- Fill in the name input 
**BUG- Path is not generated** 

![image](https://user-images.githubusercontent.com/3728712/122921173-dba5c380-d36a-11eb-97ce-5de91162211b.png)

PR #3252, opened 2021-06-24
Content Wizard - display name and path are cleared after opening a co…

ashklianko:
…ntent #3248


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/2017

alansemenov (approved):

Issue #3253, opened 2021-06-24
B/I/U buttons are missing in HTML Editor's fullscreen mode


alansemenov:
Another issue (in addition to https://github.com/enonic/app-contentstudio/issues/3235) caused by https://github.com/enonic/app-contentstudio/issues/3074:

We used to hide B/I/U in the "default" toolbar view, but show them in the "Fullscreen" mode since we have more space there. Now they are gone from there as well:

![image](https://user-images.githubusercontent.com/11612490/123244543-50e8d400-d4e4-11eb-8211-8ebba64ceafc.png)

The only case when they should not be shown in the Fullscreen mode is if they were specifically excluded by the `<exclude>` config.

PR #3261, opened 2021-06-25
B/I/U buttons are missing in HTML Editor's fullscreen mode #3253

ashklianko:
-Adding those only if thy are not excluded in config

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
@@ -975,6 +975,20 @@ class HtmlEditorConfigBuilder {
             this.includeTool('Fullscreen');
         }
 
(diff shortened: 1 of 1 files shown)


modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
```diff
@@ -975,6 +975,20 @@ class HtmlEditorConfigBuilder {
...
+            }
+
+            if (!this.isToolDisabled('Italic')) {
+                this.includeTool('Italic');
+            }
+
+            if (!this.isToolDisabled('Underline')) {
+                this.includeTool('Underline');
+            }
+        }
```
alansemenov: `['Bold', 'Italic', 'Underline'].forEach((tool: string) => !this.isToolDisabled(tool) && this.includeTool(tool));`

alansemenov (changes requested):

alansemenov (approved):

Issue #3260, opened 2021-06-25
Implement a new way of displaying inbound references in the Delete dialog


alansemenov:
Delete content dialog sends ResolveDependenciesRequest to fetch references from/to the content that is about to be deleted. If there are inbound references, we show the toast popup at the bottom and also show a link underneath the item to open dependencies in a new tab.

The problem is that we only do that for the main item(s), but not for their child items. This can be easily tested by trying to delete "`Ref*`" items under `/features/input-types/references/` (which will give you the warning about incoming dependencies) and then the entire "Features" site (which will NOT give the warning even though "`Ref*`" items are in the list as well).

Things to do:
1. Fetch references for the child items as well when sending `ResolveDependenciesRequest`. This can be tricky because the list of child items is lazy loaded, but we need to know inbound dependencies of each and every item that is about to be deleted, not only those that are loaded so far.
2. Once the list of inbound dependencies is fetched, traverse it and remove ids that are in the list of items to be deleted. i.e. if item A is referenced by item B, but both A and B are in the list, then will not notify about this since both A and B will be deleted.
3. Change the way we notify about incoming dependencies:
* Remove the toast popup (and the literal resource used in it)
* Display a message at the top of the list like we do in the Publish dialog: "```<icon> <b>Inbound dependencies!</b> One or more items are referenced from another content```" (add the new literal resource)
* The icon is `icon-link` which we already use in the Dependencies widget 
* For each item in the list (both main items and child items) replace the content icon with the `icon-link`
* Add a tooltip to the icon: "Click to show the inbound references"
* Open the tab with dependencies on icon click (same as what we do when clicking the link)
* Remove the "`Inbound dependency: X`" link (and the literal resources) that we now show for each item

![image](https://user-images.githubusercontent.com/11612490/123412853-5c0a3580-d5b2-11eb-88b6-cde5a3135dee.png)

PR #3272, opened 2021-06-30
Implement a new way of displaying inbound references in the Delete di…

ashklianko:
…alog #3260


modules/lib/src/main/resources/i18n/dialogs.properties
```diff
@@ -109,6 +106,9 @@ dialog.rename.label=New name
 dialog.confirm.resetInheritance=Resetting this content item will revert all the local changes. This action cannot be undone.
 dialog.project.delete.confirm.subheader=You are about to delete this project with all of its content. This action cannot be undone.
 dialog.layers.notAvailable=No access to the project
+dialog.delete.hasInbound.part1=Inbound dependencies!
```
alansemenov, line 109: Can you rename `dialog.delete.dependencies` to `dialog.delete.hasInbound.part1` (and add "!") in all other language files as well, so that we could reuse the existing translations of this phrase?

alansemenov (changes requested):
One more thing (that should've been mentioned in the original task description): we should change the order of child items so that the ones with the inbound references are shown on top of the list (just like we do with invalid/in progress items in the Publish dialog).

alansemenov:
There's also some Codacy issues:

image

ashklianko:
Updated

alansemenov (approved):
One more thing, that should've been mentioned in the original task description

Issue #2964, opened 2021-03-10
Switch from unpackDevResources to copyDevResources


rymsha:
same as https://github.com/enonic/lib-admin-ui/issues/1848 but for lib-contentstudio

More discription in
https://github.com/enonic/xp-gradle-plugin/issues/62

PR #3314, opened 2021-07-16
Build config optimisation

edloidas:
Fixed the tasks dependencies and execution order between `lib-admin-ui`  and `app-contentstudio`.
Refactored `clean` task definitions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,13 +7,16 @@ plugins {
     id 'org.aim42.htmlSanityCheck' version '1.1.6'
 }
 
+allprojects {
+    apply from: "$rootDir/gradle/env.gradle"
+    apply from: "$rootDir/gradle/node.gradle"
+    apply from: "$rootDir/gradle/lib-admin.gradle"
+}
+
 subprojects {
     apply plugin: 'java'
     apply plugin: 'com.enonic.defaults'
     apply plugin: 'com.enonic.xp.base'
-    apply from: "$rootDir/gradle/env.gradle"
-    apply from: "$rootDir/gradle/node.gradle"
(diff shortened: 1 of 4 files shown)


ashklianko (approved):
Couldn't fully check it because original issue is reproducing occasionally

build.gradle
```diff
@@ -7,13 +7,16 @@ plugins {
     id 'org.aim42.htmlSanityCheck' version '1.1.6'
 }
 
+allprojects {
```
rymsha: why root needs it?

gradle/lib-admin.gradle
```diff
@@ -25,7 +66,12 @@ ext {
     hasLibAdminUi = this.&hasLibAdminUi
 }
 
-task flush( type: Delete, dependsOn: clean ) {
-    description = 'Clean the project from built sources and dependencies'
-    delete '.xp'
+afterEvaluate {
+    if ( hasTask( 'clean' ) )
+    {
+        task flush( type: Delete, dependsOn: clean ) {
```
rymsha: flush is rudimental. clean should be sufficient. this is likely the reason we still observe the problem

edloidas:
@rymsha Yup, my bad. I have tested everything again. And it seems the original issue was actually fixed by you already in #2964. 
Git failed to update my branch and I reproduced the problem on a stale code (when there were no `*-dev-resources` jars). 
As for the `flush` task, it wasn't the source of the original problem.
I've updated the code, keeping only small optimizations, that are still relevant. Build order problem is gone already.

rymsha (approved):

alansemenov (approved):

Issue #3368, opened 2021-08-02
Permissions are incorrectly set in Edit Permissions dialog


sgauruseu:
1. Create a folder and an user with `Content Manager App` and  `Administration Console Login`
2. Select the folder, open `Edit Permissions` dialog and select `Full Access` in  Access Selector, click on Apply.

![image](https://user-images.githubusercontent.com/3728712/127859615-503098db-43f2-4b85-a904-e3ab24947573.png)

3.  Reopen `Edit Permissions` dialog

**BUG**: ### Incorrect permissions are displayed after reopening the modal dialog 

![image](https://user-images.githubusercontent.com/3728712/127859459-b398b212-8a09-4a9f-a2dd-d971312ce055.png)

PR #3403, opened 2021-08-12
Permissions are incorrectly set in Edit Permissions dialog #3368

ashklianko:


modules/lib/src/main/resources/assets/js/app/view/AccessControlEntryView.ts
```diff
@@ -136,37 +136,34 @@ export class AccessControlEntryView
...
-        return this.canPublish(allowed) && allowed.length === 5;
+        return this.canPublish(allowed) && allowed.length === 1;
     }
 
     private static isFullAccess(allowed: Permission[]): boolean {
-        return this.canPublish(allowed) &&
-               allowed.indexOf(Permission.READ_PERMISSIONS) >= 0 &&
+        return allowed.indexOf(Permission.READ_PERMISSIONS) >= 0 &&
                allowed.indexOf(Permission.WRITE_PERMISSIONS) >= 0;
     }
```
alansemenov, line 159: Hmm, why should _only_ `READ_PERMISSIONS` + `WRITE_PERMISSIONS` mean `FULL ACCESS`? FA is when **all** available permissions are applied. In the current code it was achieved with `this.canPublish(allowed)` + these two, where `this.canPublish` is a composite check for the other 5.

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/view/AccessControlEntryView.ts
ashklianko: The thing is that getPermissionsValueFromAccess() considers access is FULL when just these two READ_PERMISSIONS + WRITE_PERMISSIONS present

ashklianko: Ah, I see now why - there was no `break;` before in `SWITCH` statement. Then yes, will fix in other way

Issue #3407, opened 2021-08-13
Custom Selector incorrectly loads options


ashklianko:
@sgauruseu commented on [Wed Aug 11 2021](https://github.com/enonic/lib-admin-ui/issues/2086)

**Case 1:**
1. Select Features , open wizard for new Custom Selector 
2. type `test` in the options filter input
 
![image](https://user-images.githubusercontent.com/3728712/129028213-d5e1def4-8996-4495-bc9f-f80904a1ece8.png)

3. Clear the input via the Backspace key
**BUG**: input is cleared, but options are not loaded 

![image](https://user-images.githubusercontent.com/3728712/129028402-e795a2a9-f447-431c-a19c-bae3218c66bd.png)

**Case 2:**
1. Select Features , open wizard for new Custom Selector 
2. Copy `Option number 1` in buffer
3. Press Ctrl+P and Insert `Option number 1` to the options filter Input
![image](https://user-images.githubusercontent.com/3728712/129031953-d8bfc929-2ffc-411c-895c-5cdc3a914efa.png)

  Expected option is filtered - OK 
4. Press backspace and update the value in the input : 

![image](https://user-images.githubusercontent.com/3728712/129032196-c985688f-45ad-4c1a-b9a0-e54f7e7ab9f9.png)

**BUG**- `Option number 2` is not filtered,   `No matching items is displayed`

**Expected**:  

![image](https://user-images.githubusercontent.com/3728712/129032520-3bacdf77-498b-4472-a01b-64f1a1f774d5.png)

Issue #3201, opened 2021-05-28
Paging doesn't work in Custom Selector


alansemenov:
Custom Selector request doesn't include `start` parameter, therefore paging won't work. Got broken around v3.1

PR #3408, opened 2021-08-13
Custom Selector - options are not loaded after a non-existing option …

ashklianko:
…is entered into the filter input #3407

-Request is paginated and might have query params, thus caching should not be available

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/customselector/CustomSelectorComboBox.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/customselector/CustomSelectorComboBox.ts
@@ -15,6 +15,18 @@ export class CustomSelectorComboBox
         super(builder);
     }
 
+    protected reload(inputValue: string): Q.Promise<any> {
+        if (this.getLoader().isLoaded() && this.getLoader().getSearchString() === inputValue) {
+            return this.getLoader().search(inputValue);
+        }
+
+        this.getLoader().setSearchString(inputValue);
+
+        return this.getLoader().load().then(() => {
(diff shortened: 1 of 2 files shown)


alansemenov (changes requested):
This fix will break paging and Custom Selector will never load more than 10 items, something that was fixed in https://github.com/enonic/app-contentstudio/issues/3201 (`start` parameter never gets appended to service url).

So while the fix itself is correct, we should find out why the paging got broken in the first place.

To test, use content type called "Color picker" in the Features app that has 16 items in its Custom Selector.

ashklianko:
Updated

modules/lib/src/main/resources/assets/js/app/inputtype/customselector/CustomSelectorComboBox.ts
```diff
@@ -15,6 +15,18 @@ export class CustomSelectorComboBox
...
+    protected reload(inputValue: string): Q.Promise<any> {
+        if (this.getLoader().isLoaded() && this.getLoader().getSearchString() === inputValue) {
+            return this.getLoader().search(inputValue);
+        }
+
+        this.getLoader().setSearchString(inputValue);
+
+        return this.getLoader().load().then(() => {
+            return this.getLoader().search(inputValue);
+        });
```
alansemenov: Can we replace `this.getLoader()` with a `const` please?

ashklianko: Done

alansemenov (approved):

Issue #3422, opened 2021-08-17
Delete/Archive dialog - don't mark items with dependencies that will be deleted


alansemenov:
Depends on https://github.com/enonic/app-contentstudio/issues/3359

When displaying the list of items to be deleted in the Delete modal dialog, we notify about items that have inbound dependencies. However, if these dependencies are also in the list of items to be deleted, then we don't need to do that since both the item and its references will be deleted. I.E. if item A is referenced by item B, but both A and B are in the list, then we do not need to notify about this since both A and B will be deleted.

alansemenov:
#3359 is ready

alansemenov:
`resolveForDelete` now returns an array of inbound reference ids in `inboundDependencies`.

![image](https://user-images.githubusercontent.com/11612490/131339682-c816bb12-0286-45c1-a919-7d466cb129f8.png)

We should call `resolveForDelete` for the entire list of items in the dialog (both the main ones and their children) and show the dependency icon only for those items whose `inboundDependencies` are NOT on this list.

PR #3491, opened 2021-09-01
Delete/Archive dialog - don't mark items with dependencies that will …

ashklianko:
…be deleted #3422

-Using new api that returns resolved inbound dependencies for items to delete


alansemenov (approved):
Works as a charm, great job!

Issue #3476, opened 2021-08-30
B/I/U disappeared from Text component's toolbar


alansemenov:
B/I/U buttons should always be available inside Text component. It's only in HtmlArea input type where they are removed by default and can be added via config.

![image](https://user-images.githubusercontent.com/11612490/131364696-9570183f-1202-4aaa-bcfd-84dc78ded952.png)

Issue #3192, opened 2021-05-25
HtmlArea config - include config does not work with `<exclude>*</exclude>`


sgauruseu:
1, Create the next htm;Area type
```
<input name="myhtmlarea" type="HtmlArea">
  <label>My HtmlArea</label>
  <default><h3>Enter description here</h3></default>
  <config>
    <exclude>*</exclude> 
    <include>JustifyLeft JustifyRight | Bold Italic</include> 
    <allowHeadings>h2 h4 h6</allowHeadings> 
  </config>
</input>
```
2. Open newwizard for this content 

**BUG** 
**Expected** - Only 2 items should be present in the toolbar
![image](https://user-images.githubusercontent.com/3728712/119517620-71f0c480-bd80-11eb-870e-0152a08c3a9c.png)

**Actual** - the toolbar is empty 

![image](https://user-images.githubusercontent.com/3728712/119517544-60a7b800-bd80-11eb-8908-bd7b1dea179b.png)

PR #3517, opened 2021-09-09
ui-tests issue#3499

sgauruseu:
verifies  2 bugs
B/I/U disappeared from Text component's toolbar #3476
HtmlArea config - include config does not work with <exclude>*</exclude> #3192


jsi (approved):
Nice!

PR #3598, opened 2021-09-27
Filtering by allowContentType breaks when ContentSelector is inside a nested structure

alansemenov:


alansemenov:
Requires https://github.com/enonic/lib-admin-ui/pull/2146

alansemenov:
There are three places the bug appears now: x-data on a site, a part on a site or a layout on a site. It's important that it's _on_ a site, if it's any content _inside_ a site, then the bug won't appear. 

Sample form for testing below (should be inserted into an x-data or part or layout schema). Two identical selectors are used for comparison, because the bug only happens for a selector inside a set (a field-set or an option-set or an item-set). So the idea is that before the bug fix the nested selector won't filter the items correctly, while after the fix both selectors must display the same data-set.

```xml
  <form>
    <field-set>
      <label>Navigation</label>
      <items>
        <input name="segments" type="ContentSelector">
          <label>Select (and order) segments</label>
          <occurrences minimum="0" maximum="0"/>
          <config>
            <allowContentType>landing-page</allowContentType>
          </config>
        </input>
      </items>
    </field-set>
    <input name="segments2" type="ContentSelector">
      <label>Select (and order) segments 2</label>
      <occurrences minimum="0" maximum="0"/>
      <config>
        <allowContentType>landing-page</allowContentType>
      </config>
    </input>
  </form>
```

ashklianko (approved):

Issue #3609, opened 2021-09-28
Error in log when archiving items


ashklianko:
When archiving items error appears in server log, related to fetching archived content from /content root. Seems to be related with server events handling

PR #3630, opened 2021-10-06
Error in log when archiving items #3609

ashklianko:
-Problem with archiving items is that after triggering archiving we receive update->rename(not always)->move events, and attempt to fetch "updated" or "renamed" item is doomed to fail since item is already in other root (/archive). Handling this situation with smarter event handling - grouping them by type and filtering moved to archive items from update and rename events


alansemenov (changes requested):
I still get the same error when archiving the entire Large Tree folder

ashklianko:
Working on it, huge number of archived items leads to a situation when delay between update and move is bigger then debounce time, thus unwanted update event is not filtered out and update is processed for already moved item

alansemenov (approved):

Issue #3634, opened 2021-10-07
Show the exact number of archived/restored items in the confirmation message


alansemenov:
When archiving a site or a folder with multiple items, user gets the confirmation dialog where he sees the actual number of all items that will be archived. However, after the items are archived, we show the "success" message which shows the number of originally _selected_ items (in this case - 1) and not the number of items that were _actually archived_. 

The message should show exactly the same number as the confirmation dialog. Or actually the number of processed items should be returned by the API.

So, in this specific example the message should've said "_203 items are archived_".

The same goes for restore.

![image](https://user-images.githubusercontent.com/11612490/136400317-116ef71c-d097-4754-ad3d-beb1e3e1cb94.png)

![image](https://user-images.githubusercontent.com/11612490/136401921-50839429-bf0d-43bc-bb3f-904b27bd90f6.png)

alansemenov:
Also note that the number displayed on the "Archive item(s)" and "Restore item(s)" is also incorrect now. If you select a folder with 200 child items it should show "201" on the button, not 1.

PR #3637, opened 2021-10-08
Show the exact number of archived/restored items in the confirmation …

ashklianko:
…message #3634

-fixed issue with archive action button label: have to set label via button action
-hiding progressbar manager notifications for archive action since it doesn't contain full number of processed items and showing correct notification from dialog


alansemenov (changes requested):
A few small things to make it more consistent:
1. Change the button in the Archive dialog to say "Archive" instead of "Archive item(s)". It should say simply "Archive" for 1 item and "Archive (XX)" when there's more than 1.
2. When archiving 1 item, the message should be "`Item "<itemName>" is archived`", similar to how we have for delete, move etc. If this is for some reason not possible, it should say "`The item is archived`".
3. For several items the message should be "`XX items are archived`" (not `item(s)` since we have a separate message for 1 item)

ashklianko:
Updated

alansemenov (approved):

Issue #3636, opened 2021-10-08
Fix duplicate icons


alansemenov:
https://github.com/enonic/app-contentstudio/blob/epic-archive-new/modules/app/src/main/resources/assets/icons/icons.less
https://github.com/enonic/app-contentstudio/blob/epic-archive-new/modules/lib/src/main/resources/assets/icons/icons.less

ashklianko:
I've also noticed that lots of icons in contentstudio lib also overlapping with icons in lib-admin-ui, so will try to remove duplicated also from contentstudio lib

PR #3638, opened 2021-10-09
Fix duplicate icons #3636

ashklianko:
-Using common icons from lib-admin-ui in the first place
-Removed all duplicated icons from icomoon-studio-lib (those overlapping with lib-admin-ui) and updated .ts classes to use common ones
-Left empty icomoon-studio-app fonts (since all required icons exist in icomoon-studio-lib) for future use
-Moved archive icon into icomoon-studio-lib and added box-remove icon as icon-box-remove for using in versions history


modules/app/src/main/resources/assets/icons/icons.less
```diff
@@ -1,170 +1,9 @@
 @font-face {
   font-family: 'icomoon-studio-app';
   src:
-    url('fonts/icomoon-studio-app.woff2') format('woff2'),
```
alansemenov: This is wrong. WOFF2 should be the primary format used, WOFF should be fallback used for compatibility with older browsers. Restore `icomoon-studio-app.woff2`, make sure it has the the same contents as `icomoon-studio-app.woff`. Also add `icomoon-studio-lib.woff2` and make sure it's used as primary format in `@font-face`

alansemenov (changes requested):

alansemenov (approved):

PR #3662, opened 2021-10-13
ContentSummaryJson: add 'originalParentPath' and 'originalName' field…

ashklianko:
…s #3661

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/content/ContentSummary.ts
+++ b/modules/lib/src/main/resources/assets/js/app/content/ContentSummary.ts
@@ -478,7 +478,7 @@ export class ContentSummaryBuilder {
             this.originProject = source.getOriginProject();
             this.listTitle = source.getListTitle();
             this.originalParentPath = source.getOriginalParentPath();
-            this.originalName = source.getOriginalParentPath();
+            this.originalName = source.getOriginalName();
         }
     }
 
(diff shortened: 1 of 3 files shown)


alansemenov (approved):

modules/lib/src/main/resources/assets/js/app/content/ContentSummary.ts
```diff
@@ -452,6 +477,8 @@ export class ContentSummaryBuilder {
             this.inherit = source.getInherit();
             this.originProject = source.getOriginProject();
             this.listTitle = source.getListTitle();
+            this.originalParentPath = source.getOriginalParentPath();
+            this.originalName = source.getOriginalParentPath();
```
alansemenov: Copy/paste

alansemenov (changes requested):

alansemenov:
Also, implement a new method `ContentSummary.getFullOriginalPath` where you will return
```
`${this.getOriginalParentPath()}/${this.getOriginalName()}`.replace('//','/')
```
in case we need to show this path in several places. Then update https://github.com/enonic/app-contentstudio-plus/pull/239/commits/2b9ce873cc1697173d8173be8f3ad096c690c9ff

alansemenov (approved):

Issue #3670, opened 2021-10-15
Refactor ContentItemPreviewPanel and ArchiveItemPreviewPanel


alansemenov:
`ContentItemPreviewPanel` in CS and `ArchiveItemPreviewPanel` in CS+ share too much identical code. Extract common logic into a new class in CS, and change both `ContentItemPreviewPanel` and `ArchiveItemPreviewPanel` to derive from it.

PR #3705, opened 2021-10-20
Refactor ContentItemPreviewPanel and ArchiveItemPreviewPanel #3670

ashklianko:
-using ViewItem instead of ContentSummaryAndCompareStatus
-Making this class easier to extend


alansemenov (approved):
Good stuff!

Issue #3655, opened 2021-10-12
Enable hiding the Default project


alansemenov:
In XP8 the default repository/project will not be a part of a new XP instance, but until then we will support a config to enable hiding the Default project in the UI.

Call the new config "`settings.hideDefaultProject`".

Currently, the Default project is visible/enabled for a user who has `cms.cm.app` or admin role. The new config will override this behaviour, so even if user permissions allow accessing the Default project, the config value set to `true` will restrict the access.

Content Studio already has logic for handling missing permissions for a project:
- it should be hidden from the project selection modal dialog and the Settings grid
- CS should handle the case when there's only Default project in the system and the config is set to "true"
- CS should handle the case when last selected project was the Default one but the config is set to "true", so it should pick the next available project
etc.

![image](https://user-images.githubusercontent.com/11612490/136947491-02e1b1c5-f41d-4825-9dbe-4efb923c9607.png)

![image](https://user-images.githubusercontent.com/11612490/136947535-d002cbf2-5e9e-4f4d-8709-4320c303d062.png)

PR #3749, opened 2021-10-27
Implement config for hiding the Default project #3655

ashklianko:


alansemenov (approved):
Great job, @ashklianko

Issue #3747, opened 2021-10-27
Less codacy changes second batch


poi33:
We still have a lot of styling/linting issues on our less files.

Fix more of the codacy issues related to less files.

PR #3750, opened 2021-10-27
Codacy less changes #3747

poi33:
More less issues fixed so codacy does not complain and gives us a bad grade


modules/app/src/main/resources/assets/styles/main-app-wrapper.less
```diff
@@ -11,7 +10,6 @@
     width: @app-header-height;
     position: absolute;
     top: 0;
-    left: 0;
```
alansemenov, line 14: Any special reason to remove this? `left: 0` is not always implied by default.

alansemenov (approved):

modules/app/src/main/resources/assets/styles/main-app-wrapper.less
poi33: It is defined again futher down. Duplicate rule "left: -@app-header-height"

Issue #3823, opened 2021-11-10
Incorrect validation message for an invalid required occurrence


alansemenov:
Incorrect validation message is given when an invalid attachment is uploaded into an AttachmentUploader with min 1 required occurrence. The message shown is "_This field is required_" but the attachment is already uploaded, so the correct message to show in this case is the same as for occurrences.minimum=0, which is "_One or more attached items are invalid_".

![image](https://user-images.githubusercontent.com/11612490/141146289-664f62f3-75b0-43f7-9284-6e45c7cc884e.png)

Input definition:
```xml
    <input name="file" type="AttachmentUploader">
      <label>File</label>
      <custom-text>file</custom-text>
      <occurrences minimum="1" maximum="1"/>
      <config/>
    </input>
```

PR #3855, opened 2021-11-19
Incorrect validation message for an invalid required occurrence #3823

ashklianko:
-For case with invalid occurrence for min=1 max=1 replacing InputValidationRecording with one that will trigger custom error

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/upload/AttachmentUploader.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/upload/AttachmentUploader.ts
@@ -307,11 +307,26 @@ export class AttachmentUploader
             });
 
             this.toggleClass('invalid', hasError);
+
             if (hasError) {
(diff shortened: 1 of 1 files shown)


alansemenov (changes requested):
Not completely there yet...
1. Error messages for occurrences should be hidden by default, user should click "Show details" to see them. Check how it's done for attachments.

2. Even though an occurrence has a custom validation error, the input itself still has "`valid`" class on it. This results in green border around invalid input
image

3. A bad one. After adding a **valid** occurrence to an input that already has an invalid one, the entire input (and the content itself) is marked as valid:
image

alansemenov:
Correction: item 3 should be fixed in lib-admin-ui as described [here](https://github.com/enonic/lib-admin-ui/pull/2276#pullrequestreview-815081352). Maybe some of the first two should also be fixed in lib-admin-ui.

ashklianko:
Updated in lib-admin-ui

alansemenov (approved):

Issue #3873, opened 2021-11-23
Incorrect behaviour of preview panel in mobile mode


alansemenov:
This was changed some time during CS 4.0 development.

Was (and how it should be):
![CS_33](https://user-images.githubusercontent.com/11612490/143004446-73208b52-f039-4d52-90a4-810360fd0582.gif)

Now:
![CS_40](https://user-images.githubusercontent.com/11612490/143003776-08e2b6f6-545f-48a8-80ea-c5000c5863b4.gif)

alansemenov:
Make sure https://github.com/enonic/app-contentstudio/issues/3347 isn't reproducing after the fix

PR #3883, opened 2021-11-25
Incorrect behaviour of preview panel in mobile mode #3873

ashklianko:
-adjusted styling of a preview panel and actions toolbar on small resolutions to correspond with it's look in previous versions

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseToolbar.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseToolbar.ts
@@ -28,20 +28,16 @@ export class ContentBrowseToolbar
         this.hideMobilePreviewButton.onClicked(action);
     }
 
-    setFoldButtonLabel(value: string) {
-        this.fold.setLabel(value);
-    }
-
     enableMobileMode(): void {
         this.setLocked(true);
         this.addAction(this.allActions.getPublishActions()[0]).addClass('publish-action');
-        this.doFold(true);
+        this.fold(true);
     }
 
     disableMobileMode(): void {
         this.setLocked(false);
         this.removeAction(this.allActions.getPublishActions()[0]);
-        this.doExpand();
+        this.expand();
     }
 
     protected isItemAllowedToFold(elem: Element): boolean {
--- a/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseToolbar.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseToolbar.ts
@@ -1,13 +1,50 @@
 import {ActionName, ContentTreeGridActions} from './action/ContentTreeGridActions';
 import {Toolbar} from 'lib-admin-ui/ui/toolbar/Toolbar';
+import {Button} from 'lib-admin-ui/ui/button/Button';
+import {Element} from 'lib-admin-ui/dom/Element';
 
 export class ContentBrowseToolbar
     extends Toolbar {
 
+    private readonly hideMobilePreviewButton: Button;
+
+    private readonly allActions: ContentTreeGridActions;
+
     constructor(actions: ContentTreeGridActions) {
         super();
+
+        this.allActions = actions;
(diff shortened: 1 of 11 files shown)


modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseToolbar.ts
```diff
@@ -1,13 +1,50 @@
...
+        this.prependChild(this.hideMobilePreviewButton);
+    }
+
+    onFoldClicked(action: () => void) {
+        this.hideMobilePreviewButton.onClicked(action);
+    }
+
+    setFoldButtonLabel(value: string) {
+        this.fold.setLabel(value);
+    }
```
alansemenov: This method should be in the base class.

alansemenov (changes requested):

alansemenov:
A few more things after testing the UI:
1. We need to fix checkboxes in the grid so that they are right-aligned (how it used to be in the mobile resolution before the changes). @edloidas has [fixed](https://github.com/enonic/app-applications/pull/571/files#diff-b1f48b4550018a304fb8684792f88aea89d699cc4cf98404dd4dbfe8e31be898R29) it for app-applications, but it has to be consistent everywhere. Here's how it should look:
image

2. When an item is tapped and preview panel slides in, we need to unhighlight the row. Otherwise you have to click the item twice to open the preview again.

3. The dropdown menu in the preview panel should have a fixed width of 200px. Now it's too narrow.

image

ashklianko:
All updated

edloidas:
A few suggestions here:
1. In mobile mode, align the grid checkboxes with the toolbar "Check All" checkbox. Will be enough to add some [grid styles](https://github.com/enonic/app-applications/blob/master/src/main/resources/assets/styles/browse/application-tree-grid.less#L47-L49).
2. Set mobile statistics header (ContentBrowseToolbar) height to 44px (now it's 40px), just like in grid view. This will make the transition from the grid to statistics more organic.
3. I know that we have a different DOM structure right now, but I guess we should avoid the statistics transition from bottom to top because it looks kinda annoying. Besides, animating it from right to left is pretty much enough.

ashklianko:
> A few suggestions here:
> 
> 1. In mobile mode, align the grid checkboxes with the toolbar "Check All" checkbox. Will be enough to add some [grid styles](https://github.com/enonic/app-applications/blob/master/src/main/resources/assets/styles/browse/application-tree-grid.less#L47-L49).
> 2. Set mobile statistics header (ContentBrowseToolbar) height to 44px (now it's 40px), just like in grid view. This will make the transition from the grid to statistics more organic.
> 3. I know that we have a different DOM structure right now, but I guess we should avoid the statistics transition from bottom to top because it looks kinda annoying. Besides, animating it from right to left is pretty much enough.

The idea was to make it look like in previous versions (it started looking different after my refactoring), so 3. Is just to make it correspond to previous version. 1. Isn't it aligned? And it also won't be aligned when grid has a scrollbar 2. Can do in other task

edloidas:
1. Sorry, my bad. They were displaced due to page scaling. As for the scrollbars, we don't have them on mobile.
2. & 3 Yes, that's why I'm suggesting this. Previously, the statistics header was exactly the same size, as in grid view. And there was only one animation/transition: from right to left, and not from bottom to top. Here is the comparison between the current and the old statistics panels.

| Current | Old |
| --- | --- |
| <br/><br/> localhost_8080_admin_tool_com enonic app contentstudio_main(iPhone X) (2) | localhost_8080_admin_tool_com enonic app contentstudio_main(iPhone X) (3) |

alansemenov:
+1 for fixing the toolbar height. Dropdown icon we can keep as it is now, it better fits the purpose.

ashklianko:
Our regular toolbar height  for desktop resolutions is 40px, it is irrational to make it bigger for a mobile mode. ContentAppBar on the top is 44px, but with current animation we accurately covering it with toolbar.

alansemenov (approved):

Issue #3876, opened 2021-11-23
Incorrect panel selected when a Text component is in edit mode


alansemenov:
When a text component is being edited, the inspect panel is shown as if no component is selected (Insert tab is active and Page is selected instead of Text):

![image](https://user-images.githubusercontent.com/11612490/143028257-81fccfde-129e-46bc-9362-e5d1891f6f84.png)

After closing the text mode everything is correct again:
![image](https://user-images.githubusercontent.com/11612490/143028408-2762aa38-0896-45d0-b956-bd1883cd5e93.png)

PR #3888, opened 2021-11-26
Incorrect panel selected when a Text component is in edit mode #3876

edloidas:
Prevented deselection when text component enters edit mode.
Added selection switch, when changing focus between multiple text components.


alansemenov (approved):
Nice

Issue #3890, opened 2021-11-29
Mobile mode: adjust classes and styles with recent app-cs changes


ashklianko:
@ashklianko commented on [Mon Nov 29 2021](https://github.com/enonic/app-contentstudio-plus/issues/314)

ashklianko:
Extracting common mobile mode behaviour to a new parent class so it can be overridern both in studio and archive

PR #3891, opened 2021-11-29
Mobile mode: adjust classes and styles with recent app-cs changes #3890

ashklianko:
-Extracting mobile mode functionality to a separate classes so archive (and any other class) could reuse it
-Reusing parents handling of selection and highlighting change


modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
```diff
@@ -68,6 +60,9 @@ export class ContentBrowsePanel
     protected initElements() {
         super.initElements();
 
+        this.getBrowseActions().getAction(ActionName.UNDO_PENDING_DELETE).setVisible(false);
```
alansemenov: We're no longer using Undo Pending Delete. Shouldn't we get rid of this action altogether?

alansemenov (commented):
Nice refactoring

modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
ashklianko: Will remove

ashklianko: Done

alansemenov (approved):

PR #3922, opened 2021-12-03
AppContainer: let inheritors override most of the methods #3921

ashklianko:


modules/lib/src/main/resources/assets/js/app/AppContainer.ts
```diff
@@ -12,13 +12,13 @@ export abstract class AppContainer
 
     protected appPanel: AppPanel;
 
-    constructor() {
-        super('app-container');
+    protected constructor(className?: string) {
+        super('app-container ' + (className ? className : ''));
```
alansemenov: `app-container ${className || ''}`

modules/lib/src/main/resources/assets/js/app/AppContainer.ts
```diff
@@ -29,27 +29,30 @@ export abstract class AppContainer
...
-            this.appendChild(this.appPanel);
+            this.appendElements();
 
             return rendered;
         });
     }
 
+    protected appendElements(): void {
+        this.appendChild(this.appBar);
+        this.appendChild(this.appPanel);
```
alansemenov: this.appendChildren(...[this.appBar, this.appPanel]);

alansemenov (changes requested):

Issue #3852, opened 2021-11-18
Inconsistent behaviour of the Page Editor & Context panel layout


sgauruseu:
1. Set a browser in full screen mode, Open new `Article` or double wizard
2. Click `Show Page Editor` 
3. Click on `Hide Page Editor` 

BUG - Splitter and scrollbar gets not visible after click on  `Hide Page Editor` 

![image](https://user-images.githubusercontent.com/3728712/142435485-f63c9cf4-1296-4c15-adea-0d8f1bf39c08.png)

alansemenov:
Current issues with Context panel:
* No splitter/scrollbar between the Content form and the Context panel, but only when there's no Page Editor panel. When Page Editor is present, the splitter is there.
* Context panel stopped being responsive, but only when Page Editor IS present. When there's no Editor, the Context panel is responsive.

PR #3937, opened 2021-12-10
Inconsistent behaviour of the Page Editor & Context panel layout #3852

ashklianko:
-Refactored massively ContextSplitPanel and related classes: now only 1 DockedContextPanel used instead of 3 (-FloatingContextPanel, -MobileContextPanel previously)
-Removed redundant context manager classes: ActiveContextPanelManager, NonMobileContextPanelsManager
-Enriched ContextSplitPanel with required API to handle context panel behaviour and state
-Introduced ContextSplitPanel inheritor for ContentWizardPanel to handle live panel correctly and remove non related code from ContextSplitPanel to it
-Updated size balancing between context panel and rest panels to correctly spread available size according to https://github.com/enonic/app-contentstudio/issues/3226
-fixed non related issue with ShowLiveEditAction to make live panel toggler change color when showing/hiding live panel


modules/lib/src/main/resources/assets/js/app/view/context/ContextSplitPanel.ts
```diff
@@ -2,134 +2,262 @@ import {ResponsiveManager} from 'lib-admin-ui/ui/responsive/ResponsiveManager';
...
         }
-        nonMobileContextPanelsManagerBuilder.setSplitPanelWithContext(this);
-        nonMobileContextPanelsManagerBuilder.setDefaultContextPanel(this.dockedContextPanel);
-        this.floatingContextPanel = new FloatingContextPanel(this.contextView);
-        nonMobileContextPanelsManagerBuilder.setFloatingContextPanel(this.floatingContextPanel);
+    }
+
+    protected getExpectedContextPanelMode(): ContextPanelMode {
+        const parentWidth: number = this.getParentElement().getEl().getWidthWithBorder();
+        const letPanelFloatingModeResponsiveRange: ResponsiveRange = this.getLeftPanelResponsiveRangeToSwitchToFloatingMode();
```
alansemenov: Typo in `le_f_tPanelFloatingModeResponsiveRange`?

modules/lib/src/main/resources/assets/js/app/view/context/ContextSplitPanel.ts
```diff
@@ -2,134 +2,262 @@ import {ResponsiveManager} from 'lib-admin-ui/ui/responsive/ResponsiveManager';
...
+    private renderAfterDockedPanelReady(): void {
+        this.hideContextPanel();
+        this.subscribeContextPanelsOnEvents();
+    }
 
-        this.subscribeContextPanelsOnEvents(this.contextPanelsManager);
+    private requiresCollapsedContextPanel(): boolean {
+        const totalWidth: number = Body.get().getEl().getWidthWithBorder();
+        return this.getExpectedContextPanelMode() === ContextPanelMode.FLOATING ||
+               ResponsiveRanges._1620_1920.isFitOrSmaller(totalWidth);
```
alansemenov: Swap the two conditions so that a simpler condition is called before the one with all the calculations.

alansemenov (changes requested):
Check why tests are failing

alansemenov (approved):

Issue #3086, opened 2021-04-28
Publishing wizard: Can't reach "mark as ready" with keyboard alone


thomasheartman:
In the publishing dialog, the "mark as ready" option is unreachable via keyboard. I can navigate to and click the arrow that opens the dropbown options (next to publish now), but I can't navigate to ittems within it. In the below image, I can't reach the "Mark as ready (1)" option without a mouse.

![image](https://user-images.githubusercontent.com/17786332/116359585-41f4e680-a7ff-11eb-98e5-ca4d6d8255e5.png)

PR #3939, opened 2021-12-10
Added keyboard accessibility to 'mark as ready' (#3086)

reisfmb:
Now the user can use the keyboard (tab + enter) to navigate and activate the "Mark as ready" option:

https://user-images.githubusercontent.com/67838246/145598830-d2c38234-81c6-4596-8a29-cfa2eb6b3e04.mp4

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/dialog/BasePublishDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/BasePublishDialog.ts
@@ -23,7 +23,8 @@ import {DefaultErrorHandler} from 'lib-admin-ui/DefaultErrorHandler';
 import {Action} from 'lib-admin-ui/ui/Action';
 import {showFeedback} from 'lib-admin-ui/notify/MessageBus';
 import {ContentId} from '../content/ContentId';
-import {Menu} from 'lib-admin-ui/ui/menu/Menu';
+import {MenuItem} from 'lib-admin-ui/ui/menu/MenuItem';
+import {MenuButton} from 'lib-admin-ui/ui/button/MenuButton';
 import {AccessibilityHelper} from '../util/AccessibilityHelper';
 
 export abstract class BasePublishDialog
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/app/dialog/BasePublishDialog.ts
```diff
@@ -331,5 +333,13 @@ export class PublishDialogButtonRow
     setTotalInProgress(totalInProgress: number) {
         this.toggleClass('has-items-in-progress', totalInProgress > 0);
         this.getMenuActions()[0].setLabel(i18n('action.markAsReadyTotal', totalInProgress));
+        this.addAccessibilityToMarkAsReadyTotalElement();
```
ashklianko: `setTotalInProgress()` is executed each time you open `ContentPublishDialog `or `RequestContentPublishDialog` since they are singletons, this method is required for updating button labels when items selected change thus it is not right to execute same operation each time dialog is open
Better option is to override `makeActionMenu` method that creates MenuButton and trigger assigning of a tab index there:

```
makeActionMenu(mainAction: Action, menuActions: Action[], useDefault?: boolean): MenuButton {
        const menuButton: MenuButton = super.makeActionMenu(mainAction, menuActions, useDefault);

        this.addAccessibilityToMarkAsReadyTotalElement();

        return menuButton;
    }
```

modules/lib/src/main/resources/assets/js/app/util/AccessibilityHelper.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export class AccessibilityHelper {
+
+    static tabIndex(element: Element, index: number = 0) {
+       // Tab navigation
+       element.getEl().setTabIndex(index);
+
+       // Enter event will execute a click
+       element.onKeyPressed((event: KeyboardEvent) => {
+           event.keyCode === 13 && element.getHTMLElement().click();
```
ashklianko: keyCode is already deprecated, have to use  `event.key === 'Enter'` instead.
I have to admit that we still have a lot of places in our studio code where it is still used, however common principle is to update and remove such a code whenever you current task leads you to the affected code. And ofcourse not to produce new places :)
If we go further, we might need a common class like KeyValues to hold those key values created in lib-admin-ui which is our common lib holding basic elements and classes.

modules/lib/src/main/resources/assets/js/app/dialog/BasePublishDialog.ts
```diff
@@ -331,5 +333,13 @@ export class PublishDialogButtonRow
...
+    }
+
+    private addAccessibilityToMarkAsReadyTotalElement(): void{
+        const menu = this.actionMenu.getChildren()
```
ashklianko: 1. One of the main reasons to use a typescript is to have static typing so you could always know what's behind every variable and what you're expecting to see there, that prevents from some runtime issues. So need to set types in here for all constants, inclduing those used in the callback functions.
2. However on this particular case you can access desired MenuItem by simpler code, without any type conversions:
```
private addAccessibilityToMarkAsReadyTotalElement(): void {
        const markAsReadyMenuItem: MenuItem = this.actionMenu.getMenuItem(this.getMenuActions()[0]);
        AccessibilityHelper.tabIndex(markAsReadyMenuItem);
    }
```

ashklianko (changes requested):

ashklianko (approved):

Issue #2986, opened 2021-03-17
Unable to open 'select project' dialog with keyboard


thomasheartman:
# Issue

I'm unable to open to the 'select project' dialog with the keyboard. In fact, I can't even seem to navigate to the button with the keyboard at all.

# Expected behavior

When tabbing through the elements of the page, I'd expect to reach the 'select project' button at some point. This should also be visibly indicated.

PR #3941, opened 2021-12-10
Added keyboard accessibility to open 'select project' dialog (#2986)

reisfmb:
Now the user can use the keyboard (tab + enter) to open the 'select project' dialog:

https://user-images.githubusercontent.com/67838246/145607342-ebaf911b-dae4-4d7d-b631-73cc0cda0b1b.mp4


ashklianko:
One more thing to fix in this issue for a ProjectSelectionDialog:

image

Focus remains on the project selection icon, pressing TAB will move focus to the next tabbable element outside of the dialog.
Need to set focus on the first element in the dialog. 
Origin of the issue is that when this dialog is being opened we are supposed to set focus on the action button on the bottom which is missing for this dialog.

ashklianko (approved):

Issue #3943, opened 2021-12-13
ProjectSelectionDialog doesn't have focus set inside of the dialog


ashklianko:
ProjectSelectionDialog after opening it  :

![image](https://user-images.githubusercontent.com/4255706/145827466-b07ec877-218d-4074-920a-63c49b4de312.png)

Focus remains on the project selection icon, pressing TAB will move focus to the next tabbable element outside of the dialog.
Need to set focus on the first element in the dialog. 
Origin of the issue is that when this dialog is being opened we are supposed to set focus on the action button on the bottom which is missing for this dialog.

ashklianko:
Done in https://github.com/enonic/lib-admin-ui/issues/2321

PR #3946, opened 2021-12-13
Added focus to first project in 'select project' dialog (#3943)

reisfmb:
Now the first element in the dialog will be focused when the dialog is opened:

https://user-images.githubusercontent.com/67838246/145865738-58a19365-f95a-4b57-a7f9-87975f37536e.mp4


modules/lib/src/main/resources/assets/js/app/settings/dialog/ProjectSelectionDialog.ts
```diff
@@ -92,6 +92,7 @@ export class ProjectSelectionDialog
     open() {
         if (!this.updateOnOpen) {
             this.showItems();
+            this.addFocusToFirstProjectListElement();
```
ashklianko: Will work only if no update needed. Try to create some layer or project in Settings app, and then open dialog again - focus will not be set

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/app/settings/dialog/ProjectSelectionDialog.ts
ashklianko: Dialog has a lazy loading for a projects, so if no updates on existing projects/layers were made then it won't load entire list of projects, however if create/delete event happened then dialog will refetch it's items and focus will have to be set after that

Issue #2989, opened 2021-03-17
Select project: tab selection only cycles backward


thomasheartman:
# Issue

In the select project dialog, once focus is trapped within the modal, it allows you to use `tab` to select projects. However, it only allows cycling one way (backwards).

# Expected behavior

The modal allows you to cycle through choices in both directions.

# Video

Both videos below only use the tab key (shifted or unshifted) for navigation (after trapping focus). 

This video shows the selection tabbing backwards. When it reaches the topmost element and the user tabs again, you wrap around to the bottom.

https://user-images.githubusercontent.com/17786332/111479249-13c8b500-8731-11eb-8b79-ec422db96434.mp4

This video shows what happens when you tab forwards (without shift). Once you reach the bottom elements any subsequent tab presses don't do anything.

https://user-images.githubusercontent.com/17786332/111479270-1925ff80-8731-11eb-918f-44a22e578f92.mp4

PR #3947, opened 2021-12-13
Added projects tab cycle in the 'select project' dialog (#2989)

reisfmb:
If the last project is focused and the tab is pressed, the focus will change to the first element:

https://user-images.githubusercontent.com/67838246/145883870-b9a6b0ce-219f-422b-96fb-321625412720.mp4


modules/lib/src/main/resources/assets/js/app/settings/dialog/ProjectSelectionDialog.ts
```diff
@@ -137,4 +139,19 @@ export class ProjectSelectionDialog
         });
     }
 
+    // Cycle to the first project element when tabbing in the last element
+    private setLastProjectCycleOnTabEvent(): void {
+        if(this.projectsList.getItemViews().length === 0) return;
```
ashklianko: According to our code conventions we always use brackets after `if` statement, no matter if it has only return after, so should be: 
```
if (this.projectsList.getItemViews().length === 0) {
            return;
}
```

modules/lib/src/main/resources/assets/js/app/settings/dialog/ProjectSelectionDialog.ts
```diff
@@ -137,4 +139,19 @@ export class ProjectSelectionDialog
...
+
+        const firstProject: ProjectListItem =
+            this.projectsList.getItemViews()[0] as ProjectListItem;
```
ashklianko: Everything is fine in here, just let's use our commonly used angle-bracket syntax:

`const firstProject: ProjectListItem = <ProjectListItem>this.projectsList.getItemViews()[0];`

ashklianko (approved):

ashklianko (approved):

PR #3961, opened 2021-12-14
ui-tests- issue#3942

sgauruseu:
1. Migrated testes for "Disable macros"  option
2. Migrated tests for reverting versions in content with Image Selector


jsi (approved):
Nice!

PR #3962, opened 2021-12-14
Content Wizard - error in console after opening a content with Preview Panel

alansemenov:


ashklianko (changes requested):
I'd prefer to modify issue origin in `PageEditorWidgetItemView:`
```
private static isPreviewAvailable(item: ContentSummaryAndCompareStatus): 
Q.Promise<boolean> {
        return item.hasContentSummary() ? new IsRenderableRequest(item.getContentId()).sendAndParse() : Q(false);
    }

```

Otherwise we might get some strange issues when code that is supposed to fail because of missing id will execute with wrong renderable

alansemenov:
Valid point, updated

ashklianko (approved):

Issue #3087, opened 2021-04-28
Publishing wizard: can't remove items or select child items of an item to be published


thomasheartman:
In the content publishing wizard, I'm unable to interact with the list of items to be published via the keyboard. In particular, I'm unable to reach the "X" or the "select child items" icons without using the mouse.

![image](https://user-images.githubusercontent.com/17786332/116359841-91d3ad80-a7ff-11eb-84e2-cd7326e846ca.png)

PR #3969, opened 2021-12-14
Added tab navigation to items to be published (#3087)

reisfmb:
I wasn't able to find the correct place where the child items (such as 'include child items' and 'remove item' for example) of an item to be publish are defined. 

Previously, the 'remove item' was the first child in DOM, even though being shown as the last element, and that was causing an unexpected behavior when adding tab index to it. Therefore my workaround was to replace the 'remove item' as the last child and then add tab index to all the item to be published child.

It's working, but there's probably a better place to define the order of the child elements of an item to be published. If so, where is it @ashklianko ?


ashklianko:
Hi Bruno! You're right: there's no proper way to resolve it without correctly positioning elements in parent classes, so I've created PRs for master that will help you:
https://github.com/enonic/lib-admin-ui/pull/2315
https://github.com/enonic/app-contentstudio/pull/3971

After they are merged you'll be able to update this PR accordingly

reisfmb:
> Hi Bruno! You're right: there's no proper way to resolve it without correctly positioning elements in parent classes, so I've created PRs for master that will help you: [enonic/lib-admin-ui#2315](https://github.com/enonic/lib-admin-ui/pull/2315) #3971
> 
> After they are merged you'll be able to update this PR accordingly

Hi Sasha. After the update in the lib-admin-ui, now the remove element is positioned as the second-last element as you can see in the following screenshot:

image

This makes the tab navigation not work as expected, i.e, there's an extra tab after you reach the remove element that goes backwards to status (published / unpublished) element.

Anyway, I'll commit the code removing the swap of elements that I did initially... with that I believe that after the fix of the positioning  I've mentioned in the lib-admin-ui, we'll be safe to close this PR.

PR #3983, opened 2021-12-16
Improved focus in the first project of the 'select project' dialog (#3943)

reisfmb:


ashklianko (changes requested):
TAB is working OK, but Shift+TAB on the last element will still move focus to the first element, not the one before last

ashklianko:
One more question: do we need those `setTimeout `functions wrappers around keydown handlers?

reisfmb:
> One more question: do we need those `setTimeout `functions wrappers around keydown handlers?

Yes, tried without them before and it wasn't working properly.

ashklianko:
Was fixed in https://github.com/enonic/lib-admin-ui/issues/2321

PR #3984, opened 2021-12-16
Added tab navigation to elements of publish dialog (#3087)

reisfmb:


ashklianko (approved):
Hey, Bruno! You may check my commit that I've added to the PR, pay attention to certain moments:
1) Always pay attention to the code around your own changes: if you see something to be improved then don't hesitate to improve it
2) If constructor getting to big it is always good idea to split construction process to some parts, me personally usually add protected initElements to init class vars and initListeners to assign required listeners to them methods
3) I prefer adding css classes in doRender method or passing it in a super('css-class') call (if applicable) to avoid cluttering up constructor

Issue #2945, opened 2021-03-05
Feature request: shortcut for searching


thomasheartman:
The search functionality in the content tree is so useful that (I think) it deserves to be easily accessible without having to take your hands off the keyboard and without having to tab around the interface. The documentation doesn't list a shortcut for it. I wish there was a shortcut to search (and maybe to clear a search?). The `/` key is often used for this and it doesn't seem to serve any other purpose in Content Studio. Thanks! :pray:

PR #3985, opened 2021-12-16
Added shortcut for toggling browse panel (#2945)

reisfmb:
The author suggested the use of `/` to open the browse panel. From [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#numeric_keypad_keys), it seems that the suggested key ("Divide") is only available in numerical keypads. For example, in my notebook, to get `/` I need to type `AltGr + q`, and that would be handled.

@alansemenov suggested to find a shortcut combination that is not reserved in browsers (list of shortcuts for [Chrome](https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Ctab-and-window-shortcuts) and [Firefox](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly)).

Therefore I set the shortcut to be `Shift + F`. I don't know if this is a good combination, but from what I've seen it's not used in the browsers above and the `F` stands for find, which makes sense.

If accepted, might be necessary to update [this list of shortcuts for the app-contentstudio](https://developer.enonic.com/docs/content-studio/stable/keyboard-shortcuts).

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/ContentAppPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/ContentAppPanel.ts
@@ -20,7 +20,7 @@ export class ContentAppPanel
 
     protected resolveActions(panel: Panel): Action[] {
         const actions = super.resolveActions(panel);
-        return [...actions, ...this.getBrowsePanel().getNonToolbarActions()];
+        return [...actions, ...this.getBrowsePanel().getNonToolbarActions(), this.getBrowsePanel().getToggleSearchAction()];
     }
 
     calculateOffset() {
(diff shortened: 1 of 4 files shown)


ashklianko:
`Shift+F` is fine! However this solution has a potential problem: when set this shortcut will also work when you have Settings and Archive panel open. I've found other option: to assign it via ToggleSearchPanelAction that is used in ContentBrowsePanel. Important thing here is that when switching between Studio, Settings and Archive we stash hidden panel shortcuts and enable shortcuts of a panel to be opened thus shortcuts only work when their panel is open. That is done in `AppPanel.handlePanelShown()`

ashklianko:
NB: We also have a `KeyHelper.ts` class to detect which button is pressed, to avoid duplicating checks like `event.key === TAB` , we should use it and when doing next epic task you may update your previous tasks to use this class!

reisfmb:
@ashklianko Shouldn't we open an issue (like #3992) to add an entry for this shortcut `Shift + F` in the keyboard-shortcuts.adoc file?

Issue #3944, opened 2021-12-13
Details panel loses "responsiveness" after switching from docked to floating mode and back


alansemenov:
See video. Originally everything is correct, the details panel in the docked mode is responsive and its width decreases along with width of the browser window. Problems begin when the window width is decreased to mobile resolution and then starts increasing.

Bug 1: When switching from mobile to "normal" mode, the details panel reappears with strange animation of flying from the bottom of the screen.

Bug 2: When the panel switches from floating to docked mode, it's no longer responsive (as it used to be in the very beginning) and will keep fixed width no matter how wide the browser window becomes.

https://user-images.githubusercontent.com/11612490/145842819-cf3493fe-85de-42f2-a65a-6cc43e6995b4.mov

PR #3990, opened 2021-12-17
Details panel loses "responsiveness" after switching from docked to f…

ashklianko:
…loating mode and back #3944


ashklianko:
Updated

alansemenov (approved):

PR #4013, opened 2021-12-21
Archive item(s) wizard includes outdated text

alansemenov:


ashklianko (approved):
There's also entry in studio.plus in ArchiveDeleteDialog using deleteNow phrase

PR #4050, opened 2021-12-28
Implement LazyListBox for a DialogDependantList #4049

ashklianko:


modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsDialog.ts
```diff
@@ -71,15 +68,26 @@ export abstract class DependantItemsDialog
...
+
+            this.loadDescendants(size, GetDescendantsOfContentsRequest.LOAD_SIZE).then((newItems: ContentSummaryAndCompareStatus[]) => {
+                if (newItems.length > 0) {
+                    this.addDependantItems(newItems);
+                }
+            }).catch(DefaultErrorHandler.handle)
+                .finally(() => {
+                    this.hideLoadMask();
+                });
+        }, 300));
```
alansemenov: Let's move this to a separate method.

modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsDialog.ts
```diff
@@ -71,15 +68,26 @@ export abstract class DependantItemsDialog
...
+            this.showLoadMask();
+
+            this.loadDescendants(size, GetDescendantsOfContentsRequest.LOAD_SIZE).then((newItems: ContentSummaryAndCompareStatus[]) => {
+                if (newItems.length > 0) {
+                    this.addDependantItems(newItems);
+                }
+            }).catch(DefaultErrorHandler.handle)
+                .finally(() => {
+                    this.hideLoadMask();
+                });
```
alansemenov: No need for curly brackets here

`.finally(() => this.hideLoadMask());`

alansemenov (changes requested):
Instead of having `scrollContainer` as a constructor argument in `LazyListBox` I suggest setting it via setter. Then we won't have to send in `this.getBody()` into every constructor of every descendant of `DialogDependantList` and it will be enough to set it once in `DependantItemsDialog.initElements()`: 

`this.dependantList = this.createDependantList().setScrollContainer(this.getBody());`

ashklianko:
> Instead of having `scrollContainer` as a constructor argument in `LazyListBox` I suggest setting it via setter. Then we won't have to send in `this.getBody()` into every constructor of every descendant of `DialogDependantList` and it will be enough to set it once in `DependantItemsDialog.initElements()`:
> 
> `this.dependantList = this.createDependantList().setScrollContainer(this.getBody());`

`ScrollContainer` is an integral part of `LazyListBox `and must be used immediately to construct `IntersectionObserver`, so it is essential to set it in a constructor, using setter will be more error-prone rather then explicitly setting it in inheritors (imaging if items are set before container is there). One thing I can propose is using Builder class for a `LazyListBox `where constainer could be set together with other observer properties, thus `DialogDependantList `inheritors will need to extend a builder, then  `createDependantList `will need to return Builder class

ashklianko:
Updated

alansemenov (approved):
Great stuff

PR #4066, opened 2022-01-04
ui-tests- issue#4052

sgauruseu:
migrated tests for  content Item Preview Panel, context menu and content with name in upper and lower cases


jsi (approved):
Nice!

Issue #4035, opened 2021-12-23
"Move" action is disabled in the search view


alansemenov:
https://discuss.enonic.com/t/move-action-during-the-search/2656

1. Open Content Studio
2. Expand filter panel on the left
3. Search for some content
4. Select some content in the grid

Bug: "Move" button remains disabled.

<img width="880" alt="Screenshot 2021-12-23 at 14 47 36" src="https://user-images.githubusercontent.com/11612490/147249924-16680842-9d69-448d-bbb9-effc0d338312.png">

PR #4073, opened 2022-01-05
"Move" action is disabled in the search view #4035

ashklianko:
-making Move button enabled only if there's at least one non-selected root item

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/action/MoveContentAction.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/action/MoveContentAction.ts
@@ -4,8 +4,10 @@ import {ContentSummaryAndCompareStatus} from '../../content/ContentSummaryAndCom
 import {i18n} from 'lib-admin-ui/util/Messages';
 import {ContentTreeGridAction} from './ContentTreeGridAction';
 import {ContentTreeGridItemsState} from './ContentTreeGridItemsState';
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/app/browse/action/MoveContentAction.ts
```diff
@@ -4,27 +4,35 @@ import {ContentSummaryAndCompareStatus} from '../../content/ContentSummaryAndCom
...
+    private isAnyRootItemNotSelected(): boolean {
+        // if there's at least one non-selected root item then there's at least one option where to move selected items
+        const selectedIds: string[] = this.getHighlightedOrSelectedIds();
+        const rootItemsIds: string[] = this.grid.getRootItemsIds();
+        const diff: string[] = ArrayHelper.difference(rootItemsIds, selectedIds, (a: string, b: string) => a === b);
+
+        return diff.length > 0;
+    }
+
+    private getHighlightedOrSelectedIds(): string[] {
```
alansemenov: Any special reason in a different order of words here than in the grid class' method? `getHighlightedOrSelectedIds` vs `getSelectedOrHighlightedItems`.

ashklianko: No, will update method's name

alansemenov (approved):

PR #4081, opened 2022-01-06
Enable importing license via POST

alansemenov:


rymsha (commented):
But why making a hack for tests only? It is possible to make a multipart form post on the client side

sigdestad:
You know that license files kan be placed on file as well, right?

Issue #4083, opened 2022-01-10
Line breaks get removed from TextArea inside a macro config


ashklianko:
https://discuss.enonic.com/t/line-break-on-textarea-field-in-macro/2654

1. Create a macro with a config form containing TextArea input fields.
2. Open macro dialog and select recently created macro
3. Fill in the Insert Macro dialog's fields with text containing line breaks and press Insert
4. Select in HtmlArea recently added macro and open macro dialog again

Now those line breaks are gone and text in the input fields is in on one line

PR #4101, opened 2022-01-11
Line breaks in HtmlArea's macro TextArea fields are being removed #4083

ashklianko:
-using cke element's innerText property to get/set macro text to preserve line breaks

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/MacroModalDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/MacroModalDialog.ts
@@ -202,8 +202,8 @@ export class MacroModalDialog
     }
 
     private insertUpdatedMacroIntoTextArea(macroString: string) {
-        this.selectedMacro.element.setText(
-            this.selectedMacro.element.getText().replace(this.selectedMacro.macroText, macroString));
+        this.selectedMacro.element.$.innerText =
+            this.selectedMacro.element.$.innerText.replace(this.selectedMacro.macroText, macroString);
         this.getEditor().fire('saveSnapshot'); // to trigger change event
     }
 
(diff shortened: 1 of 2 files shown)


sigdestad:
Has security aspect related to this been considered? @rymsha

rymsha:
Yes, I would double check... 
`innerText` seem to not only allow line breaks, but allows basically anything, including `<script>` tags.
But I can't tell right now how bad it is.

ashklianko:
CKEditor has built-in content filter (which we, ofcouse, can't rely on, and developers mention that) that strips and escapes vulnerable code (tags, attributes, etc. until they are allowed in the config). Basic XSS stuff definately won't  work because of innerText in that case. We also extract data from editor not by using such a dangerous stuff but by cke's provided api. The rest filtering is supposed to be on the backend.

rymsha:
It is not only stored XSS. Other XSS flavors exist.
Depending on where and how the value fetched by innerText is inserted it may cause DOM based XSS.

I'm not familiar with getText() method. I would check what it does inside

ashklianko:
If we talk about any types of the xss then the change in this task won't create additional risks, since cke content might be changed in multiple places, especially in Source dialog. Seems like cke vulnerability is a separate topic

sigdestad:
Would be nice to know how this fix actually works though. Are we sure the "old way of working" will not break? Must be testes carefully with both XP and Guillotine rendering

alansemenov:
@sigdestad this fix is not for HtmlArea

sigdestad:
? But macros are used in Htmlarea

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/MacroModalDialog.ts
```diff
@@ -202,8 +202,8 @@ export class MacroModalDialog
     }
 
     private insertUpdatedMacroIntoTextArea(macroString: string) {
-        this.selectedMacro.element.setText(
-            this.selectedMacro.element.getText().replace(this.selectedMacro.macroText, macroString));
+        this.selectedMacro.element.$.innerText =
+            this.selectedMacro.element.$.innerText.replace(this.selectedMacro.macroText, macroString);
```
alansemenov, line 206: @ashklianko Is it `setText` or `getText` that does sanitisation in CKE? Can we keep `setText` and only replace `getText` with `$.innerText`?

ashklianko: setText and getText don't preserve line breaks, so just innerText works

alansemenov:
Will be fixed in https://github.com/enonic/app-contentstudio/pull/4109

PR #4109, opened 2022-01-13
Line breaks get removed from TextArea inside a macro config

alansemenov:


alansemenov:
Will be fixed in https://github.com/enonic/app-contentstudio/pull/4101

ashklianko (approved):

Issue #3245, opened 2021-06-22
Show spinner in the preview panel while preview is being loaded


alansemenov:
After a content with preview is selected in the grid, the panel gets cleared and preview starts loading. If it takes a long time to load (especially visible on slow connection for a heavy site), you'd just stare on an empty panel not sure if you'll eventually see the preview or if something went wrong. Show the spinner while the preview is being loaded.

alansemenov:
This is now fixed with recent preview changes

PR #4132, opened 2022-01-18
Show spinner in the preview panel while preview is being loaded #3245

ashklianko:


alansemenov:
This is now fixed with recent preview changes

Issue #4085, opened 2022-01-10
Display current editors of the content


alansemenov:
* In the ProjectViewer div decrease `padding`-right to 15px and remove `margin-right`
* Center-align "`<status> by <user-name>`" inside the toolbar
* Remove status icon (to the left of the workflow button dropdown)
* Hide "`...by <user-name>`" on resolutions <1200px
* Display the list of users currently having the content wizard open in the form of avatar icons. Use the same layout we have in UserAccessWidgetItemView where we use up the available space and then display "`+ X`" for remaining ones.
* Each avatar should have a tooltip displaying user's display name, and "`+ X`" should have a tooltip displaying the entire list of users (one display name per line).
* Current user's avatar should have a special styling/border assigned to it
* In mobile mode replace "More" (fold button's caption) with "...", remove left-padding of the fold button and make sure avatar icons are correctly aligned with enough space on the sides.

![image](https://user-images.githubusercontent.com/11612490/148758299-0ccb0c59-e193-40f3-90f0-72756840c66a.png)

PR #4196, opened 2022-02-07
Display current editors of the content #4085

ashklianko:


alansemenov (changes requested):
Found a possible bug:
1. I was logged in as User1 (SU) in Chrome, opened a content
2. Then I logged in as User2 (AL) in Firefox, opened the same content.

In Chrome I got AL added to SU in the toolbar, but in Firefox I had only AL, but not SU (from the first browser). See pic1.

3. Then I logged in as User3 (JD) in a 3rd window. I got AL and JD in the toolbar, but (again) not SU from the very first session.

4. Then I reloaded the page in the first browser (with SU). After that I finally got SU added in the browsers 2 and 3.

PS. While I was writing this review, all users disappeared from all browsers except those that were originally logged in (see pic 2). So it seems that over a short time, other users disappear from the store even though their sessions are still alive.

image
image

Issue #4071, opened 2022-01-05
Content-Security-Policy header


rymsha:
Similar to https://github.com/enonic/xp/issues/9293. 

All `admin/tool/com.enonic.app.contentstudio/*` endpoint responses must contain CSP HTTP header.
A good starting point for the value would be: `default-src 'self'; object-src 'none'; style-src 'self'`

This should mitigate XSS in CS itself.

Note, that CS has full control of it's content (vs preview endpoints) and the final header value could be very strict - allowing only content sources that are absolutely necessary.

alansemenov:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy

PR #4197, opened 2022-02-07
Content-Security-Policy header #4071

ashklianko:


ashklianko:
NB: breaks studio until all inline scripting is removed!

Issue #4219, opened 2022-02-10
Content grid is not properly refreshed after move/archive


alansemenov:
See the attached video.

To reproduce:
1. Make sure you have app-features deployed which creates Large Tree nodes
2. Create an empty folder ("Copy")
3. Select all items (100) under, f.ex. Large Tree Node 2
4. Move them to the empty folder

Everything looks good up to the point where modal dialog's progress is over, then strange behaviour ensues:
1. Moved items don't disappear for a very long while
2. While you scroll the list of moved items, the grid is empty
3. When finally the moved items disappear, 20 of them still remain under the original folder
4. When you click the refresh icon, then finally all of the moved items disappear

The end of the dialog's progress should indicate the absolute end of the entire moving process. When this is over, the grid should be refreshed and all of the moved items should be correctly displayed, without having to wait or refresh manually.

https://user-images.githubusercontent.com/11612490/153406140-a9e0d7b5-7874-4fc4-8a41-51783e280151.mov

alansemenov:
It's also very important that we don't refresh grid on any kind of timeout as the process duration varies based on how many items are being processed. When the server-side task reports completion, all the changes in storage are supposed to be finished and it should be safe to refresh the grid.

PR #4246, opened 2022-02-15
Content grid is not properly refreshed after move/archive #4219

ashklianko:
-when moving/deleting items we are receiving updates not only from current repo but also from all child layers where those changes are being replicated, thus our debounced server event handlers (that group events by type) couldn't fire. Fixed by separating events by not only a type but also a repo, so event triggers immediately when stops receiving calls from current repo


ashklianko:
Tested move on 400 items, items are moved in the grid immediately after move task over and progress dialog is closed

alansemenov (approved):

Issue #1518, opened 2020-02-19
Upgrade CKEditor to version 4.17.2


poi33:
Lots of vulnerabilities fixes since 4.11.1 we are currently on.

Check that all UI tests are passing and that our custom plugins and changes to existing plugins are not ruined.

https://github.com/ckeditor/ckeditor4/tags

alansemenov:
This should fix https://github.com/enonic/app-contentstudio/issues/1518

alansemenov:
This upgrade must be done before the next feature release (4.1.0).

PR #4257, opened 2022-02-18
Upgrade CKEditor to version 4.17.2 #1518

ashklianko:


ashklianko:
Codacy complains about cke files, may be skipped

alansemenov (changes requested):
Great stuff, but we definitely don't need the `samples` folder.

ashklianko:
will remove, also will add numbered list properties in this pull

ashklianko:
Done, list properties are available via context menu (right click on list)

alansemenov (approved):
Yippee!

Issue #937, opened 2019-09-09
Spinner should be shown over Issue Details dialog while items are being loading


sgauruseu:
1. Select a folder with about 50 children and create new publish request.
2. Open List Issues dialog and click on the request.
 
Spinner should be shown over Issue Details dialog

![image](https://user-images.githubusercontent.com/3728712/64534409-15384680-d31e-11e9-98e7-de4a32d6ef51.png)
 

![image](https://user-images.githubusercontent.com/3728712/64534474-3436d880-d31e-11e9-83da-c7f90cf8ace5.png)

PR #4278, opened 2022-02-25
Spinner should be shown over Issue Details dialog while content items…

ashklianko:
… are being loaded #937


alansemenov (changes requested):
Maybe move `this.isLoadInProgress = true;` inside `showLoadMask` and `this.isLoadInProgress = false;` inside `hideLoadMask`?

ashklianko:
Updated

alansemenov (approved):

Issue #4270, opened 2022-02-22
Custom selector won't filter by description


alansemenov:
When using filter in a Custom selector, it will only look in display name, not in description.

Unfiltered:
<img width="843" alt="image" src="https://user-images.githubusercontent.com/11612490/155087731-454ee13d-325b-4004-af93-f112f472c0b6.png">

Filtered by a keyword in description:
<img width="866" alt="image" src="https://user-images.githubusercontent.com/11612490/155087869-fa558a03-be9a-49a2-b593-a1d48386e050.png">

tajakobsen:
Is it possible to get this changed, so that the frontend doesn't do any filtering at all?

I put a lot of effort into writing my CustomSelector-service, and then the frontend just thinks it knows better then me what to show. 

I was very confused about that in the beginning.

--- 

_Example:_

Sometimes I might get hits in a query where the field data is not part of `displayName` or `description`, but I would want to display that hit to the user.

PR #4285, opened 2022-02-28
Custom selector won't filter by description #4270

ashklianko:
-removing frontend filtering to let service decide what result to filter and return and to avoid filtering by display name only


alansemenov (approved):
Perfect

PR #4293, opened 2022-03-03
Server-side error when archiving content #4291

ashklianko:
Issue is in that when archiving items we also receive update events prior to delete/move events, that triggers fetching of content that is already can't be found in the root

-improved event filtering: we use debounce for all types of events, so when it is time to execute update event handler we check if items in update queue are also can be found in the move/delete queue, and if they do then we remove those items from update queue
-Not executing lazy load handler in dependants list if dialog is closed

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/event/ServerEventAggregator.ts
+++ b/modules/lib/src/main/resources/assets/js/app/event/ServerEventAggregator.ts
@@ -29,7 +29,7 @@ export class ServerEventAggregator {
                 this.typesAndDelayedFunctions.get(typeAndRepo)();
             } else {
                 const debouncedFunc: Function = AppHelper.debounceWithInterrupt(() => {
-                    const items: ContentServerChangeItem[] = this.typesAndChangeItems.get(typeAndRepo);
+                    const items: ContentServerChangeItem[] = this.filterItems(typeAndRepo);
 
                     this.typesAndDelayedFunctions.delete(typeAndRepo);
                     this.typesAndChangeItems.delete(typeAndRepo);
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/app/event/ServerEventAggregator.ts
```diff
@@ -47,6 +47,41 @@ export class ServerEventAggregator {
...
+            return items.filter((item: ContentServerChangeItem) => this.isNotPendingMoveOrDeleteItem(item));
+        }
+
+        return items;
+    }
+
+    private isNotMoveOrDeleteEvent(typeAndRepo: string): boolean {
+        const eventType: string = typeAndRepo.substring(0, typeAndRepo.indexOf(':'));
+        return eventType !== NodeServerChangeType[NodeServerChangeType.MOVE]
+               && eventType !== NodeServerChangeType[NodeServerChangeType.DELETE];
```
alansemenov: Maybe
`return typeAndRepo.startsWith(NodeServerChangeType[NodeServerChangeType.MOVE]) || typeAndRepo.startsWith(NodeServerChangeType[NodeServerChangeType.DELETE]);`?

alansemenov: Or `const eventType: string = typeAndRepo.split(':')[0];`

ashklianko: Updated

ashklianko:
Updated, however 'Project node sync failed' keep coming from backend, not related to frontend

alansemenov:
Will be fixed in the backend

PR #4316, opened 2022-03-14
CustomSelector sends two identical requests

alansemenov:
* Store start index of next batch of items to load to prevent duplicate load of the first batch
* Small refactoring (use forced reload in `BaseRichComboBox`)


ashklianko:
Made some updates

Issue #4290, opened 2022-03-02
Support new widget interface for the Sidebar


alansemenov:
Widget selector in the Context panel is populated with internal widgets + external widgets with interface `contentstudio.contextpanel`.

We will use the same approach for the left sidebar menu also.

Content Studio tab (pencil icon) will be internal/hardcoded, the rest will be dynamic.

* Change studio to fetch widgets with interface `contentstudio.sidebar` (new interface) and dynamically inject tabs into the Sidebar (after the Content Studio's one).
* Use widget icons on the tabs, use widget display names for tooltips
* Sidebar should be refreshed whenever an app with widget(s) with the new interface is started or stopped
* Make Settings into a widget with the new interface. Make sure it's always the last tab in the Sidebar.
* Change Settings' HTML template if necessary, to make proper injection of HTML into Content Studio (similar to how we did with `<widget>` tags for ordinary widgets)
* Remember that this must be a robust and consistent solution that will be used for injection of Content Studio Plus, Guillotine app and potentially any "fullscreen" widget.

PR #4336, opened 2022-03-17
Support new widget interface for the Sidebar #4290

ashklianko:


modules/app/src/main/resources/admin/widgets/settings/settings.html
```diff
@@ -0,0 +1,3 @@
+<widget id="settings-widget">
```
alansemenov: Let's use `widget-settings`. This will make it consistent with "regular" widgets, and for external widgets we will use "`widget-<appId>`".

modules/app/webpack.config.js
```diff
@@ -68,20 +69,7 @@ module.exports = {
...
-                defaultVendors: {
-                    test: /[\\/]node_modules[\\/]/,
-                    reuseExistingChunk: true,
-                    minChunks: 2,
-                    priority: -10,
-                    filename: 'js/vendors.main~editor.js'
-                }
-            }
-        }
+        ]
```
alansemenov: This deletion now causes this:
image

modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -410,22 +412,37 @@ function getTheme(): string {
...
+    const baseAppToBeOpened = url.indexOf('#') < 0 ? baseApps[0] :
+                              baseApps.find((baseApp) => url.endsWith(`/${baseApp.getUrlPath()}`));
 
     if (baseAppToBeOpened) {
         commonWrapper.selectApp(baseAppToBeOpened);
     } else {
-        commonWrapper.onAppAdded((app) => {
-            if (!commonWrapper.hasAppSelected() && url.endsWith(`/${app.getUrlPath()}`)) {
-                commonWrapper.selectApp(app);
+        commonWrapper.onItemAdded((item) => {
```
alansemenov: Use `(item: App | Widget)` and check for instance type instead of checking for properties that might disappear or be renamed (`item['getUrlPath']`).

modules/lib/src/main/resources/assets/js/app/util/WidgetHelper.ts
```diff
@@ -0,0 +1,51 @@
...
+        document.getElementsByTagName('head')[0].appendChild(scriptNode);
+
+        return scriptNode;
+    }
+
+    static injectWidgetElementInto(widgetContainer: HTMLElement, target: UIElement): UIElement {
+        const widgetEl: HTMLElement = widgetContainer.querySelector('widget');
+
+        if (!widgetEl) {
+            return;
```
alansemenov: Should probably throw an exception or give a proper toast error if someone is trying to inject a widget without proper formatting (`<widget>` tag).

modules/lib/src/main/resources/assets/js/app/util/WidgetHelper.ts
```diff
@@ -0,0 +1,51 @@
+import {Element as UIElement} from 'lib-admin-ui/dom/Element';
+
+export class WidgetHelper {
+
+    static injectWidgetHtml(html: string, target: UIElement): { scriptElements: HTMLScriptElement[], resultElement: UIElement } {
+        const widgetContainer: HTMLHtmlElement = document.createElement('html');
+        widgetContainer.innerHTML = html;
```
alansemenov: Create an Element using `UIElement.fromHtmlElement` as early as possible, then use `Element.setHtml()` to ensure HTML gets escaped.

alansemenov (changes requested):

alansemenov:
A few more things to fix:
* Change attributes for widget script to be data attributes with all-lower-case names
```html
<widget id="widget-settings">
    <script src="{{assetsUri}}" data-config-service-url="{{configServiceUrl}}" data-widget-id="widget-settings"></script>
</widget>
```
* In lib-admin-ui's `Element.ts` implement a version of `fromString` method which doesn't sanitize the string:

```Javascript
    static fromString(s: string, loadExistingChildren: boolean = true): Element {
        const sanitizedHtml: string = DOMPurify.sanitize(s);
        return Element.fromHtml(sanitizedHtml, loadExistingChildren);
    }

    static fromHtml(html: string, loadExistingChildren: boolean = true): Element {
        const htmlEl = $(html).get(0);

        if (!htmlEl) {
            return null;
        }

        let parentEl: Element;

        if (htmlEl && htmlEl.parentElement) {
            parentEl = Element.fromHtmlElement(htmlEl.parentElement);
        }

        return this.fromHtmlElement(htmlEl, loadExistingChildren, parentEl);
    }
````

* Use this new method inside `injectWidgetHtml`. We don't really need sanitisation since we now enforce Content Security Policy in Content Studio. Note validation of the root element to have `<widget>` tag.

```Javascript
    static injectWidgetHtml(html: string, target: UIElement): { scriptElements: HTMLScriptElement[], widgetContainer: UIElement } {
        const widgetContainer: UIElement = UIElement.fromHtml(html);
        if (widgetContainer?.getHTMLElement().tagName !== 'WIDGET') {
            throw 'Widget contents must be wrapped inside <widget></widget> tags';
        }

        const scriptElements: HTMLScriptElement[] = WidgetHelper.injectScriptsToHead(widgetContainer);
        target.appendChild(resultElement);

        return {scriptElements, widgetContainer};
    }
````

* `injectWidgetElementInto` (interesting name) is no longer needed.

ashklianko:
Updated

alansemenov (approved):

Issue #4347, opened 2022-03-23
Add rendering mode param to IsRenderableRequest


pmi:
This will help rendering server to know whether to render catch-all page (every mode except edit) or not (edit mode, to allow selection of descriptor)

PR #4348, opened 2022-03-23
Add rendering mode param to IsRenderableRequest #4347

pmi:


alansemenov (changes requested):
@pmi I don't like the idea of always sending the `mode` parameter, even if it's not used. Let's make `mode` an optional argument in the constructor, remove the default value and append `mode` to params only when it's set.

sigdestad:
Any reason why we don't make this a header instead?

alansemenov:
> Any reason why we don't make this a header instead?

Because we are already making Content Studio & NextJS starter tightly coupled.

sigdestad:
I'm just saying that a parameter is more intrusive than header, next.js should eat whatever we throw at it. Anyway, this is a temporary workaround until page editor gets more flexible - right?

alansemenov:
> Anyway, this is a temporary workaround until page editor gets more flexible - right?

Until XP stops returning 200 for a page that cannot in fact be edited, yes.

alansemenov (approved):

Issue #4379, opened 2022-03-30
Sidebar widgets: implement common header for widgets


ashklianko:
Add common header with Project Selector / Display name, widgets content will be loaded under it

PR #4387, opened 2022-04-04
Sidebar widgets: implement common header for widgets #4379

ashklianko:
-Reused existing code without major refactoring
-ContentAppBar is made singleton and used as a common header for all widgets, additional public api methods added
-Showing app/widget name right after project selector, and hiding it on small screens (except for Settings)
-Hiding project selector for a Settings widget to make it look as before

Potential problems:
-ContentAppBar uses ShowIssuesDialogButton in header for main studio app, had to hide it when other widgets selected. Potential fix is to move it somewhere else inside studio app
-Settings app (now widget) used to have panel tabs for creating/editing layers within it's app bar, but now we're gonna have common ContentAppBar, so had to make ContentAppBar inherit TabbedAppBar to let settings show/hide it's tabs. Possibly might be a problem if some other widget will also have panel tabs. Potential fix is to attach Settings tabs to other element, like actions toolbar for example. That also requires keeping app name interactable for Settings to preserve behavior of switching between opened tabs in Settings browse panel


ashklianko:
Also fixed Sidebar menu for small screens, tested everything on small screens also

modules/lib/src/main/resources/assets/styles/browse/content-app-panel.less
```diff
@@ -1,3 +1,3 @@
 .content-app-panel {
-  top: 44px;
+  /* top: 44px !important; */
```
alansemenov: Remove the file?

alansemenov (changes requested):
I _really_ like the new concept and new DOM layout after these changes. But we need to make sure there's no blinking of default elements when a widget is loaded directly via URL. See below how opening the Settings widget directly looks like on a slightly delayed bandwidth.

So basically, there should be no default project selector which gets hidden and there should be no default "Content Studio" in the header which gets replaced by the actual widget name.

https://user-images.githubusercontent.com/11612490/161774327-e216e3ec-cedd-4f8f-b267-45ec5347eab2.mov

modules/lib/src/main/resources/assets/styles/browse/content-app-panel.less
ashklianko: Will do

ashklianko: Regarding blinking: ContentAppBar knows nothing about loaded widgets, and it is a responsibility of a widget to hide/show viewer on demand when they get loaded. However I can remove initial blink by not showing viewer and app name until first widget gets selected.

ashklianko: Removed extra blink until first widget selected. Further improvements will require additional investigation and refactoring

alansemenov (changes requested):
I played around a bit with it and I think we need to make some conceptual changes. The problem is that classes deriving from `AppWrapper` are trying to be too smart.

For example, `SettingsAppWrapper` itself shows and hides the Project Selector which is wrong - it's not supposed to know about it at all. What if we switch from Settings to another widget that is not supposed to have the Project Selector either? `SettingsAppWrapper` calls `showProjectSelector` on its hide, which means that the new widget should again call `hideProjectSelector`, which will again result in blinking.

What we should do is implement config for AppWrapper which will contain properties for showing tabs, project selector, issues button etc. in the AppBar (all of them should be hidden by default). Each of the classes deriving from AppWrapper will specify in its config whether it needs tabs, project selector, issues button etc., and AppBar will show these elements based on the config.

When the widget gets hidden/switched from, AppWrapper should hide everything inside AppBar and then show elements based on the config of the newly selected widget.

alansemenov:
Not sure if that is possible but as en even better solution, instead of clearing AppBar when switching between widgets, we could compare configs of the current widget and the new one, and don't hide elements that are enabled in both configs.

alansemenov (approved):
Works like a charm. Great job!

Issue #4388, opened 2022-04-04
Implement feature flag for displaying concurrent editors


alansemenov:
Displaying of concurrent editors is an experimental feature and will in its first revision be **disabled by default**. Implement a new property in the app config file called `contentWizard.enableCollaboration` that can be used to enable the feature for those who want to experiment with it. If this property doesn't exist in the config or is set to anything but `true`, the feature should be disabled and Content Wizard header should look the same as before the changes. 

Make a PR to [this branch](https://github.com/enonic/app-contentstudio/tree/epic_collab).

Selenium tests in the [epic PR](https://github.com/enonic/app-contentstudio/pull/4378) are now failing but in the new PR for the feature flag all tests must pass since the feature will be disabled by default.

PR #4393, opened 2022-04-04
Implement feature flag for displaying concurrent editors #4388

ashklianko:


alansemenov (changes requested):
* The "default" layout is missing workflow icon to the left of the publish button. It's ok to always keep the modifier centered, but workflow icon should be shown in disabled collaboration mode as there's enough space in the toolbar for it. 

You can see that Selenium tests are failing because they cannot find the icon. If everything is done properly, the tests will pass.

Here's the new version with disabled collaboration mode on top, and the old version (before the feature was implemented) at the bottom:
image

* Also something is wrong in the mobile mode, probably also because of the missing workflow icon:
image

ashklianko:
Updated

alansemenov (approved):
Nice!

Issue #4405, opened 2022-04-06
Image wizard laptop resolution -  image toolbar is overlapping options in the wizard's toolbar dropdown


sgauruseu:
1. Create a site, then add new page-template and select Image in the support
2. Upload an image to the site.
3. Select the image and click on `Edit` button.
4. Click on `Show Page Editor` button.
5. Click on the wizard toolbar menu

BUG:  image toolbar is overlapping options in the expanded menu

![image](https://user-images.githubusercontent.com/3728712/161981440-be05e7aa-077b-46cb-a73f-994e0725fd5c.png)

alansemenov:
Most likely a simple z-index fix

PR #4420, opened 2022-04-11
Fixed image toolbar overlapping dropdown options (#4405)

reisfmb:


ashklianko:
One more thing to fix: when in the edit mode and some changes done to the image, reset button makes toolbar wrap on next line:
image
Update styles so toolbar doesn't wrap on next line

ashklianko (changes requested):
Need to fix toolbar

ashklianko:
Moved latest issue to the https://github.com/enonic/app-contentstudio/issues/4435

Issue #4435, opened 2022-04-12
Image Editor: move action buttons to the bottom toolbar


ashklianko:
On small screens action buttons in the editor's toolbar might be wrapped on next line, overlapping active working zone of the editor tools. This problem will evolve when new editor tools added. See a picture below:

![image](https://user-images.githubusercontent.com/4255706/162973900-fb0e515a-ac55-424c-b1a5-43238f657934.png)

To fix:
Add a bottom toolbar with action buttons moved into it:

![image](https://user-images.githubusercontent.com/4255706/162976157-b52e7c2b-b86b-418c-b1c2-c6a359fe650b.png)

alansemenov:
If we do that, Apply and Cancel buttons must swap places and Reset should be left-aligned.

sgauruseu:
Image Editor Tollbar  should not move when the image is cropping 


https://user-images.githubusercontent.com/3728712/163144413-1a4a0caf-4e2b-4111-813c-b081d99687a2.mp4

sgauruseu:
I updated ui-tests, but one test is still failing due to the next issue:
`Reset Mask` should not be visible when the edit mode is closed.

![image](https://user-images.githubusercontent.com/3728712/163144562-6291bdc0-0e12-456a-80e9-5355f1d8aa75.png)

reisfmb:
@sgauruseu 

> Image Editor Tollbar should not move when the image is cropping

Still discussing with Sasha the expected behavior of the bottom toolbar in crop mode.

> I updated ui-tests, but one test is still failing due to the next issue: `Reset Mask` should not be visible when the edit mode is closed.

This was fixed [here](https://github.com/enonic/app-contentstudio/commit/5252ae050c19b56f2345c0b84296193bd2e3ef75)

alansemenov:
Will be solved by https://github.com/enonic/app-contentstudio/issues/4557

PR #4436, opened 2022-04-12
Moved action buttons to a new bottom toolbar (#4435)

reisfmb:


reisfmb:
@ashklianko `testInputTypes` have failed because now there's no "close button", only a gray "cancel button" which does not have the class that is used by the test. How to proceed?

ashklianko:
> @ashklianko `testInputTypes` have failed because now there's no "close button", only a gray "cancel button" which does not have the class that is used by the test. How to proceed?

@sgauruseu will update test

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -788,6 +751,11 @@ export class ImageEditor
         this.uploadButton.setEnabled(false).addClass('button-upload');
         standbyContainer.appendChildren(this.resetButton, this.uploadButton);
 
+        let rightContainer = new DivEl('right-container');
```
ashklianko: 1. Don't hesitate using const, thus we declare that variable is not going to change
2. Always set type of a variable, not only in new code, but also when you update some piece of code where vars are missing their types or have `let` instead of `const`

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -788,6 +751,11 @@ export class ImageEditor
...
+        rightContainer.appendChildren(standbyContainer/* , editContainer */);
+
+        let zoomContainer = this.createZoomContainer();
```
ashklianko: Same as above

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -856,6 +819,53 @@ export class ImageEditor
         return toolbar;
     }
 
+    private createBottomToolbar(): DivEl {
+        let toolbar = new DivEl('bottom-toolbar');
```
ashklianko: Same: `const` instead of `let`, and variable types

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -856,6 +819,53 @@ export class ImageEditor
...
+
+        let editContainer = new DivEl('edit-container');
```
ashklianko: Same as above

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -856,6 +819,53 @@ export class ImageEditor
...
+            if (this.isFocusEditMode()) {
+                this.resetFocusPosition();
+                this.resetFocusRadius();
+            } else if (this.isCropEditMode()) {
+                this.resetZoomPosition();
+                this.resetCropPosition();
+            }
+        });
+
+        let applyButton = new Button(i18n('button.apply'));
```
ashklianko: Same as above

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -856,6 +819,53 @@ export class ImageEditor
...
+            event.stopPropagation();
+
+            if (this.isCropEditMode()) {
+                this.disableCropEditMode();
+            } else if (this.isFocusEditMode()) {
+                this.disableFocusEditMode();
+            }
+        });
+
+        let cancelButton = new Button(i18n('action.cancel'));
```
ashklianko: Same as above

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
reisfmb: My bad here. I extracted the code from the top toolbar to create the bottom one, so I wasn't paying attention to those details. Thanks!

ashklianko:
One more thing: lets replace crop drag handler (circle) that is might not be seen on the image depending on it's color with the next drag control on the bottom toolbar:

image

Use class `icon-arrow-top` on the element representing button (div or whatever), with `@admin-medium-gray-border` color

reisfmb:
Remains to update position of the drag handler in crop mode. It should be inside the bottom toolbar. Will do that as 1st thing on Monday.

alansemenov (changes requested):
Not exactly right. 
1. The drag control should be a circle with two arrows facing up and down. See below for how it was before.
2. When you drag the control up/down, the footer should stay put - it's only the line with the control that should change position.
Screenshot 2022-04-18 at 23 56 28

ashklianko:
@alansemenov we've tried with keeping bottom toolbar at place, that didn't look good, also drag handler may blend into the background

reisfmb:
@alansemenov just to add more context...

1. Opted to go with Sasha's suggestion since the drag handler might blend with the image if it isn't fully in the bottom toolbar. 

2. Last week I recorded a video and asked Sasha's opinion on it... we both agreed that it didn't look good to have the bottom toolbar at place.

https://user-images.githubusercontent.com/67838246/164074061-fb94fcdf-24ee-44d3-a3d6-f9bd607e31df.mp4

How to proceed?

alansemenov:
Ok, I agree about the moving bottom toolbar, but dragger has to be changed. The argument that it might blend with the image doesn't stand when you look at the 50 shades of grey in the image below. I had a really hard time realising the dragger was even there. Let's make it the way it was (see a screenshot above) and invert the colors inside, so that the circle is dark and arrows inside are the same grey color as the toolbar.

image

alansemenov:
Made a new issue: https://github.com/enonic/app-contentstudio/issues/4557

PR #4480, opened 2022-04-22
Websocket connection fails with current CSP

alansemenov:
Also changed styling of avatar icons in the collaboration block.


ashklianko:
Works fine on IPhone Safari:
File

ashklianko (approved):

PR #4486, opened 2022-04-25
Implement filtering by workflow state #4468

ashklianko:
-Added some missing on frontend filters


ashklianko:
ContentBrowseFilterPanel requires heavy refactoring, it was not touched for ages and in general search is not optimized for adding new aggregations

alansemenov (changes requested):
1. We should make sure that none of the Workflow states are hardcoded. 
For example, `WorkflowState[WorkflowState.READY].toLowerCase()` will give you 'ready' and so on. You can also create an instance of the `Workflow` class and then you can use `getStateAsString()` to get string value.
2. Use existing `status.workflow.in_progress` and `status.workflow.read` literals for texts, no reason for new translations. You can also get correct literal id by concatenating `status.workflow.` with `Workflow.getStateAsString()`

Issue #4466, opened 2022-04-21
Search improvements


alansemenov:
We need to add some additional parameters to the filter panel:
1. `Workflow`:
* _In progress_
* _Ready_
2. `Modified by` (add after `Last modified`):
* _Me_ (current user)
* _User(s)_: a dropdown with aggregation of modifiers of content in the data set
3. `Owner`
* _Me_ (current user)
* _User(s)_: same as above
4. `Language` (aggregation on languages, sorted by the most used ones in descending order)

Note that the new properties should have aggregated numbers in brackets, like current properties do. Ideally for the user dropdowns as well - we should aggregate the dataset by modifier/owner, then build the dropdown with number of items in brackets for each user. 

![image](https://user-images.githubusercontent.com/11612490/164431127-0bbb8786-f692-4c97-9b92-d99615ba9c01.png)

PR #4503, opened 2022-04-26
Implement filtering by workflow state

ashklianko:


alansemenov (changes requested):
Duplicating review from the closed PR here:

1. We should make sure that none of the Workflow states are hardcoded. 
For example, `WorkflowState[WorkflowState.READY].toLowerCase()` will give you 'ready' and so on. You can also create an instance of the `Workflow` class and then you can use `getStateAsString()` to get string value.
2. Use existing `status.workflow.in_progress` and `status.workflow.read` literals for texts, no reason for new translations. You can also get correct literal id by concatenating `status.workflow.` with `Workflow.getStateAsString()`

ashklianko:
Updated, also made a first portion of refactoring. Modifier field added

alansemenov (changes requested):
Last commit is linked to the issue of adding filtering by Owner, but in fact adding filtering by Modifier... It's not the same thing, these are two different properties.

Also, please read the specs of how filters by Owner and Modifier should look like: https://github.com/enonic/app-contentstudio/issues/4466. It should not be aggregation checkboxes like for other fields, it should be a radio group with two radio "Me" (Current user) and a dropdown with aggregations.

To make it a bit easier, we can have the checkboxes but the first one should always be "Me" and represent current user.

Workflow is fine now.

ashklianko:
@alansemenov, update on PR:
- All required aggregations added
- Currently new aggregations may have just one selection,  ContentBrowseFilterPanel logic is to be reworked to allow multiple selection, current logic is extensive and requires multiple queries requests, will work on it soon
- After multiple selection for new aggregations will start working I will add aggregation dropdowns as mentioned in issue's epic

modules/lib/src/main/resources/assets/js/app/browse/filter/AggregationsProcessor.ts
```diff
@@ -0,0 +1,141 @@
...
+        const workflowAggr: Aggregation = aggregations.find((aggr: Aggregation) => aggr.getName() === ContentAggregations.WORKFLOW);
+
+        if (workflowAggr) {
+            this.updateWorkflowAggregation(<BucketAggregation>workflowAggr, total);
+        }
+    }
+
+    private updateWorkflowAggregation(workflowAggr: BucketAggregation, total: number): void {
+        // contents might not have a workflow property, thus aggregation won't see those contents, but they are treated as ready
+        const inProgressKey: string = WorkflowState[WorkflowState.IN_PROGRESS].toLowerCase();
```
alansemenov: `WorkflowState[WorkflowState.IN_PROGRESS].toLowerCase()`

As we seem to use this all over the place and since `Workflow.getStateAsString` is not used at the moment, I suggest we change it like this:

```Typescript
    static getStateAsString(state: WorkflowState): string {
        return WorkflowState[state].toLowerCase();
    }
```

and start using `Workflow.getStateAsString(WorkflowState.IN_PROGRESS)`

ashklianko: Or since we're going to modify lib-admin-ui then we can do this:
```
export enum WorkflowState {
    READY = 'ready', IN_PROGRESS = 'in_progress'
}
```

and remove `getStateAsString` at all

alansemenov: Yeah, this is even better. I assume we are not sending numeric enum values to the API.

alansemenov (changes requested):
image

1. Current user or Me should not be in the user selector as we already have it above. Having it in both places is confusing.
2. If there are no users inside the selector (for example, current user is owner of all items), then selector should not be shown, only the "Me" checkbox.
3. Font inside the selector should be the same as for the checkboxes above the selector.
4. OK button should have the blue background instead of green.

ashklianko:
> image
> 
> 1. Current user or Me should not be in the user selector as we already have it above. Having it in both places is confusing.
> 2. If there are no users inside the selector (for example, current user is owner of all items), then selector should not be shown, only the "Me" checkbox.
> 3. Font inside the selector should be the same as for the checkboxes above the selector.
> 4. OK button should have the blue background instead of green.

Ok for all points except 1) - it is much more handy to let user deselect multiple items at once, including Me, as in the list above, and it is also consistent with other options selected and you can select/deselect them in both list above and selector

alansemenov (approved):

Issue #4557, opened 2022-05-13
Broken layout in the "Image Editor" toolbar in mobile resolution


alansemenov:
Originally https://github.com/enonic/app-contentstudio/issues/4435

On **mobile resolution**:
* Change "Apply" button to have "Ok" text on it
* Make buttons the same size
* Change link text to only say "Reset"
 
![image](https://user-images.githubusercontent.com/11612490/168307726-cb7e9f4e-cd28-4167-9daa-a73b15422800.png)

PR #4570, opened 2022-05-17
Fixed Image Editor mobile layout (#4557)

reisfmb:


modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
             }
         });
 
+        const editContainerButtonsResponsiveHandler = () => {
+            const item = new ResponsiveItem(this);
```
alansemenov: Make this a private property of the class to avoid creating a new ResponsiveItem every time the handler is called.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+
+            if(item.isInRangeOrSmaller(ResponsiveRanges._360_540)) {
+                applyButton.setLabel('Ok');
+                applyButton.setTitle('Ok');
```
alansemenov: Never hardcode localisable texts. Use `i18n('action.ok')`. Also, use `const okLabel: string = i18n('button.reset')`, then use `okLabel` in the two next lines to avoid calling `i18n` twice.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+                applyButton.addClass('ok-button');
```
alansemenov: You should not need a separate CSS class. If you want to apply certain CSS formatting for a specific resolution, this can be done directly in CSS. In this case, since your `ResponsiveItem` is the `ImageEditor` div, it will have a CSS class like '`_360-540`' assigned to it, which shows current width range of the element.

So, add _once_ a CSS class to the "Apply" button when it's created, like 'apply-button':

`const applyButton: Button = new Button(i18n('button.apply')).addClass('apply-button');`

Then, in CSS (`image-editor.less`) you can simply do:

```CSS
.image-editor {
  padding: 0;
  &._360-540 {
     .apply-button {
         padding: 0 6px;
     }  
  }
}
```

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+
+                this.editResetButton.setLabel(i18n('button.reset'));
+                this.editResetButton.setTitle(i18n('button.reset'));
+                this.editResetButton.getHTMLElement().style.padding = '0px';
```
alansemenov: Never apply inline styling in the code.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+            } else {
+                applyButton.setLabel(i18n('button.apply'));
+                applyButton.setTitle(i18n('button.apply'));
```
alansemenov: See the comment above about calling `i18` once.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+                applyButton.removeClass('ok-button');
+
+                let resetLabel = this.getEditResetLabel();
```
alansemenov: `const`

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
...
+
+                this.editResetButton.setLabel(resetLabel);
+                this.editResetButton.setTitle(resetLabel);
+                this.editResetButton.getHTMLElement().removeAttribute('style');
```
alansemenov: Should not be needed.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -752,6 +754,32 @@ export class ImageEditor
             }
         });
 
+        const editContainerButtonsResponsiveHandler = () => {
+            const item = new ResponsiveItem(this);
+
+            if(item.isInRangeOrSmaller(ResponsiveRanges._360_540)) {
```
alansemenov: Since `editContainerButtonsResponsiveHandler` can be called numerous times on resolution changes, it will make sense to store current mode in a variable and compare it with a new mode to check if the mode has changed and any changes are required. This way we avoid setting labels/titles on controls when they should be kept the same.

Something like this:

```Typescript
this.responsiveItem = new ResponsiveItem(this);
let wasMobileMode: boolean;

const editContainerButtonsResponsiveHandler = () => {
  const isMobileMode: boolean = this.responsiveItem.isInRangeOrSmaller(ResponsiveRanges._360_540);
  if (wasMobileMode === isMobileMode) {
    return;
  }
  wasMobileMode = isMobileMode;
  if (isMobileMode) {
    // do stuff for mobile mode
  } else {
   // do stuff for desktop mode
  }
};

ResponsiveManager.onAvailableSizeChanged(this, editContainerButtonsResponsiveHandler);

```

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -2285,4 +2315,17 @@ export class ImageEditor
     private notifyImageError(event: UIEvent) {
         this.imageErrorListeners.forEach(listener => listener(event));
     }
+
+    private getEditResetLabel(): string {
+        const item = new ResponsiveItem(this);
```
alansemenov: Use class property (`this.responsiveItem`) that you will create in `createStickyToolbar`

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -2285,4 +2315,17 @@ export class ImageEditor
...
+
+        let resetLabel = i18n('button.reset');
+
+        if(item.isInRangeOrBigger(ResponsiveRanges._360_540)){
+            if(this.isFocusEditMode()) { resetLabel = i18n('editor.resetautofocus'); }
+            if(this.isCropEditMode()) { resetLabel = i18n('editor.resetmask'); }
```
alansemenov: Don't inline condition body, it should be on a separate line even if it's one line.

```Typescript
if (this.responsiveItem.isInRangeOrBigger(ResponsiveRanges._360_540)){
  if (this.isFocusEditMode()) {
    return i18n('editor.resetautofocus');
  }
  if (this.isCropEditMode()) {
    return i18n('editor.resetmask');
  }
}

return i18n('button.reset');
```

alansemenov (changes requested):
"Reset filters" won't always change to "Reset". Try rotating the image (3rd icon in the toolbar) - the link will say "Reset filters", but should say "Reset"
image

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -718,20 +723,25 @@ export class ImageEditor
...
-            }
-        });
+                if (this.isFocusEditMode()) {
+                    this.resetFocusPosition();
+                    this.resetFocusRadius();
+                } else if (this.isCropEditMode()) {
+                    this.resetZoomPosition();
+                    this.resetCropPosition();
+                }
+            });
```
alansemenov: Some extra tabs here? We should never use tabs for indentation.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/selector/image/ImageEditor.ts
```diff
@@ -718,20 +723,25 @@ export class ImageEditor
...
 
-        let applyButton = new Button(i18n('button.apply'));
-        applyButton.addClass('blue').onClicked((event: MouseEvent) => {
+        const applyButton = new Button(i18n('button.apply'));
+        applyButton.addClass('apply-button').addClass('blue');
```
alansemenov: You can set multiple classes in one go: 

`applyButton.addClass('apply-button blue');`

modules/lib/src/main/resources/assets/styles/inputtype/selector/image/image-editor.less
```diff
@@ -1,6 +1,17 @@
...
+  &._360-540 {
+    .apply-button {
+      padding-left: 6px;
+      padding-right: 6px;
+    }
+    
+    .reset-button {
+      padding: 0px;
+    }
+  }
```
alansemenov: Can you move this to the end of the class? We usually have responsive styling at the bottom.

modules/lib/src/main/resources/i18n/phrases.properties
```diff
@@ -69,6 +69,7 @@ action.archiveMore=Archive...
 action.fold=Fold
 action.show.more=Show more
 action.show.less=Show less
+action.ok=Ok
```
alansemenov: No need to add a new one. We already have it in lib-admin-ui's common.properties and they are concatenated with Content Studio's.

alansemenov (changes requested):

alansemenov (approved):

Issue #3377, opened 2021-08-04
Can't access content permissions via keyboard


thomasheartman:
When editing a piece of content, I am unable to open the permissions dialog using the keyboard only. The content ribbon contains elements that react to mouse clicks, but they're divs, meaning they're not focusable by keyboard. Arguably, they should be `<button>`s instead.

Here's the ribbon and the elements in question:

![deployment-open-permissions](https://user-images.githubusercontent.com/17786332/128172593-1ee6d913-2991-4949-ba17-d1e6b1a4766d.png)

PR #4589, opened 2022-05-24
Added accessibility to steps tab items (#3377)

reisfmb:


modules/lib/src/main/resources/assets/styles/wizard/content-wizard-panel.less
```diff
@@ -138,10 +138,10 @@
 
     .edit-permissions-button {
       order: 3;
-      padding: 15px;
+      //padding: 15px;
       cursor: pointer;
       z-index: 1;
-      line-height: @toolbar-height;
+      //line-height: @toolbar-height;
```
ashklianko: It is always better to not leave commented stuff that pollutes code, just delete it if it is not needed

modules/lib/src/main/resources/assets/styles/wizard/content-wizard-panel.less
```diff
@@ -138,10 +138,10 @@
 
     .edit-permissions-button {
       order: 3;
-      padding: 15px;
+      //padding: 15px;
```
ashklianko: It is always better to not leave commented stuff that pollutes code, just delete it if it is not needed

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,7 +1229,24 @@ export class ContentWizardPanel
         this.settingsWizardStep = new ContentWizardStep(i18n('field.settings'), this.settingsWizardStepForm, 'icon-wrench');
         steps.push(this.settingsWizardStep);
 
-        return steps;
+        const editPermissionsIcon = this.canEveryoneRead(this.getPersistedItem()) ? 'icon-unlock' : 'icon-lock';
```
ashklianko: 1) `editPermissionsIconClass` or just `iconClass` would be more self-describing name for a variable
2) Type is missing

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,7 +1229,24 @@ export class ContentWizardPanel
...
+        this.editPermissionsToolbarButton = new ContentWizardStep(i18n('field.access'), this.settingsWizardStepForm, editPermissionsIcon);
+        this.editPermissionsToolbarButton.getTabBarItem().addClass('edit-permissions-button');
+        this.editPermissionsToolbarButton.getTabBarItem().onClicked(this.handleEditPermissionsButtonClicked.bind(this));
+        steps.push(this.editPermissionsToolbarButton);
+
+        return this.addAccessibilityToSteps(steps);
```
ashklianko: You're not actually creating some new steps, so it is better to add accessibility separately to avoid confusion
```
this.addAccessibilityToSteps(steps);
return steps;
```

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,7 +1229,24 @@ export class ContentWizardPanel
...
+    }
+
+    private addAccessibilityToSteps(steps: ContentWizardStep[]) {
```
ashklianko: add `void` as a function return type

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,7 +1229,24 @@ export class ContentWizardPanel
...
+        return steps.map(step => {
```
ashklianko: replace `map` with `forEach`

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,7 +1229,24 @@ export class ContentWizardPanel
...
+            const stepTabBarItem = step.getTabBarItem();
```
ashklianko: 1) Create a separate function `addAccessibilityToAStep(...)` and use it in `forEach` cycle

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,9 +1229,29 @@ export class ContentWizardPanel
...
+        this.editPermissionsToolbarButton.getTabBarItem().onClicked(this.handleEditPermissionsButtonClicked.bind(this));
+        steps.push(this.editPermissionsToolbarButton);
+
+        this.addAccessibilityToSteps(steps);
+
         return steps;
     }
 
+    private addAccessibilityToSteps(steps: ContentWizardStep[]): void {
+        steps.forEach(step => this.addAccessibilityToAStep(step));
```
ashklianko: Almost there:
`(step: ContentWizardStep) => this.addAccessibilityToAStep(step)`

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,9 +1229,29 @@ export class ContentWizardPanel
...
+    }
+
+    private addAccessibilityToAStep(step: ContentWizardStep): void {
+        const stepTabBarItem = step.getTabBarItem();
```
ashklianko: `const stepTabBarItem: ContentTabBarItem = step.getTabBarItem();`

ashklianko (changes requested):

ashklianko (approved):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1234,9 +1229,29 @@ export class ContentWizardPanel
...
+        this.addAccessibilityToSteps(steps);
+
         return steps;
     }
 
+    private addAccessibilityToSteps(steps: ContentWizardStep[]): void {
+        steps.forEach((step: ContentWizardStep) => this.addAccessibilityToAStep(step));
+    }
+
+    private addAccessibilityToAStep(step: ContentWizardStep): void {
```
alansemenov: Remove "A". `addAccessibilityToAStep` -> `addAccessibilityToStep`

alansemenov (changes requested):
The wizard form is not correct after these changes. 

1) There should be no "Access" section in the form (permissions are managed via modal dialog)
2) What is now shown under "Access" (`Language` and `Owner`) should be under "Settings"

To put it simply, "Access" icon should be in the toolbar but not as a separate step in the form.

Before the changes:
image

After the changes:
image

alansemenov (approved):

Issue #4585, opened 2022-05-23
Don't show image selector in the Insert Image dialog until selected image is loaded


alansemenov:
When "Insert image" dialog is opened for an existing image (f.ex. when double-clicking an image inside the editor), the image selector with no selection will be shown for a while, until the selected image is loaded. In this case don't show the selector until the image is loaded and correct option is preset in the selector.

PR #4590, opened 2022-05-24
Don't show image selector in the Insert Image dialog until selected image is loaded

alansemenov:
Hide Image Selector in the Insert Image dialog until image is preloaded (#4585)


ashklianko:
Now there is a lag between opening preset image and while it is loaded we see blank space instead of a load mask. I will update

ashklianko (approved):
Just added some more changes to show load mask

Issue #4602, opened 2022-05-26
Empty aggregation buckets should not be shown in the Filter Panel


sgauruseu:
1. Create an `user1` with roles (Admin console + admin content)
2. Sign in the user and create a content
3. Login with su/password and click on `user1` checkbox in aggregation view

BUG- Owner block gets empty

![image](https://user-images.githubusercontent.com/3728712/170483061-e3403f89-8427-4b91-83db-c601f15bb51c.png)

The similar case with SU : 

![image](https://user-images.githubusercontent.com/3728712/170483209-7b9e031e-3ca1-4c51-83db-5c518e2d22f2.png)

alansemenov:
It's not only for Owner, same problem with Last Modified by. Changed the issue title.

![image](https://user-images.githubusercontent.com/11612490/171359927-f03e0d79-44b4-4f3d-a3aa-06d68cbfd1f2.png)

PR #4623, opened 2022-06-01
Empty aggregation buckets should not be shown in the Filter Panel #4602

ashklianko:


modules/lib/src/main/resources/assets/js/app/browse/filter/FilterableBucketAggregationView.ts
```diff
@@ -88,6 +87,10 @@ export class FilterableBucketAggregationView
         this.bucketListBox.clearItems();
     }
 
+    protected isToBeShown(): boolean {
+        return super.isToBeShown() || this.bucketListBox.getItemCount() > 0;
```
alansemenov: https://github.com/enonic/lib-admin-ui/pull/2607#discussion_r886757444

alansemenov (changes requested):

alansemenov (approved):

Issue #4630, opened 2022-06-03
Incorrect collapse/expand behavior of item-set inside a component config


alansemenov:
We have a convenient collapse/expand behaviour of an item-set inside a content type schema (can be tested on [this content type](https://github.com/enonic/app-features/blob/master/src/main/resources/site/content-types/minimum-occurrence-form-item-set/minimum-occurrence-form-item-set.xml) in app-features):

1. A new occurrence of an item-set (or all occurrences in a new content) is always expanded.
2. Existing occurrence of an item-set is by default collapsed.
3. Content Wizard remembers collapse/expand state of each occurrence: if user expands or collapses an occurrence and then saves the content, the occurrences will remain in the same state after save.

Unfortunately, this behaviour doesn't work when item-set is inside a part or layout config (can be tested with [this part](https://github.com/enonic/site-developer/blob/master/src/main/resources/site/parts/panel-items/panel-items.es)): all of the occurrences are always expanded by default. Especially annoying when you collapse some of them, then click "Apply" button and they all get expanded again making you scroll the panel to find where you were.

reisfmb:
@alansemenov 

I was initially looking at the code and couldn't find any problems. Then I started to mess with the content in the part itself and I think I've figured it out... but just to make sure, is there another example of a part in which the problem can be tested? 

The panel items part require all fields (minimum="1") except for the svg icon. If all fields are correctly filled, then after apply or save we'll not get them expanded, which is the correct behavior:

https://user-images.githubusercontent.com/67838246/172440565-8b7d6aa8-ca58-425c-b284-f2af57047ab3.mp4

I suspect that when you were testing, you didn't fill all the required fields for the items, and that's why they were constantly expanded. The problem is that if that's the case, i.e not filling all the required fields, we're not getting any kind of warning as we'd get in the [content type you initially mentioned](https://github.com/enonic/app-features/blob/master/src/main/resources/site/content-types/minimum-occurrence-form-item-set/minimum-occurrence-form-item-set.xml) if we changed some fields to be required:

![image](https://user-images.githubusercontent.com/67838246/172434632-c754c06b-573e-4867-a269-643776619b83.png)

Should we implement some kind of warning / red borders for validation in part config? I tested other parts that not necessarily use item-set and I can see that if we don't fill the required fields we'll also not get any warnings, so maybe that's on purpose.

reisfmb:
@alansemenov 

[...]

PR #4662, opened 2022-06-08
Changed inspect panel refresh behavior (#4630)

reisfmb:


modules/lib/src/main/resources/assets/js/app/wizard/page/LiveEditPageProxy.ts
```diff
@@ -457,7 +465,7 @@ export class LiveEditPageProxy {
                 const event: ComponentLoadedEvent = new ComponentLoadedEvent(newComponentView, componentView);
                 event.fire(this.liveEditWindow);
 
-                newComponentView.select();
+                newComponentView.select(clickPosition, menuPosition, newlyCreated, rightClicked, avoidInspectComponentRefresh);
```
alansemenov, line 468: Too many parameters. Find a way to use `ItemViewSelectedEventConfig` here, and send in only config parameters that are needed.

modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -635,7 +634,7 @@ export class LiveFormPanel
         }
     }
 
-    saveAndReloadOnlyComponent(componentView: ComponentView<Component>) {
+    saveAndReloadOnlyComponent(componentView: ComponentView<Component>, avoidInspectComponentRefresh?: boolean) {
```
alansemenov, line 637: For boolean properties it's better to use default param value rather than sending in `undefined`.

`saveAndReloadOnlyComponent(componentView: ComponentView<Component>, avoidInspectComponentRefresh: boolean = false)`

alansemenov, line 637: For boolean properties it's better to use default param value rather than sending in `undefined`.

`saveAndReloadOnlyComponent(componentView: ComponentView<Component>, avoidInspectComponentRefresh: boolean = false)`

alansemenov (changes requested):

alansemenov (approved):

Issue #4667, opened 2022-06-13
Migrate `Who's editing` to JS API 


anatol-sialitski:
Migrate implementation from Java API of SharedMap to JS API using `lib-grid`

PR #4668, opened 2022-06-13
Migrate `Who's editing` to JS API #4667

anatol-sialitski:


alansemenov (changes requested):
1) Code in `exports.join` and `exports.leave` is almost 100% identical and should be refactored
2) Make sure that all immutable vars are defined with `const`, not `let`.

ashklianko:
`func: function...` in join and leave methods to be extracted to a separate method

anatol-sialitski:
At first glance the methods are very similar, but if you look in more detail, you will see that this is not so. I can add more methods, but it will not be useful.

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+function CollaboratorsModifier(params) {
+    this.collaboratorKey = `${params.sessionId}=${params.userKey}`;
+    this.contentId = params.contentId;
+    this.collaborators = [];
+
+    this.removeExpired = function (timeAt) {
+        const self = this;
+        const originalSize = this.collaborators.length;
+        this.collaborators = this.collaborators.filter(collaborator => {
+            let lastJoin = self.extractLastJoin(collaborator);
```
ashklianko: const

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+            return timeAt - lastJoin < TTL_SECONDS;
+        });
+        return originalSize > this.collaborators.length;
+    };
+
+    this.removeCurrent = function () {
+        const originalSize = this.collaborators.length;
+        const collaboratorKey = this.collaboratorKey;
```
ashklianko, line 34: Why do we need to assign const `collaboratorKey = this.collaboratorKey;`? you ca use this.collaboratorKey

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+        this.collaborators = this.collaborators.filter(collaborator => !collaborator.startsWith(collaboratorKey));
+        return originalSize > this.collaborators.length;
+    };
+
+    this.getUniqueCollaborators = function () {
+        let uniqueCollaborators = new Set();
```
ashklianko: const

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+        this.collaborators.forEach(collaborator => uniqueCollaborators.add(collaborator));
+
+        let collaborators = [];
```
ashklianko: const

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+            distributed: true,
+            data: {
+                contentId: contentId,
+                collaborators: userKeys
+            }
+        });
+    };
+
+    this.extractLastJoin = function (collaboratorId) {
+        let idParts = collaboratorId.split("=", -1);
```
ashklianko: const

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+
+    this.getStoredCollaborators = function (storedCollaborators, isJoin) {
+        if (typeof storedCollaborators === 'undefined' || storedCollaborators === null) {
+            return isJoin ? [] : null;
+        }
+        return storedCollaborators;
+    };
+}
+
+CollaboratorsModifier.prototype.modify = function (oldCollaborators, isJoin) {
```
ashklianko: why not to keep it as the other functions?

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+        const self = this;
+        const originalSize = this.collaborators.length;
+        this.collaborators = this.collaborators.filter(collaborator => {
+            let lastJoin = self.extractLastJoin(collaborator);
+            return timeAt - lastJoin < TTL_SECONDS;
+        });
+        return originalSize > this.collaborators.length;
+    };
+
+    this.removeCurrent = function () {
```
ashklianko, line 32: this function is confusing - it not only removes current collaborator, but also returns something, this should not be like this. Split it into 2 methods: `removeCurrent()` should just filter out current user, `isCurrentUserRemoved()` should be used further in the code and it must use `Array.some()` to determine if user is there or not
However in some cases this is appropriate, so you may leave it as it is
Same issue with `removeExpired()`

modules/app/src/main/resources/lib/collaboration.js
```diff
@@ -1,33 +1,133 @@
...
+            return isJoin ? [] : null;
+        }
+        return storedCollaborators;
+    };
+}
+
+CollaboratorsModifier.prototype.modify = function (oldCollaborators, isJoin) {
+    this.collaborators = this.getStoredCollaborators(oldCollaborators, isJoin);
+
+    if (isJoin === false && this.collaborators === null) {
```
ashklianko: Why not use `!isJoin`? isJoin is sent explicitly so it won't have undefined value. 
you can write like this:
```
if (!isJoin && !this.collaborators) {
            return null;
        } 
```

ashklianko (changes requested):

modules/app/src/main/resources/lib/collaboration.js
anatol-sialitski: In the filter's callback function `this` will be pointed to a different context, where property `collaboratorKey` will be absent

modules/app/src/main/resources/lib/collaboration.js
anatol-sialitski: Other functions are private for `CollaboratorsModifier`. This function is public.

modules/app/src/main/resources/lib/collaboration.js
anatol-sialitski: I will leave it as is. Typically remove method returns true/false or previous (removed) value.

modules/app/src/main/resources/lib/collaboration.js
ashklianko: It doesn't need to be public to be accessible from doJoinOrLeave

modules/app/src/main/resources/lib/collaboration.js
anatol-sialitski: done

modules/app/src/main/resources/lib/collaboration.js
ashklianko: You are using an arrow function, so this will be taken from outer function context

modules/app/src/main/resources/lib/collaboration.js
ashklianko: You can put modify directly into the CollaboratorsModifier

ashklianko (changes requested):

ashklianko (approved):

modules/app/src/main/resources/services/collaboration/collaboration.js
```diff
@@ -47,13 +47,15 @@ exports.webSocketEvent = function (event) {
...
-        });
-        log.debug(`List of collaborators ${JSON.stringify(collaborators)} for content with id = "${event.data.contentId}"`);
-        break
+        if (event && event.data) {
+            let collaborators = collaborationLib.leave({
+                contentId: event.data.contentId,
+                sessionId: event.session.id,
+                userKey: event.session.user.key
+            });
+            log.debug(`List of collaborators ${JSON.stringify(collaborators)} for content with id = "${event.data.contentId}"`);
```
alansemenov: Let's remove logging unless there's some specific need for that.

alansemenov (changes requested):

alansemenov (approved):

modules/lib/src/main/resources/assets/js/app/event/CollaborationServerEvent.ts
```diff
@@ -7,7 +7,7 @@ import {ClassHelper} from '@enonic/lib-admin-ui/ClassHelper';
 export class CollaborationServerEvent
     extends Event {
 
-    public static EVENT_NAME: string = 'edit.content.collaborators.update';
+    public static EVENT_NAME: string = 'custom.edit.content.collaborators.update';
```
alansemenov, line 10: Why is it '`custom.edit.content.collaborators.update`' here, but '`edit.content.collaborators.update`' (without `custom.` prefix) in [collaboration.js](https://github.com/enonic/app-contentstudio/pull/4668/files#diff-f19f59e038f6a56f1445ee348ab987e34fc1d24cd9e4b3f7132fa02303fc965bR56)?

PR #4683, opened 2022-06-14
getVersionsForView is missing a version created after sorting changes

alansemenov:


ashklianko:
Moved to https://github.com/enonic/app-contentstudio/pull/4840

Issue #4664, opened 2022-06-09
Alphabetical order of custom widget icons


alansemenov:
It seems like only the "main" icon and "Settings" have fixed positions, the rest is random. Fix it so that "Archive" always goes second (if the CS+ app is deployed, of course) and the rest of custom widget icons are in the alphabetical order, sorted by widget's display name.

PR #4684, opened 2022-06-14
Alphabetical order of custom widget icons #4664

ashklianko:
-Moved toolbar classes to the separate files
-Sorting non default apps alphabetically


modules/lib/src/main/resources/assets/js/app/widget/WidgetsToolbar.ts
```diff
@@ -0,0 +1,160 @@
...
+                return false;
+            }
+
+            if (this.isSettingsApp(existingButton)) { // inserting before Settings button
+                pos = index;
+                return true;
+            }
+
+            // finding
+            if (newButton.getWidgetDisplayName().toLowerCase().localeCompare(existingButton.getWidgetDisplayName().toLowerCase()) < 0) {
```
alansemenov: `newButton.getWidgetDisplayName().toLowerCase()` should be called once before `.some()`. Actually, we don't even need entire `newButton` in this function, we only need its display name.

alansemenov (changes requested):

Issue #4572, opened 2022-05-18
Published/Unpublished statuses in the Version History should not show user name


llabaj-gj:
It seems that "Published by" status shows name of the user who marked content as "Ready" not the person who actually published it afterwards. See screenshot below, where it is conflicting with what is shown in version history widget.

The issues is in Content Studio 4.0.4, could not check if it also exists in 4.1.0

![image](https://user-images.githubusercontent.com/81151513/169005956-6418a00d-aa88-443b-bb53-9a33709a099b.png)

sigdestad:
In order to make room for new stuff in header. 
I propose we simply remove the "by user" and only keep the state e.g. "Published" here

alansemenov:
Places to remove "`by UserName`" after Published & Unpublished:
* Version History widget
* Content preview panel
* Content Wizard toolbar

PR #4686, opened 2022-06-14
Published/Unpublished statuses in the Version History should not show…

ashklianko:
… user #4572

Places where "by UserName" after Published & Unpublished was removed:

Version History widget
Content preview panel
Content Wizard toolbar


sigdestad:
Hmm.. What is this task? We should'nt remove this info from version history?

sigdestad:
Idea was to remove "by user" for all statuses, not limited to publish/unpublish

alansemenov:
> Hmm.. What is this task? We should'nt remove this info from version history?

Yes, we should. Showing "`Published by Elton John`" in the Version history (when it's not necessary "Elton John" who published it) is just as confusing in the widget as it is in the toolbar, there's no difference.

> Idea was to remove "by user" for all statuses, not limited to publish/unpublish

Well, not originally. For other statuses we correctly show the user who last made the changes. For example, for "`Modified by <user>`" everything is fine. What's confusing is a case like "`New by <user>`" when `<user>` is not necessarily the one who created the content. Ideally we should've done something like "`Published. Last modified by Elton John`". But for now we can just remove user from the status completely. People can use version history for the detailed changes.

Issue #4685, opened 2022-06-14
Upgrade CKEditor to version 4.19.0


alansemenov:
https://ckeditor.com/cke4/release/CKEditor-4.19.0

PR #4690, opened 2022-06-15
Upgrade CKEditor to version 4.19.0 #4685

ashklianko:


alansemenov (changes requested):
We don't need the `/samples/` folder.

ashklianko:
Removed

Issue #4702, opened 2022-06-17
HTML is compressed in HTML Editor's Source mode


alansemenov:
In Source mode HTML is now compressed. It didn't use to be like that before 4.1.0, so it's most likely caused by CKE upgrade.

4.1.0:
![image](https://user-images.githubusercontent.com/11612490/174298899-9d46d265-881e-404d-ae64-513e963d90f9.png)

3.3.4:
![image](https://user-images.githubusercontent.com/11612490/174298526-201d136b-129d-4efc-839f-989fc9ef6d7a.png)

PR #4730, opened 2022-06-21
HTML is compressed in HTML Editor's Source mode #4702

ashklianko:
-After updating to 4.18 we've missed adding a tiny but important plugin that helps to process formatting in source dialog


alansemenov (approved):
Nice!

Issue #4692, opened 2022-06-15
Enable appending query params & anchor to content URL


alansemenov:
When a non-media content is selected on the Content tab, we should show additional controls for appending anchor and/or query parameters to the URL.

1) Anchor
User can add only one anchor. Clicking the Add icon for Anchor will replace the button with a text input field and the X icon to the right of it. The field is mandatory. If user changes his mind, he will click the X icon and this will remove the input/icon and put back the Add button.

2) Query parameters
User can add unlimited number of query parameters. Clicking the Add button will add a new pair of Name/Value inputs and move the Add button down to add another pair etc. Each of the Name/Value fields are mandatory and will also have X icon to the right on each line to delete the pair of inputs.

![image](https://user-images.githubusercontent.com/11612490/173911062-ec2126d5-ead1-451d-9d2c-70880960f318.png)

PR #4734, opened 2022-06-22
Added anchor and query params to content tab(#4692)

reisfmb:


modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -352,6 +329,88 @@ export class LinkModalDialog
         return this.createFormItem(formItemBuilder);
     }
 
+    private createAnchorOption(id: string, label: string): FormItem {
+        const addButton: Button = new Button('Add');
```
alansemenov: Use the existing i18n resource.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -352,6 +329,88 @@ export class LinkModalDialog
...
+
+        const hideAnchorFormButton = new Button();
+        hideAnchorFormButton.addClass('remove-button transparent icon-close');
```
alansemenov: Extract to a separate method called `createRemoveButton` or something. You are using exactly the same button for query string params, no point in repeating the same code.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -352,6 +329,88 @@ export class LinkModalDialog
...
+    }
+
+    private createParamsOptions(id: string, label: string): FormItem {
+        this.paramsFormIds = [];
+
+        const addButton: Button = new Button('Add');
+
+        this.paramsFormItem = this.createFormItem(new ModalDialogFormItemBuilder(id, label)
+            .setValidator(this.validationQueryParams.bind(this)));
+        this.paramsFormItem.removeChild(this.paramsFormItem.getInput().getParentElement());
```
alansemenov, line 441: What does this line do and why?

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -471,6 +530,61 @@ export class LinkModalDialog
...
+            if (key && value) {
+                keyValueMap[decode(key)] = decode(value);
+            }
+        });
+
+        return keyValueMap;
+    }
+
+    private createKeyValueFormItems(initialKey: string = '', initialValue: string = ''): void {
+        const n = this.paramsFormIds.length
```
alansemenov: This should have a more explanatory name, not just `n`

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -64,13 +65,17 @@ export class LinkModalDialog
...
+    private anchorFormItem: FormItem;
+    private paramsFormItem: FormItem;
     private protocolsDropdownButton: MenuButton;
 
     private contentId: ContentId;
     private parentSitePath: string;
 
     private tabNames: any;
 
+    private paramsFormIds: {keyId: string, valueId: string}[];
```
alansemenov, line 78: Use [Map](https://howtodoinjava.com/typescript/maps/) instead. This will allow you use `.set`, `.delete` etc.

modules/lib/src/main/resources/i18n/dialogs.properties
```diff
@@ -146,6 +146,9 @@ dialog.link.radio.options.open=Open file
 dialog.link.radio.options.download=Download file
 dialog.link.radio.options.link=Link to page (treat media as regular content)
 dialog.link.urlprotocols.relative=Relative
+dialog.link.anchor=Anchor
+dialog.link.parameters=Parameters
+dialog.link.queryparams.empty=No key / value occurrence can be empty
```
alansemenov: Change to "`All parameters must have a name and a value`"

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -471,6 +530,61 @@ export class LinkModalDialog
...
+        return keyValueMap;
+    }
+
+    private createKeyValueFormItems(initialKey: string = '', initialValue: string = ''): void {
+        const n = this.paramsFormIds.length
+            ? parseFloat(this.paramsFormIds[this.paramsFormIds.length - 1].keyId.split('-')[1]) + 1
+            : 0;
+
+        const keyFormItemId = `paramsKey-${n}`;
+        const keyFormItem = this.createFormItemWithPostponedValue(keyFormItemId, '', () => initialKey, null, 'Key');
```
alansemenov: Rename '`Key`' to '`Name`', and it should come from i18n.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -471,6 +530,61 @@ export class LinkModalDialog
...
+
+        const valueFormItemId = `paramsValue-${n}`;
+        const valueFormItem = this.createFormItemWithPostponedValue(valueFormItemId, '', () => initialValue, null, 'Value');
```
alansemenov: "`Value`" should come from i18n.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -299,30 +291,15 @@ export class LinkModalDialog
...
-        });
-
-        if (this.getAnchor()) {
-            dropDown.setValue(this.getAnchor());
-        }
-
-        const formItemBuilder = new ModalDialogFormItemBuilder('anchor', i18n('dialog.link.tabname.anchor')).setValidator(
-            Validators.required).setInputEl(dropDown);
-
-        return this.createFormItem(formItemBuilder);
```
alansemenov: Can't remove this. We have a "hidden" feature for adding links to anchors on the same page. Can be tested as follows:
1. Use Anchor button (the one with the flag) in the editor toolbar to insert an anchor into the HTML text
2. Open the Link dialog.
3. You will see the "Anchor" tab and you can select your anchor from the dropdown to link to it.

image

image

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
reisfmb: I don't see why Map would benefit since I'm using some array methods on it, such as reduce to create the query parameters string.

modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
reisfmb: _**What does it do?**_
It removes the default `textInput` created by the `createFormItem` method since a `formItem` expects a `formItemBuilder` which requires an input in its constructor.

_**Why?**_
Because I want `paramsFormItem` to be a `formItem` without an input, since I only want to use it as a wrapper for all name / value pairs, to handle the validation of all pairs with a single validator.

alansemenov (changes requested):
We have changed syntax of the expected link format in the [backend task](https://github.com/enonic/xp/issues/7571), so we have to make according changes in this one as well.

If there are query string params, they should be concatenated (separated with `&`), then Url Encoded, then added to the resulting link as query parameter:
`content://<id>?query=<url_encoded_value_of_the_concatenated_query_string>`

If there's anchor, it should be Url Encoded, then added to the resulting link as fragment parameter:
`content://<id>?fragment=<url_encoded_value_of_the_anchor_field>`

If there are both:
`content://<id>?query=<query_string>&fragment=<anchor>`

alansemenov (approved):

Issue #4578, opened 2022-05-20
Layer Wizard - long user name should be ellipsis-wrapped


sgauruseu:
1. Create an user , name should be more than 35 chars
2. Open new layer wizard, and select the user in `Roles` form

**Bug** - broken layout in Roles form

![image](https://user-images.githubusercontent.com/3728712/169492184-dcc87197-255e-43a1-8663-9f8b8fba9425.png)

PR #4752, opened 2022-06-27
Fixed broken layout in roles form (#4578)

reisfmb:


modules/lib/src/main/resources/assets/styles/project/project-access-control.less
```diff
@@ -39,6 +39,7 @@
       float: left;
       margin: 0 10px 0 0;
       min-width: 200px;
+      width: 70%;
```
ashklianko: I would try to avoid using width for it, and update `.access-selector` to have a `flex-shrink: 0`

ashklianko (changes requested):

ashklianko:
Also when menu is expanded styling within looks weird:
image
Selection with a blue background should spread to entire width of a menu

ashklianko:
Good job, Bruno! 
Could you please also check why menu is not collapsing when opening other menu? It looks like this:
image

ashklianko (changes requested):
Need to check why menus not collapsing

reisfmb:
@ashklianko 
https://user-images.githubusercontent.com/67838246/176555820-164e5639-4c8a-4034-9144-e4d1717b221c.mp4

modules/lib/src/main/resources/assets/js/app/settings/wizard/panel/form/element/ProjectAccessSelector.ts
```diff
@@ -27,6 +27,10 @@ export class ProjectAccessSelector
 
     }
 
+    hideMenu(): void {
```
ashklianko: That is redundant if you don't perform any extra actions

modules/lib/src/main/resources/assets/js/app/settings/wizard/panel/form/element/ProjectAccessControlEntryView.ts
```diff
@@ -2,17 +2,25 @@ import {ProjectAccessControlEntry} from '../../../../access/ProjectAccessControl
...
+import {PrincipalContainerSelectedOptionView} from '@enonic/lib-admin-ui/ui/security/PrincipalContainerSelectedOptionView';
 import {ProjectAccessValueChangedEvent} from '../../../../event/ProjectAccessValueChangedEvent';
+import {ProjectACESelectedOptionsView} from './ProjectAccessControlComboBox';
+
+type projectACESelectedOptionsViewChild = PrincipalContainerSelectedOptionView<ProjectAccessControlEntry>;
 
 export class ProjectAccessControlEntryView
     extends PrincipalContainerSelectedEntryView<ProjectAccessControlEntry> {
 
+    private projectACESelectedOptionsView: ProjectACESelectedOptionsView;
```
ashklianko: I think this is bad idea to pass and keep a reference to it's parent: in general class shouldn't know nothing about it's structure and it's parents

modules/lib/src/main/resources/assets/js/app/settings/wizard/panel/form/element/ProjectAccessControlEntryView.ts
```diff
@@ -2,17 +2,25 @@ import {ProjectAccessControlEntry} from '../../../../access/ProjectAccessControl
...
+
     private accessSelector: ProjectAccessSelector;
 
     public static debug: boolean = false;
 
-    constructor(ace: ProjectAccessControlEntry, readonly: boolean = false) {
+    constructor(projectACESelectedOptionsView: ProjectACESelectedOptionsView, ace: ProjectAccessControlEntry, readonly: boolean = false, ) {
```
ashklianko: Same comment as above

modules/lib/src/main/resources/assets/js/app/settings/wizard/panel/form/element/ProjectAccessControlEntryView.ts
```diff
@@ -51,10 +63,23 @@ export class ProjectAccessControlEntryView
...
             });
+            this.accessSelector.onClicked(() => this.closeOtherAccessSelectors());
             this.appendChild(this.accessSelector);
         }
         this.accessSelector.setValue(this.item.getAccess(), true);
 
         this.appendRemoveButton();
     }
+
+    private closeOtherAccessSelectors(): void {
```
ashklianko: I think the better solution would be to check `TabMenu.hideMenuOnOutsideClick` method

ashklianko (changes requested):

reisfmb:
@ashklianko 

menus not collapsing is fixed from this: https://github.com/enonic/lib-admin-ui/pull/2660

ashklianko (approved):

Issue #4728, opened 2022-06-21
CKEditor: inject cke bundle on  demand


ashklianko:
CKE script is used to be put into the main.html `<head>` and executed immediately after it is loaded, however at that moment it is missing `CKEDITOR_BASEPATH` global variable that was used to be also set there before loading cke script. After adding [CSP ](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) we had to remove that global var from head and implement a workaround with wrapping cke source script with global function that might be called on demand. Thus in main.ts we set `CKEDITOR_BASEPATH` and call that wrapper function that is responsible for a cke initialization.

-Implement fix to load CKE in a deferred manner without wrapping it
-Cleanup build's assets/lib from redundant files and folders and make it compile in a single bundle file if possible

PR #4773, opened 2022-06-29
CKEditor: inject cke bundle on demand #4728

ashklianko:


modules/app/src/main/resources/admin/tools/main/main.html
```diff
@@ -33,7 +33,8 @@
     <link rel="stylesheet" type="text/css" href="{{assetsUri}}/admin/common/styles/lib.css">
     <link rel="stylesheet" type="text/css" href="{{assetsUri}}/styles/contentlib.css">
     <script type="text/javascript" src="{{assetsUri}}/admin/common/js/lib.js"></script>
-    <script type="text/javascript" src="{{assetsUri}}/lib/vendors.js"></script>
+    <script type="text/javascript" src="{{assetsUri}}/lib/ckepath.js" data-asset-url="{{assetsUri}}"></script>
+    <script type="text/javascript" src="{{assetsUri}}/lib/ckeditor/ckeditor.js"></script>
```
alansemenov, line 37: Can't we import './ckeditor/ckeditor' inside `ckepath.js`, let Webpack bundle them together and avoid having to add two references to cke assets in all of our html templates?

ashklianko: Will check

ashklianko: The idea in general is that CKEDITOR_BASEPATH to be set before invoking ckeditor.js that requires this var to be already set,  however we can't put import of ckeditor.js on the first line in the ckepath.js - that will be the same problem when this required var won't be set when ckeditor.js will be executed. So we need to keep it in a separate modules and make ckepath.js always come before ckeditor.js to guarantee CKEDITOR_BASEPATH is set

Issue #4575, opened 2022-05-19
User Access widget - extra user count is not updated after expanding the widget


sgauruseu:
1. Select a folder and open Details Panel.
2. Drag the splitter and expand User Access widget.

BUG - the extra user count should be updated after expanding the widget.
Hide then show details panel , the number gets correct

![image](https://user-images.githubusercontent.com/3728712/169224489-d3dee0c7-e5da-47d8-b712-0a85f6dcc25b.png)

https://user-images.githubusercontent.com/3728712/169224137-87447c96-a75f-4ffc-bdb3-e18b6d78aa0c.mp4

PR #4789, opened 2022-07-01
Fixed user count update on widget resize (#4575)

reisfmb:


ashklianko (changes requested):
If we use ResizeObserver then we don't need `DockedContextPanelResizeEvent` and any changes in the DockedContextPanel at all - just put a resize observer (that will trigger resizeListener) inside of the  `doLayout` in the `UserAccessListItemView`

PR #4792, opened 2022-07-04
Fixed user count update on widget resize

alansemenov:


modules/lib/src/main/resources/assets/js/app/security/UserAccessListItemView.ts
```diff
@@ -44,7 +44,9 @@ export class UserAccessListItemView
 
             this.resizeListener = this.setExtraCount.bind(this);
             ResponsiveManager.onAvailableSizeChanged(this, this.resizeListener);
```
ashklianko: Seems we also don't need this and overridden `remove()` method :)
Responsive manager is also adding some classes describing size (in addition to reacting to size change), but they are not used in styling, thus we can remove it

ashklianko (changes requested):

reisfmb:
Closed and reopened PR since build step was stuck:

build: _Expected — Waiting for status to be reported_

ashklianko (approved):

PR #4795, opened 2022-07-04
Edit page saving can reload the page and loose scrolling position

alansemenov:
-Fixed PageComponentsTreeGrid not to affect scrolling to the item views
-Updated TextComponentView to let set/get editor's cursor position
-Small update in HtmlEditor to remove redundant param
-Forcing scrolling to the loaded(saved) ItemView and scrolling to the last selected item after save
-Saving latest cursor position within TextComponent's editor before saving content and then restoring selection after save is complete

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/HtmlEditor.ts
@@ -624,7 +624,7 @@ export class HtmlEditor {
                 const config: any = {
                     editor: editor,
                     editorParams: this.editorParams,
-                    cursorPosition: this.getCursorPosition(editor)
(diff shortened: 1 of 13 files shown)


alansemenov (commented):
@ashklianko it works fine when you save with the Save button, but if you save with Cmd-S then text component will go out of edit mode (and of course lose its cursor position). can you fix that as well?

alansemenov:
Changes since the last version:
* Selected component is always scrolled into view (not only a new one)
* SelectedHighlighter is removed when a text component goes into edit mode (before, text component still had a blue mask over it even in edit mode)
* Show Apply button inside the Inspect panel for a Text component (to enable saving/reload of just that component), but only when it's in edit mode
* Keep cursor position also when using the Apply button

@ashklianko please check/review

ashklianko (approved):

PR #4797, opened 2022-07-04
CKEditor: inject cke bundle on demand

alansemenov:


alansemenov:
@ashklianko that's how you bundle the two files required for CKE together:

**lib/webpack.config.js**
```Javascript
    entry: {
...
        'lib/ckeditor': {
            import: ['./lib/ckepath.js', './lib/ckeditor/ckeditor.js']
        },
    },
```

Then the files will be bundled in the right order and you can use just one asset in HTML templates instead of two.

`<script type="text/javascript" src="{{assetsUri}}/lib/ckeditor.js" data-asset-url="{{assetsUri}}"></script>`

Another thing. We now _always_ load the CKE asset inside `main.html`. This is wrong, because we don't need it in the browse view. We should use `isBrowseMode` property to avoid loading the asset in browse mode:

```HTML
{{^isBrowseMode}}
  <script type="text/javascript" src="{{assetsUri}}/lib/ckeditor.js" data-asset-url="{{assetsUri}}"></script>
{{/isBrowseMode}}
```

Issue #4786, opened 2022-07-01
Filter requests: send params and don't construct NoQL search string on the frontend


ashklianko:
@vot-enonic noticed that when opening content's inbound references load mask does not disappear and filter search queries return 403 error

ashklianko:
![image](https://user-images.githubusercontent.com/4255706/176873807-5358f85d-3cf3-43ac-974a-33e03922f316.png)

PR #4806, opened 2022-07-06
Filter requests: send params and don't construct NoQL search string o…

ashklianko:
…n the frontend #4786

-Totally removed NoQL from filter panel request
-Using DSL on the backend to process filter panel requests


ashklianko:
NO MERGE FOR NOW: using PR to run automated checks

PR #4808, opened 2022-07-07
Filter requests: send params and don't construct NoQL search string o…

ashklianko:
…n the frontend #4786

-Totally removed NoQL from filter panel request
-Using DSL on the backend to process filter panel requests
-Some other classes still use queryExpr and will be fixed in other tasks
-Added tests


alansemenov (approved):
Looks fine. I'll leave review of the Java code to smart(er) people

anatol-sialitski (approved):

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentQueryJsonToContentQueryConverter.java
```diff
@@ -112,6 +119,30 @@ private void addQueryFilters( final ContentQuery.Builder builder )
...
+    }
+
+    private DslExpr createDslExpr()
+    {
+        return DslExpr.from( JsonToPropertyTreeTranslator.translate( JsonHelper.from( contentQueryJson.getQuery() ) ) );
+    }
+
+    private List<OrderExpr> createDslSortExpr()
+    {
+        return contentQueryJson.getQuerySort().stream()
```
vbradnitski, line 148: It fails with NPE if getQuery() result is not null and getQuerySort() is null

vbradnitski (changes requested):

vbradnitski (approved):

Issue #4836, opened 2022-07-13
Incorrect behaviour of Parameters validation in the Insert Link dialog


sgauruseu:
1. Open Insert Link modal dialog.
2. Select a content (not a media content)
3. Click on `Add parameters` button
4. Do not fill the param-name input but click on Insert button
5. Validation message gets visible - OK
6. Fill in the parameter name input

**BUG:**  Validation message remains visible after filling the name input
![image](https://user-images.githubusercontent.com/3728712/178738356-1162e699-8d18-4fda-bee8-3280187d2469.png)

Expected behaviour should  look like as in the requred long input, for example:

https://user-images.githubusercontent.com/3728712/178738801-a9b1c6a7-2fbf-4497-8b35-d697b6f89764.mp4

PR #4838, opened 2022-07-13
Fixed incorrect behavior of validator in parameters (#4836)

reisfmb:


modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -657,7 +657,13 @@ export class LinkModalDialog
 
         this.paramsFormIds.push({keyId: keyFormItemId, valueId: valueFormItemId});
 
-        keyFormItem.getInput().giveFocus();
+        const keyFormItemInput = <TextInput>keyFormItem.getInput();
+
+        keyFormItemInput.onValueChanged(() => {
+            if (this.validationQueryParams.bind(this)() === undefined) {
+                this.paramsFormItem.validate(new ValidationResult(), true);
+            }
```
alansemenov: It feels like a strange way to solve this... You are checking if validation error message is undefined, and in this case you are running validation. 

Maybe it's the architecture of these methods that is wrong originally. It should be something like this:

```Javascript
private isQueryParamsValid(): boolean {
  const isValid = Array.from(this.paramsFormItem.getHTMLElement().getElementsByTagName('input'))
  .filter((input: HTMLInputElement) => input.className.indexOf('params-key') >= 0)
  .every((input: HTMLInputElement) => input.value.trim() !== '');

  return isValid;
}

private validationQueryParams(): string {
  return !this.isQueryParamsValid() ? i18n('dialog.link.queryparams.empty') : undefined;
}
```

Then in your listener you could do this:
```Javascript
keyFormItemInput.onValueChanged(() => {
  if (!this.isQueryParamsValid()) {
    this.paramsFormItem.validate(new ValidationResult(), true);
  }
});
```

alansemenov (changes requested):

alansemenov:
Another thing: `createKeyValueFormItems` method is quickly getting out of hands and must be refactored.

We already have a separate `createRemoveButton` for the remove icon, the same should be done for `keyFormItem` and `valueFormItem`. 

Remember that event listeners should be added inside corresponding methods where you create the elements (or even in separate methods, if there are too many listeners). So there's no reason why `removeButton.onClicked` should be added inside `createKeyValueFormItems` and not inside `createRemoveButton` where it naturally belongs. Hint: if `createRemoveButton` is used in several places, create additional methods which call it and add their own listeners:

```Javascript
private createRemoveButtonForParams(): Button {
  const button: Button = this.createRemoveButton();
  button.onClicked(...);
  button.onShown(...);
}

private createRemoveButtonForSmthElse(): Button {
  const button: Button = this.createRemoveButton();
  button.onClicked(...);
  button.onSmthElse(...);
}
```

So the goal is to eventually have something like this in `createKeyValueFormItems`:

```Javascript
...
const paramKeyFormItem: FormItem = this.createParamKeyFormItem();
const paramValueFormItem: FormItem = this.createParamValueFormItem();
const removeButton: Button = this.createRemoveButtonForParams();

const divWrapper: DivEl = new DivEl('params-wrapper')
  .appendChild(paramKeyFormItem)
  .appendChild(paramValueFormItem)
  .appendChild(removeButton);
...
```

Finally, I would like you to rename `createKeyValueFormItems` to `createParamsFormItems`, as current name sounds too generic.

alansemenov (approved):
Good stuff

Issue #4837, opened 2022-07-13
Incorrect behaviour of Anchor validation in the Insert Link dialog


sgauruseu:
1. Open Insert Link modal dialog.
2. Select a content (not a media content)
3. Click on `Add anchor` button
4. fill in the anchor input.
5. Click on `Insert` button.
6. Reopen the  `Insert Link` modal dialog
7. Clear the Anchor input.
8. Click on `Insert` button.

**BUG**: Dialog should not close, validation message should be displayed in the modal dialog (Anchor input is required)

https://user-images.githubusercontent.com/3728712/178747235-acfcd336-038b-4efb-9ea1-180ff46fc5df.mp4

PR #4839, opened 2022-07-13
Fixed incorrect behavior of validator in anchor (#4837)

reisfmb:


sgauruseu:
https://user-images.githubusercontent.com/3728712/178975136-18592a4b-9630-4763-9cf6-367b7f2324ce.mp4

Red validation message(This input is required) should appear after I cleared (via Backspace) the anchor input

alansemenov (changes requested):
I've debugged a little and some things are plain wrong.

For example, in `createSelectorFormItem` you have an event listener for `onLoaded` which then fires `handleSelectorValueChanged`. This is wrong - loading data set for a selector is not the same as selecting a value in it. 

This triggers lots of incorrect actions, for example `handleSelectorValueChanged` hides a lot of elements when no value is selected, but since this method is called on data load, these elements get hidden on dropdown expand. Since you've now added validation on `anchorInput` hide/show, we get a chain of actions like this: content selector expanded -> anchor input gets hidden -> validation is called. This doesn't make any sense.

1) Use event handlers that are appropriate for an action. For option selected we have `onOptionSelected`, for option deselected we have `onOptionDeselected`.

```Javascript
contentSelector.onOptionSelected(() => 
  this.handleSelectorValueChanged(contentSelector.getSelectedContent(), formItem)
);
```

2) If you are using `onShown` or `onHidden` event handlers, make sure they are not called several times for the same element within the same stack trace. For example, inside `createFragmentOption` you have

  `anchorInput.onHidden(() => ...);`

and then on the next line 

`anchorInput.hide();`

In this case your `onHidden` handler will be called already after initial hide. The question to ask is - is this really needed? Do you really need to call everything inside `onHidden` upon initialisation of this element? If not, then maybe the event handler should be added _after_ `anchorInput.hide();`, and not before.

3) If there's no way to prevent calling `onShown` and `onHidden` several times within the same stack trace, maybe it makes sense to check inside the event handler that this element is _already_ shown or hidden.

```Javascript
anchorInput.onHidden(() => {
  if (!anchorInput.isVisible()) {
    return;
  }
  // do stuff
});
```

4) You don't need `this.anchorFormItem.setValidator(Validators.required)` inside `addButton.onClicked` if you added it in `anchorInput.onShown`.

alansemenov (approved):

Issue #4830, opened 2022-07-12
Versions Widget  - several active versions displayed after restoring content


sgauruseu:
1. Create a folder then archive it .
2. Go to Archive and restore the folder.
3. Select the folder and open Versions Widget
4. Expand several version items. 

**BUG**:  Only one item should be with  `Active version`  button

https://user-images.githubusercontent.com/3728712/178438133-ed4e4630-8b8b-474c-8b4a-d711e413a0af.mp4

PR #4844, opened 2022-07-14
Versions Widget - several active versions displayed after restoring c…

ashklianko:
…ontent #4830

-Fixed incorrect setting of active version
-Removed 'revertable' prop from VersionHistoryItem that doesn't belong there
-Updated styling for sorted/permissions items to be not interactable

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/context/widget/version/VersionHistoryListItem.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/context/widget/version/VersionHistoryListItem.ts
@@ -33,11 +33,11 @@ export class VersionHistoryListItem
     private createVersionViewer(): VersionHistoryListItemViewer {
         const versionViewer: VersionHistoryListItemViewer = new VersionHistoryListItemViewer();
 
-        if (this.version.isRevertable() || this.version.isActive()) {
+        if (this.isInteractableItem()) {
             this.addOnClickHandler(versionViewer);
         }
 
-        if (this.version.isRevertable()) {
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/app/view/context/widget/version/VersionHistoryListItem.ts
```diff
@@ -33,11 +33,11 @@ export class VersionHistoryListItem
     private createVersionViewer(): VersionHistoryListItemViewer {
         const versionViewer: VersionHistoryListItemViewer = new VersionHistoryListItemViewer();
 
-        if (this.version.isRevertable() || this.version.isActive()) {
+        if (this.isInteractableItem()) {
             this.addOnClickHandler(versionViewer);
         }
 
-        if (this.version.isRevertable()) {
+        if (this.isInteractableItem()) {
```
alansemenov: Any reason for not joining these two `if`s?

ashklianko: Missed it, will fix

alansemenov (changes requested):
It seems to work fine, but there's a new bug. 

After sorting is changed on an item selected in the grid, Version History widget doesn't get refreshed, so you don't see the new version created after the change of sorting. You have to unselect the row, then select it again to see the updates.

ashklianko:
Updated

alansemenov (approved):

Issue #4867, opened 2022-07-18
Update Compare Versions dialog with new changes


alansemenov:
The list of versions in the Compare Versions dialog should reflect the latest changes in the widget: new versions for sort and permissions changes, correct icons etc.

PR #4888, opened 2022-07-19
Update Compare Versions dialog with new changes #4867

ashklianko:


sgauruseu:
1. Compare Versions dialog - I can click on `Revert` button many times when a 'Permissions updated' item is selected in the dropdown 
2.  `Unknow` status appears in the widget and in browse panel after closing the modal dialog

https://user-images.githubusercontent.com/3728712/180006124-5a636e90-4d9c-45da-8293-9f6a1756dabe.mp4

sgauruseu:
image


1. **Version item with the button `Active version` is not present in the widget** 



image


2.  No changes to revert appears after clicking on Revert button

modules/lib/src/main/resources/i18n/phrases.properties
```diff
@@ -378,6 +378,7 @@ settings.projects.name.helptext=NB! This field cannot be changed after the proje
 settings.projects.notfound=No projects found
 settings.projects.nopermissions=Contact your administrator to set up project permissions!
 settings.dialog.new=Create New
+settings.dialog.project.wizard=Create Project
```
alansemenov: This doesn't belong here

modules/lib/src/main/resources/assets/js/app/view/context/widget/version/VersionHistoryItem.ts
```diff
@@ -201,4 +206,171 @@ export class VersionHistoryItem {
...
+
+        return versionAliasBuilder.build();
+    }
+
+    clone(): Builder {
+        return new Builder(this);
+    }
+}
+
+export class Builder {
```
alansemenov: Why do we need to export this class? If we do, then it shouldn't have such a generic name.

modules/lib/src/main/resources/assets/js/app/view/context/widget/version/VersionHistoryItem.ts
```diff
@@ -23,106 +24,114 @@ export interface CreateParams {
...
 
-        item.dateTime = publishInfo.getTimestamp();
-        item.user = publishInfo.getPublisherDisplayName() || publishInfo.getPublisher();
+    static fromPublishInfo(contentVersion: ContentVersion): Builder {
+        const builder: Builder = new Builder();
+        const publishInfo: ContentVersionPublishInfo = contentVersion.getPublishInfo();
+
+        builder.setVersion(contentVersion);
+        builder.setDateTime(publishInfo.getTimestamp());
+        builder.setUser(publishInfo.getPublisherDisplayName() || publishInfo.getPublisher());
```
alansemenov: We should chain builder's setters:
```Typescript
builder
  .setVersion()
  .setDateTime()
...
```

modules/lib/src/main/resources/assets/js/app/ContentVersions.ts
```diff
@@ -4,39 +4,31 @@ import {GetContentVersionsForViewResultsJson} from './resource/json/GetContentVe
...
 
-    constructor(contentVersions: ContentVersion[]) {
+    private readonly activeVersionId?: string;
+
+    constructor(contentVersions: ContentVersion[], activeVersionId?: string) {
         this.contentVersions = contentVersions;
+        this.activeVersionId = activeVersionId;
     }
 
     getContentVersions(): ContentVersion[] {
```
alansemenov: Let's rename this to `get`. Now we have `this.contentVersions.getContentVersions()` all over the place and it will be unambiguous (and look better) as `this.contentVersions.get()`

alansemenov (changes requested):

alansemenov (approved):

Issue #4510, opened 2022-04-27
Workflow icon should not be displayed in a wizard when the content is published


sgauruseu:
1. Green icon is not displayed in grid for published contend - OK
 
![image](https://user-images.githubusercontent.com/3728712/165544274-c39781e0-bf2d-4fc0-b509-803e61c0681f.png)

2. Open this published content.

**BUG** - Green icon is  displayed in the wizard

![image](https://user-images.githubusercontent.com/3728712/165544564-ca33264d-c883-4278-bead-8bbf832166fe.png)

PR #4907, opened 2022-07-26
Workflow icon should not be displayed in a wizard when the content is…

ashklianko:
… published #4510

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/WorkflowStateManager.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/WorkflowStateManager.ts
@@ -96,11 +96,11 @@ export class WorkflowStateManager {
 
     private toggleWorkflowStateClasses(element: Element) {
         element.toggleClass(WorkflowStateManager.INVALID_CLASS, this.status.invalid);
-        element.toggleClass(WorkflowStateManager.READY_CLASS, this.canMarkIconElementAsReady());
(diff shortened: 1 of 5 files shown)


modules/lib/src/main/resources/assets/js/app/wizard/WorkflowStateManager.ts
```diff
@@ -94,10 +94,22 @@ export class WorkflowStateIconsManager {
...
-        element.toggleClass(WorkflowStateIconsManager.INVALID_CLASS, status.invalid);
-        element.toggleClass(WorkflowStateIconsManager.READY_CLASS, status.ready);
-        element.toggleClass(WorkflowStateIconsManager.IN_PROGRESS_CLASS, status.inProgress);
+    private toggleWorkflowStateClasses(element: Element) {
+        element.toggleClass(WorkflowStateManager.INVALID_CLASS, this.status.invalid);
+        element.toggleClass(WorkflowStateManager.READY_CLASS, this.canMarkIconElementAsReady());
+        element.toggleClass(WorkflowStateManager.IN_PROGRESS_CLASS, this.status.inProgress);
+    }
+
+    private canMarkIconElementAsReady(): boolean {
```
alansemenov: I don't think this is a correct method name (hard to understand, too). Should be similar to `isPublished` and `isMoved` - `isReady` or `isInReadyWorkflowState` or `isMarkedAsReady`.

ashklianko: Done

alansemenov (approved):

Issue #4915, opened 2022-08-01
OptionSet wizard - Save button gets enabled after updating permissions


sgauruseu:
1. Select `Features`, open wizard for new `optionset` content
2. Fill in the name input, click on `Save` button
3. Open Edit Permissions dialog , uncheck `Inherit permissions` and remove several entries. click on Apply

**BUG** - `Save` button gets enabled aftfer applying new permissions

![image](https://user-images.githubusercontent.com/3728712/182115345-cb3a682c-6339-4f16-ae95-3173dfbc1086.png)

 **Issue 2**  -Click on button `Save` , expand the selector then click on `Option1`
Error apears in the console:

![image](https://user-images.githubusercontent.com/3728712/182115579-1a469acb-a14c-4319-8ad0-9be1d8453c80.png)

PR #4941, opened 2022-08-04
OptionSet wizard - Save button gets enabled after updating permission…

ashklianko:
…s #4915

-issue was that permissions update handler was not enriching persisted item data with empty properties that are set during forms layout; Thus comparing viewed and persisted data was saying that those unequal; Fixed by using update handler that will put fresh content data (received after permissions updated) through wizard forms and enrich it with missing empty props that makes viewed and persisted data equal

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -2281,10 +2281,6 @@ export class ContentWizardPanel
         return this.getPersistedItem().getDisplayName() !== this.getWizardHeader().getDisplayName();
     }
 
-    private isNameUpdated(): boolean {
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -2287,6 +2281,10 @@ export class ContentWizardPanel
         return this.getPersistedItem().getDisplayName() !== this.getWizardHeader().getDisplayName();
     }
 
+    private isNameUpdated(): boolean {
+        return this.getPersistedItem().getName().getValue() !== this.getWizardHeader().getName();
+    }
+
```
alansemenov: Not used?

alansemenov (changes requested):

alansemenov (approved):

Issue #4939, opened 2022-08-04
Dashes are not allowed in URLs in the Insert Link 


alansemenov:
This URL won't be validated in the Insert Link dialog: https://something-xyz.mysite.net/ApplicationForm/SinglePageApplicationForm.aspx?cid=139&departmentId=21678&ProjectId=183198

**A PR should be made only to the 4.2 branch, not to the master (since we have a brand new implementation of the Insert Link dialog there).**

PR #4944, opened 2022-08-04
Dashes are not allowed in URLs in the Insert Link #4939

ashklianko:


ashklianko:
Studio 4.2 depends on lib-admin-ui 4.3 that doesn't have latest changes in common url validator, thus dash is not passing url check. Copied actual validator to studio 4.2 specifically for LinkModalDialog

alansemenov (changes requested):
> Studio 4.2 depends on lib-admin-ui 4.3 that doesn't have latest changes in common url validator, thus dash is not passing url check. Copied actual validator to studio 4.2 specifically for LinkModalDialog

We will release a new version of lib-admin-ui (4.3.1) which CS 4.2.2 will depend on, so it's ok to make a separate issue/PR (targeting the 4.3 branch) for lib-admin-ui which adds the required validation.

ashklianko:
The we don't need this pull at all, issue will be fixed with [lib-admin-ui changes](https://github.com/enonic/lib-admin-ui/pull/2714)

Issue #4945, opened 2022-08-05
Version Widget  - version items are not loaded after opening the widget


ashklianko:
@sgauruseu commented on [Wed Aug 03 2022](https://github.com/enonic/app-contentstudio-plus/issues/632)

1. Do login then open `Archive` browse panel.
2. Select an exisiting folder then open versions widget.

**BUG**: version items are not loaded
![err_expand_version398394](https://user-images.githubusercontent.com/3728712/182599959-5ca7968f-89c8-4292-894f-f2e9b0124375.png)

Items appear after refreshing the page in browser,

PR #4946, opened 2022-08-05
Version Widget - version items are not loaded after opening the widge…

ashklianko:
…t #4945

-catching errors from promises
-added NPE check in ContextView.ts that was spawning error with loading history items in archive (and was not catched without catch clause)

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/context/ContextView.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/context/ContextView.ts
@@ -36,6 +36,7 @@ import {ReloadActiveWidgetEvent} from './ReloadActiveWidgetEvent';
 import {ContentId} from '../../content/ContentId';
 import {WidgetItemView} from './WidgetItemView';
 import {VersionContext} from './widget/version/VersionContext';
+import {DefaultErrorHandler} from '@enonic/lib-admin-ui/DefaultErrorHandler';
 
 export class ContextView
     extends DivEl {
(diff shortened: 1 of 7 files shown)


modules/lib/src/main/resources/assets/js/app/view/context/ContextView.ts
```diff
@@ -105,8 +106,8 @@ export class ContextView
         this.onRemoved(() => ApplicationEvent.un(handleApplicationEvents));
 
         VersionContext.onActiveVersionChanged((contentId: string, version: string) => {
-            if (this.item.getId() === contentId && this.isVisible() && this.activeWidget === this.versionsWidgetView) {
-                this.updateActiveWidget();
+            if (this.item?.getId() === contentId && this.isVisible() && this.activeWidget === this.versionsWidgetView) {
+                this.updateActiveWidget().catch(DefaultErrorHandler.handle);
```
alansemenov: Why not catch once inside `updateActiveWidget`'s return instead of repeating catch for every single call of `updateActiveWidget`?

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/view/context/ContextView.ts
ashklianko: Right, fixed

alansemenov (approved):

PR #4950, opened 2022-08-08
Project Wizard Dialog: add "Copy from parent" buttons

ashklianko:


ashklianko:
Will merge myself to use in next tasks

Issue #5034, opened 2022-08-29
Inspect panel should not auto-expand for a text component in edit mode


alansemenov:
When a Text component is in edit mode, any click inside the editor will currently expand the Inspect panel. If user collapses it and then clicks somewhere inside the editor, the panel will get expanded again. It should not.

Test use case 1:
1. Open a page
2. Single-click a text component
3. The Inspect panel opens.
4. Double-click the component to switch to the edit mode
5. The Inspect panel should close to give max available width to the editor.
6. Click anywhere inside the component's editor. The Inspect panel should remain closed.
7. Go out of the edit mode. The Inspect panel should remain closed until user selected some component on the page.

Test use case 2:
1. Open a page
2. Double-click a text component. 
3. Edit mode opens. The Inspect panel should not open because the text component was never selected (you went from "nothing selected" directly to the edit mode).

The rest is the same as 6-7 in the previous case.

Note, that if browser window is wide enough to allow for **docked** Inspect panel, then it should work as now (always be open by default until explicitly closed).

PR #5057, opened 2022-08-31
Inspect panel should not auto-expand for a text component in edit mod…

ashklianko:
…e #5034


ashklianko:
Ok, I see where the problem is, will need to fix original issue in the task different way

alansemenov (approved):

Issue #5025, opened 2022-08-25
Publishing dialog is not locked while items are being marked as ready


alansemenov:
When using bulk "Mark as Ready" function in the Publishing dialog, there's no locking of the dialog content and no visual indication that something is being changed. If there are many items to be changed and connection is relatively slow, you can easily execute "Mark as Ready" several times before you'll get notification from the server and items in the dialog will be updated.

So, once "Mark as Ready" is clicked, apply a mask with a spinner over the dialog and remove them once the notification from the server is received.

PR #5061, opened 2022-08-31
Publishing dialog is not locked while items are being marked as ready…

ashklianko:
… #5025


ashklianko:
Will ad more changes, including in lib-admin-ui, that will make tests fail until lib-admin-is merged

alansemenov (approved):

Issue #5073, opened 2022-09-02
Path field gets unlocked after a published content is modified


alansemenov:
1. Publish a content. Path field gets locked with a pencil icon next to it
2. Make some changes to the content.
3. Save.

Bug: The pencil icon disappears and the path field gets "unlocked". The content is still published, so it should still not be easy to change its path.

![image](https://user-images.githubusercontent.com/11612490/188159504-fecc3c92-d659-489d-8594-22db157715d6.png)

PR #5076, opened 2022-09-05
Path field gets unlocked after a published content is modified #5073

ashklianko:
-Locking path/name change for a content in a wizard when it is not New or Unpublished


alansemenov (approved):
Looking good!

Issue #5136, opened 2022-09-14
Rename main Content Studio's widget


alansemenov:
Should be called "Content", not "Content Studio".

![image](https://user-images.githubusercontent.com/11612490/190143481-67132ede-1f41-4336-b773-72dc4bec802f.png)

PR #5150, opened 2022-09-16
Rename main Content Studio's widget #5136

ashklianko:


alansemenov (changes requested):
We should use `app.admin.tool.main` instead of `app.name` for widget name in the header and the left-side menu. But check its usage in [AdminMapper](https://github.com/enonic/app-contentstudio/blob/master/modules/app/src/main/java/com/enonic/xp/app/contentstudio/AdminToolMapper.java#L28), it might be that it shouldn't be used there.

Long story short - `app.name` should be "`Content Studio`" and should not be localised, while `app.admin.tool.main` is the main widget name which should be changed to "`Content`" and localised in other languages (which means we need to delete it from other phrases files so that it gets translated).

Issue #4916, opened 2022-08-01
Display "Move" and "Permissions updated" in the Version History widget


alansemenov:
NB! Rollback of a move operation should not be allowed.

PR #5161, opened 2022-09-19
Correctly display "move" in the Version History widget #4916

ashklianko:
-Fetching newly added content version's field 'permissions' -Adding 'Permissions' entry to the version history -Considering all non-data changes that are not sort or permissions changes as a content move


alansemenov (changes requested):
* Change icon for move to the one below
* Change text for permission update from "_Permissions_" to "_Permissions updated_"
* Allow rollback of permission update (and test it)

image

ashklianko:
Also @sgauruseu found 2 issues, so will have to work on it more

PR #5171, opened 2022-09-20
Correctly display "move" in the Version History widget #4916

ashklianko:
-Fetching newly added content version's field 'permissions' -Adding 'Permissions' entry to the version history -Considering all non-data changes that are not sort or permissions changes as a content move


sgauruseu:
What version item should be added after reverting  permissions ?   `Permissions updated` or `Edited`  ?

alansemenov:
> What version item should be added after reverting permissions ? Permissions updated or Edited ?

_Edited_ should be correct for any reverted version.

ashklianko:
@sgauruseu I've added changes related to `inheritPermissions` property change, permissions and move operations, so please check them also.

alansemenov (approved):

PR #5177, opened 2022-09-21
Cache response of GetMimeTypesByContentTypeNamesRequest

alansemenov:


ashklianko (changes requested):
Approach won't work when multiple requests are being sent simultaneously before first response was returned, parsed and cached. Promises should be cached instead

Issue #5244, opened 2022-10-04
Version History: certain operations not shown after publishing


ashklianko:
1. Publish content
2. Change permissions/ rename
3. Publish again

Result: operation from step 2 is not shown in the versions list

PR #5262, opened 2022-10-10
Version History: certain operations not shown after publishing #5244

ashklianko:
-When publishing item second or more times then backend doesn't create a new version for publish, just uses existing one and adds publishInfo for it. So we need to create virtual extra version so it is to be displayed in a version history and make CompareContentVersionsDialog.ts to correctly display it

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/context/widget/version/ContentVersionsConverter.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/context/widget/version/ContentVersionsConverter.ts
@@ -1,5 +1,5 @@
 import {DateHelper} from '@enonic/lib-admin-ui/util/DateHelper';
-import {ContentVersion} from '../../../../ContentVersion';
+import {ContentVersion, ContentVersionBuilder} from '../../../../ContentVersion';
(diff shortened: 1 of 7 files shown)


modules/lib/src/main/resources/assets/js/app/view/context/widget/version/ContentVersionsConverter.ts
```diff
@@ -119,6 +156,21 @@ export class ContentVersionsConverter {
         return createParams;
     }
 
+    private getPreviousVersion(version: ContentVersion): ContentVersion {
+        let previousVersion: ContentVersion = null;
+
+        this.filteredVersions.some((v: ContentVersion, index: number) => {
+            if (version === v) {
+                previousVersion = this.filteredVersions[index + 1];
```
alansemenov, line 164: Did you test the case when `version === this.filteredVersions[this.filteredVersions.length-1]`? In this case `previousVersion = this.filteredVersions[index + 1];` will give `undefined` (or even an exception).

ashklianko: it will give undefined which is totally ok in this situation

alansemenov (approved):

PR #5277, opened 2022-10-11
Document New Project Wizard

alansemenov:


docs/storage.adoc
```diff
@@ -0,0 +1,35 @@
+= Content Storage
+:toc: right
+:imagesdir: storage/images
+
+Content in XP is stored in <<storage/projects#,Content Projects>>, each having its own underlying repository. In a clean new instance of XP
+there's only one project (called "Default") which cannot be deleted but <<config#disable_default_project,can be disabled>> using the
+configuration file.
+
+Users with privileged permissions can create their own Content Projects and localise content in these projects using <<storage/layers#,Content Layers>>, which essentially a special type of project inheriting content from an upstream project while enabling localisation into a specific language.
```
ashklianko: ..._which essentially **IS** a special type_...

ashklianko (approved):

docs/storage.adoc
ashklianko:
```suggestion
Users with privileged permissions can create their own Content Projects and localise content in these projects using <<storage/layers#,Content Layers>>, which essentially is a special type of project inheriting content from an upstream project while enabling localisation into a specific language.
```

Issue #5279, opened 2022-10-12
Incorrect logic of version comparison


alansemenov:
* Hide "Revert" button for "Permissions Updated" and "Moved" versions. Make sure that this is done not only in the widget, but in the Version Comparison dialog as well.
* "Compare" icons should only be displayed for versions that can be compared to a previous one, i.e. the icon should never be shown for the very first version. After the changes the active version should have the comparison icon (now it doesn't), unless it's also the very first one.
* Change the logic of preselecting versions in the Compare Versions dialog, so that in the right dropdown we preselect the version that was clicked in the widget, and in the left dropdown it's the "Previous version" alias.
* Add seconds for versions inside dropdowns in the dialog

PR #5295, opened 2022-10-14
Incorrect logic of version comparison #5279

ashklianko:


ashklianko:
Will update with latest master

alansemenov (approved):
The new behaviour seems to work fine, great job!

Issue #2336, opened 2020-10-01
Broken layout in ACE selectors on long principal names


sgauruseu:
- Create an user with name that contains more than 50 chars.
- Add Contributor role to this user
-  Add this user  in permissions dialog

Case 1: 
Project Wizard

![image](https://user-images.githubusercontent.com/3728712/94788534-5f361380-03dc-11eb-9e18-47a32e5ca3bb.png)

Case 2: 
Edit permissions dialog 

![image](https://user-images.githubusercontent.com/3728712/94788603-796ff180-03dc-11eb-8669-b6b7b10838d5.png)

PR #5329, opened 2022-10-20
Broken layout in ACE selectors when principal name contains more than…

ashklianko:
… 50 symbols #2336


modules/lib/src/main/resources/assets/js/app/view/AccessControlEntryView.ts
```diff
@@ -20,7 +20,8 @@ export class AccessControlEntryView
         super(ace, readonly);
     }
 
-    doLayout(object: Principal) {
+    doLayout(object: Principal): void {
+        console.log('doLayout');
```
alansemenov, line 24: Forgot to remove `console.log`

alansemenov (changes requested):

Issue #5102, opened 2022-09-07
Modal dialog hangs after creating new request publishing with empty Assignees


sgauruseu:
This bug only reproduces on the next request creation, the previous request was created with a selected option in Assignees selector.

1. Setup the `mail.cfg`
```
smtpHost=smtp.gmail.com
smtpPort=587
smtpAuth=true
smtpUser=userNmae@enonic.com
smtpPassword=password
smtpTLS=true
```
2. Create new user with an actual email
3. Do login with the user
4. Select a content and open Request Publishing dialog
5. Fill in the Describe input
6. Select the user in Assignees selector then click on Create Requies button - new requiest is created - OK
7. Open Request Publish dialog again and fill in `Describe` input
8. Click on Create button 

**BUG - Modal dialog hangs after clicking on `Create Request` button** 

![image](https://user-images.githubusercontent.com/3728712/188918820-006ed28f-1590-442c-912e-2cb0f970c6d6.png)

```
 - Message sending failed
com.enonic.xp.mail.MailException: Empty address
        at com.enonic.xp.mail.impl.MailServiceImpl.handleException(MailServiceImpl.java:72)
        at com.enonic.xp.mail.impl.MailServiceImpl.send(MailServiceImpl.java:61)
        at com.enonic.xp.app.contentstudio.rest.resource.issue.IssueNotificationsSenderImpl.lambda$notifyIssueCreated$0(
```

PR #5334, opened 2022-10-21
Modal dialog hangs after creating new request publishing with empty A…

ashklianko:
…ssignees #5102


ashklianko:
Tests will run ok, however issue won't be fixed until lib-admin-ui pull is built

ashklianko:
Removed dependency from lib-admin task, now can be fixed after succesfull tests run since fix is directly in the studio branch

alansemenov (approved):

Issue #5413, opened 2022-11-14
Open Publishing wizard after content is marked as ready


alansemenov:
After "Mark is ready" action is executed (in Content Grid or Content Wizard), automatically open the Publishing Wizard for the updated items. Make sure the notification about updated items still pops up.

PR #5430, opened 2022-11-15
Open Publishing wizard after content is marked as ready #5413

edloidas:


ashklianko (changes requested):
@edloidas please don't forget to set variables' types

edloidas:
Okay, I added more explicit types. TBH, I find this approach as redundant nowadays, but prettifying and restructuring code will be out of the scope of this task.

modules/lib/src/main/resources/assets/js/app/browse/action/MarkAsReadyContentAction.ts
```diff
@@ -7,48 +7,62 @@ import {showFeedback} from '@enonic/lib-admin-ui/notify/MessageBus';
...
+                new ContentPublishPromptEvent({model: contentToPublish}).fire();
             }
         }).catch(DefaultErrorHandler.handle);
     }
 
     isToBeEnabled(state: ContentTreeGridItemsState): boolean {
+        this.canPublish = !state.isEmpty() && !state.isManagedActionExecuting() && state.canPublish() &&
+                          ((!state.hasAllOnline() && state.hasValidNonOnline()) || state.hasAllPendingDelete()) &&
+                          (state.canModify() || !state.hasAnyInProgress());
+
```
alansemenov: This works 100% as intended, but we need to refactor a little. As the condition you assign here to `this.canPublish` is taken from `PublishContentAction`, it makes sense to move it to `ContentTreeGridItemsState`:
```
    isReadyForPublishing(): boolean {
        return !this.isEmpty() && !this.isManagedActionExecuting() && this.canPublish() &&
                !this.hasAllOnline() && this.hasValidNonOnline() &&
                (this.canModify() || !this.hasAnyInProgress());
    }
```

Note that `this.hasAllPendingDelete()` is removed from the check as we no longer support pending deletion.

Then we can use it in both `MarkAsReadyContentAction` and `PublishContentAction`:

`PublishContentAction`:
```
    isToBeEnabled(state: ContentTreeGridItemsState): boolean {
        return state.isReadyForPublishing();
    }
```
`MarkAsReadyContentAction`:

```
    isToBeEnabled(state: ContentTreeGridItemsState): boolean {
        this.canPublish = state.isReadyForPublishing();

        return this.canPublish && state.hasAllValid() && state.canModify() && state.hasAnyCanBeMarkedAsReady();
    }

```

alansemenov (changes requested):

ashklianko:
> Okay, I added more explicit types. TBH, I find this approach as redundant nowadays, but prettifying and restructuring code will be out of the scope of this task.

There's no redundancy in explicitly specifying types we expect to be used

alansemenov (approved):

edloidas:
@ashklianko I agree for most parts. That is why we need types, sure. 

And here is a quite simple question: how would the types help? They'll make sure that value of one type won't be assigned to property of another type, right? TS in early days and with the default configuration has a feature, that can actually become an issue: `any` type. Basically, everything can be anything. But here is the sweet part: as you gradually add more types, implicit `any` will be replaced by these types. Sure, there are ESLint options and, more importantly, `tsconfig` properties like `noImplicitAny` (also, `noImplicitReturns` and others) that will help avoid using `any` in... any scenarios.

The need of explicit types in function arguments, return type, class properties are pretty obvious. But what I was addressing in the original statement were variables, defined with `const` and, sometimes, `let`.

I believe, that types in these two lines don't make much sense:
```ts
const contentIds: ContentId[] = contentToMark.map(item => item.getContentId());
const isSingleItem: boolean = contentToMark.length === 1;
```

First of all, the variables are defined using `const`: they can't be reassign, they can't be left undefined. `let` is a bit different, though. That is why we need types in cases like:
```ts
// bad
let cotentId;
cotentId = new ContentId('my-content-id');
// Potential error. Won't throw error with current tsconfig
cotentId = 'my-content-id';

// ok
let simpleId = 'my-id';
// TS2322: Type 'string' is not assignable to type 'number'.
simpleId = 1;
```
The TS will automatically set `contentIds` to `ContentId[]` and `isSingleItem` to `boolean`. You can check it by just hovering the variables in you IDE. 
Changing the right side expression type will still cause an error in places, where this variable is used, and you will be forced to fix this.
Next, `isSingleItem` is pretty much descriptive name, that implies `boolean` type here.
Also, jsut look at the arrow functions. Let me use `ad absurdum` argument here and say that setting the return type of function is important, therefore we need set it in arrow (and regular) functions too:
```ts
const contentIds: ContentId[] = contentToMark.map((item: ContentSummaryAndCompareStatus): ContentId => item.getContentId());
```
But we don't do that at all. Why? Because TS is smart enough to detect type in such cases. It reduces readability and gives zero benefits.
You can check it yourself, but almost never set the return type for arrow functions (and the regular ones, that are defined and used in method body).

As a suggestion, I would recommend update the `tsconfig` and `eslint`, when we'll have enough time. This'll allow us write shorter and more readable code, avoiding problems with `any`. And, as I said, we can start doing it now in some cases, because current version of TS already allows it.

Issue #5433, opened 2022-11-15
Change visualisation of inbound references in the Archive dialog


alansemenov:
PR to branch **epic-br-poc-archive**

* [x] Replace the static red chain icon with red "X". Use icon-big-plus class with transparent background and red color, rotate 45 deg.
* [x] Instead of replacing the entire content icon with the static icon in the list, show it over the content icon, in the top right corner, like we do with workflow icons etc.
* [x]  Regular workflow icons(in progress, ready) should not be shown in this dialog.
* [x] Bug: Content items in the top list (main items) are not clickable now. They should be clickable and open Content Wizard in a new tab, as it is in other dialogs.
* [x] For every item (dependent items as well) with inbound references, show the link "_Show references_" which will behave in the same way as the red-chain icon now - open a new tab with the list of dependencies.
* [x] "_Show references_" should be localised.

Before:
![image](https://user-images.githubusercontent.com/11612490/201969730-6f7d62c7-188f-4f39-a2d0-e86e47999cb0.png)

After:
![image](https://user-images.githubusercontent.com/11612490/201969786-144ca652-0d34-4297-b7f3-00f6b7e320ee.png)

PR #5437, opened 2022-11-16
Change visualisation of inbound references in the Archive dialog #5433

edloidas:


alansemenov (changes requested):
Looking great so far but there's a small bug:

You can't click a **dependent** item to open it in a new tab. Cursor changes to "pointer" over the icon but nothing happens on click. Over path cursor won't even change. It should behave in exactly the same way as it does in the top list.

alansemenov:
There's also a linter error which causes the build to fail:

```
/home/runner/work/app-contentstudio/app-contentstudio/modules/lib/src/main/resources/assets/js/app/remove/ContentDeleteDialog.ts
  196:10  error  This assertion is unnecessary since it does not change the type of the expression  @typescript-eslint/no-unnecessary-type-assertion
```

alansemenov (approved):

Issue #5412, opened 2022-11-14
Simplify version comparison for a published content


alansemenov:
To make it easier to see changes made to a published content, add a link "Show changes" next to the status in the Content Preview toolbar. It should only be shown when the status is "Modified".

The link will open the existing modal dialog for version comparison with the following changes:
1. "Newer" version should be set to current version
2. "Older" version should be set to the last published version
3. The section with dropdowns (between the title and comparison panels) should be hidden, so that it's not possible to select other versions.
4. Title should be "Changes since publishing"

![image](https://user-images.githubusercontent.com/11612490/201623648-f5787bf7-e22b-44bc-8e29-1e7d8630c892.png)

<img width="1149" alt="image" src="https://user-images.githubusercontent.com/11612490/201633097-80dc1dfa-4998-4c1b-8ca0-17845e730a06.png">

PR #5443, opened 2022-11-17
 Simplify version comparison for a published content #5412

pmi:


alansemenov (changes requested):
1. Never commit `package-lock` files (unless you are adding a new NPM package). Your branch is behind master so you are committing `package-lock` files with older versions of dependencies.

2. All texts must always be localised (added to `phrases.properties` file, you can also add translation to `phrases_ru.properties` right away). "_Show changes_" is now hardcoded.

3. Modal dialog's title should be "`Changes since publishing`" (#4 of the task description). This text must also be added to `phrases.properties`.

5. Bug: In the Content Wizard when you change a published content status changes to "Modified" before the content is saved, which is correct. What is not correct is that "Show changes" link shows as well. It should not be shown until content is saved, which is when it in fact becomes modified.

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -56,9 +67,19 @@ export class ContentStatusToolbar
     private updateStatus(content: ContentSummaryAndCompareStatus): void {
         this.status.setClass('status');
 
+        if (this.content) {
+            const clazz = this.content.getStatusClass();
```
alansemenov: Let's not do hipster stuff here :)
Call it `statusClass`.

modules/lib/src/main/resources/assets/js/app/dialog/ShowContentChangesDialog.ts
```diff
@@ -0,0 +1,301 @@
...
+import {ContentSummaryAndCompareStatus} from '../content/ContentSummaryAndCompareStatus';
+import {ContentVersionsLoader} from '../view/context/widget/version/ContentVersionsLoader';
+import {ContentVersions} from '../ContentVersions';
+import {VersionContext} from '../view/context/widget/version/VersionContext';
+import {ContentVersion} from '../ContentVersion';
+import {ObjectHelper} from '@enonic/lib-admin-ui/ObjectHelper';
+import {Content} from '../content/Content';
+import {ContentJson} from '../content/ContentJson';
+
+export class ShowContentChangesDialog
```
alansemenov: This should be called `ShowPublishedVersionChangesDialog`

modules/lib/src/main/resources/assets/js/app/ContentVersions.ts
```diff
@@ -33,6 +48,20 @@ export class ContentVersions {
...
+                return;
+            }
+            let pt = pi.getTimestamp().getTime();
+            if (!publishedTime || pt > publishedTime) {
+                publishedTime = pt;
+                publishedVersionId = cv.getId();
+            }
+        });
+
+        return new ContentVersions(contentVersions, activeVersionId, publishedVersionId);
```
alansemenov: This is not the right place to do it. This should be done in the constructor of `ContentVersions`, not here. Also all "pi", "pt" and so on must be replaced with human-readable names.

This should also be done differently:
```
const publishedVersions = contentVersions
    .filter(version => version.isPublished())
    .sort((version1, version2) => version2.getTimestamp().getTime() - version1.getTimestamp().getTime());
this.publishedVersionId = publishedVersions[0]?.getId();
```

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -56,9 +67,19 @@ export class ContentStatusToolbar
...
+            this.showChanges.removeClass(clazz);
+            this.status.removeClass(clazz);
```
alansemenov: `this.status.setClass()` has already set just one class on `this.status`, so there's no point in calling `this.status.removeClass` - it will never do anything.

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -56,9 +67,19 @@ export class ContentStatusToolbar
...
+        }
+
+        this.content = content;
```
alansemenov: This should be in `setItem()`, not in `updateStatus()`

pmi:
> 1. Never commit `package-lock` files (unless you are adding a new NPM package). Your branch is behind master so you are committing `package-lock` files with older versions of dependencies.

It's actually the other way around -- since I did `npm install` just yesterday my package-lock contains newest dependency versions allowed by the package.json constraints.

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -56,9 +67,19 @@ export class ContentStatusToolbar
...
+
```
alansemenov:
```
private updateStatus(content: ContentSummaryAndCompareStatus): void {
    const contentStatus = content?.getStatusClass() || '';

    if (content) {
        this.status.setHtml(content.getStatusText());
    }

    this.status.setClass(`status ${contentStatus}`);
    this.showChanges.setClass(`show-changes ${contentStatus}`);
}
```

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/ContentVersions.ts
pmi: Agree with the pt, pi, and the placement, but don't see a point in increasing execution time:
your example does 2 cycles (one of them being sorting that we don't need), while I did it in 1 go.

alansemenov: That's about having a more readable code. First only select published versions, then sort them and pick the top one. There will never be so many versions that it will affect execution time.

Let's keep your version but make it more readable and extract it to a separate method:

```
private getLastPublishedVersionId(): string {
    // We don't need to do anything if content is currently not published
    if (!this.getActiveVersion().isPublished()) {
        return;
    }

    let lastPublishedTime: number = 0;
    let lastPublishedVersionId: string;
    this.contentVersions.forEach((contentVersion: ContentVersion) => {
        if (!contentVersion.isPublished()) {
            return;
        }
        const publishedTime = contentVersion.getPublishInfo().getTimestamp().getTime();
        if (publishedTime > lastPublishedTime) {
            lastPublishedTime = publishedTime;
            lastPublishedVersionId = contentVersion.getId();
        }
    });

    return lastPublishedVersionId;
}

```

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -30,6 +32,15 @@ export class ContentStatusToolbar
...
     protected initElements(): void {
         this.status = new SpanEl('status');
+        this.showChanges = new AEl('show-changes');
+        this.showChanges.setHtml('Show changes');
+        this.showChanges.onClicked(() => this.openShowContentChangesDialog());
+    }
+
+    private openShowContentChangesDialog() {
+        ShowContentChangesDialog.get()
+            .setContent(this.content)
```
alansemenov:
```suggestion
            .setContent(this.getItem())
```

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
alansemenov: We don't need `this.content` at all - `this.getItem()` will give us content previously stored by `this.setItem()` (see the suggestion above).

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -30,6 +32,15 @@ export class ContentStatusToolbar
 
     protected initElements(): void {
         this.status = new SpanEl('status');
+        this.showChanges = new AEl('show-changes');
+        this.showChanges.setHtml('Show changes');
+        this.showChanges.onClicked(() => this.openShowContentChangesDialog());
+    }
```
alansemenov, line 35:
```suggestion
    this.showChanges().hide;
    }
```

modules/lib/src/main/resources/assets/styles/content-status-toolbar.less
```diff
@@ -11,5 +11,23 @@
         display: none;
       }
     }
+
+    .show-changes {
+      margin-left: 5px;
+      text-decoration: none;
+      display: none;
```
alansemenov: Remove `display: none;` - it's better to do this via code.

modules/lib/src/main/resources/assets/styles/content-status-toolbar.less
```diff
@@ -11,5 +11,23 @@
...
+      color: @admin-button-blue2;
+      font-style: normal;
+
+      &.modified {
+        display: inline;
+      }
```
alansemenov: Remove.

alansemenov (changes requested):
It seems that the best and simplest solution to keep the "Show changes" link hidden when a published content gets modified (but not yet saved) is to unlink it from the status text and show/hide it on data change by comparing content's modified time before and after the changes.

1) Initially hide the link (inside `initElements()`)
2) `Inside ContentStatusToolbar.setItem` hide the link if content status is not modified or  modified time hasn't changed:

```Typescript
setItem(item: ContentSummaryAndCompareStatus): void {
    if (item && !item.equals(this.getItem())) {
...

        const isModified = content.isModified() &&
            this.getItem()?.getContentSummary().getModifiedTime() !== content.getContentSummary().getModifiedTime();
        this.showChanges.setVisible(isModified);

        super.setItem(content);

...
        this.updateStatus(content);
    }
}
```

3) Clean up `updateStatus` method so that it doesn't do anything with the "Show changes"  link:

```Typescript
private updateStatus(content: ContentSummaryAndCompareStatus): void {
    const contentStatus = content?.getStatusClass() || '';

    if (content) {
        this.status.setHtml(content.getStatusText());
    }

    this.status.setClass(`status ${contentStatus}`);
}
```

sigdestad:
How about always autosaving when you click "show changes"? Save vs state will be a problem also when you modify an already modified and saved item. 

We use a similar solution when clicking preview button.

alansemenov:
We found a solution. Let's not overcomplicate this with even more features.

sigdestad:
But, will it actually work then? Or will I only be able to se modifications after saving?

alansemenov:
Only after saving.

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -17,6 +15,8 @@ export class ContentStatusToolbar
 
     protected status: SpanEl;
 
+    protected showChanges: AEl;
```
alansemenov: Make private and call showChangesLink

modules/lib/src/main/resources/assets/js/app/ContentStatusToolbar.ts
```diff
@@ -30,6 +30,10 @@ export class ContentStatusToolbar
 
     protected initElements(): void {
         this.status = new SpanEl('status');
+        this.showChanges = new AEl('show-changes');
+        this.showChanges.setHtml('Show changes');
+        this.showChanges.onClicked(() => this.openShowPublishedVersionChangesDialog());
+        this.showChanges.hide();
```
alansemenov: Extract to `createShowChangesLink` function

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardToolbar.ts
```diff
@@ -93,6 +96,17 @@ export class ContentWizardToolbar
...
+        ContentServerEventsHandler.getInstance().onContentUpdated((items) => {
+            const currentContentId = this.getItem()?.getContentId();
+            const updatedContent = currentContentId && items.find((item) => item.getContentId().equals(currentContentId));
+            if (updatedContent) {
+                // new version of content has been saved
+                // do not listen for it in setItem because it's runtime version of content there with manipulated state
+                this.isSavedVersionModified = updatedContent.isModified();
+                this.updateStatus(updatedContent);
+            }
+        });
```
alansemenov: Extract to function called `updateStatusForPersistedVersion`

modules/lib/src/main/resources/i18n/dialogs.properties
```diff
@@ -102,6 +102,9 @@ dialog.compareVersions.previousVersion=Previous version
 dialog.compareVersions.nextVersion=Next version
 dialog.compareVersions.newestVersion=Newest version
 dialog.compareVersions.versionSubName={0} by {1}
+dialog.publishedChanges.header=Changes since publishing
+dialog.publishedChanges.showUnchanged=Show entire content
```
alansemenov: This text is already used in the Compare Versions dialog, so we need to use the same localisation text in both dialogs.

alansemenov (changes requested):

sigdestad:
So, if I open a modified item - it will display "show changes". If I edit, it goes away until I save it? 
Which is more complex - autosaving when user clicks the link, or the link toggling?

For sure, the user will probably be confused by link appearing/disappearing.

pmi:
No, if you open modified item then show changes link will always be visible ! 
It is not shown until you save content that is currently in published state.

sigdestad:
So, If I make a lot of changes to a modified item, and then click show changes - I will only see what was changed before I started editing?

pmi:
Correct ! It only compares latest saved version with published one.

Issue #5438, opened 2022-11-16
Add `variantOf` field to Content data object and classes


alansemenov:
Add `variantOf` field to Content object, Content json and other content-related TS classes.

PR #5448, opened 2022-11-21
Add variantOf field to Content data object and classes #5438

anatol-sialitski:


modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
```diff
@@ -1216,6 +1222,26 @@ public AbstractContentQueryResultJson query( final ContentQueryJson contentQuery
             .execute();
     }
 
+    @GET
+    @Path("getVariants")
+    public ContentListJson<ContentSummaryJson> getVariants( @QueryParam("id") final String id,
```
rymsha:
```suggestion
    public ContentListJson<ContentSummaryJson> findVariants( @QueryParam("id") final String id,
```

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
```diff
@@ -1216,6 +1222,26 @@ public AbstractContentQueryResultJson query( final ContentQueryJson contentQuery
             .execute();
     }
 
+    @GET
+    @Path("getVariants")
```
rymsha:
```suggestion
    @Path("findVariants")
```

rymsha (approved):

alansemenov (approved):

Issue #5435, opened 2022-11-15
Refresh contents of the Archive dialog on updates of inbound references


alansemenov:
PR to branch **epic-br-poc-archive**

Listen to server events for update and delete, and refresh those content items whose inbound dependency was updated or deleted (check `inboundDependencies` array from the `resolveForDelete` response).

It's important that the Archive button automatically enables and red section at the top disappears if all of the inbound references have been removed.

PR #5466, opened 2022-11-22
Refresh contents of the Archive dialog on updates of inbound references #5435

edloidas:
_Depends on #5446_


alansemenov (changes requested):
Great job, @edloidas! Works as intended and the whole experience is very smooth. Two small things to fix:
1. We now refresh the dialog when references get _deleted_. But we should also refresh when they get _archived_. This will happen quite often.
2. We should also refresh the dialog when one of the items in the list _gets_ (not _loses_) an inbound reference while the dialog is open. This won't happen often, so do this only if it's not a complex fix.

alansemenov (approved):
Great stuff

PR #5467, opened 2022-11-22
Auto-save modified content before comparing it with published version

alansemenov:


ashklianko (commented):
1. compareVersionsPreHook is only used in ContentWizardToolbar, so can move ot there (we don't know for now if this hook will be ever needed in the  browse panel)
2.  We can fire a custom event when compare link is pressed, so those who need it will react, and we won't need then pass any extra handlers to the toolbar

alansemenov:
> * compareVersionsPreHook is only used in ContentWizardToolbar, so can move ot there (we don't know for now if this hook will be ever needed in the  browse panel)

Yes, let's do this for now.

> * We can fire a custom event when compare link is pressed, so those who need it will react, and we won't need then pass any extra handlers to the toolbar

No. The whole idea here is that the hook should finish before the link action will be executed. Event listener will be asynchronous, so it won't work.

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -598,7 +598,17 @@ export class ContentWizardPanel
...
+            className: 'content-wizard-toolbar',
+            compareVersionsPreHook: () => {
+                if (!this.hasUnsavedChanges()) {
+                    return Q.resolve();
+                }
+
+                const deferred: Q.Deferred<void> = Q.defer<void>();
+                this.saveChangesWithoutValidation().then(() => deferred.resolve());
+
+                return deferred.promise;
```
pmi: Can be simplified to:
```
return this.saveChangesWithoutValidation().then(() => null);
```

alansemenov: No, it's not the same. The prehook method returns a promise which gets resolved after save. Simply returning null won't resolve the promise.

pmi: Returning simple value or throwing an exception inside a promise resolves that promise so that will work ;)

alansemenov: Changed return type to `Q<any>` and changed the hook body to:
```
compareVersionsPreHook: () => {
    if (!this.hasUnsavedChanges()) {
        return Q.resolve();
    }

    return this.saveChangesWithoutValidation();
}
```

pmi (approved):

Issue #5436, opened 2022-11-16
Enable adding a new content directly from Content Selector


alansemenov:
We will simplify the process of adding a new content item to Content Selector's dataset by allowing to do this directly from the input.

1. If content type of the content in context allows children, show "+" button to the right of the content selector (similar to how we show upload button for Image/Media selector)
2. The "+" button will open the "New Content" dialog. 
3. Content types in this dialog should be filtered by a) allowed child content types for the parent content, b) `allowContentType` config of the content selector (if any). Theoretically, intersection of these two can result in empty list, so we should probably perform this check early on and hide the "+" button if no child content types allowed. 
4. Once a content type for a child content is selected, create a child content item and open a new browser tab with Content Wizard for the created child item.
5. Once the item is created, it should get automatically selected in the Content Selector in the parent content (the Save button should then become enabled, if it wasn't before). Every time the child item's display-name or path are updated, we should update the selected option in the Content Selector.

![image](https://user-images.githubusercontent.com/11612490/202146211-38aa83bf-9a78-4eac-afc0-aae08068ff9d.png)

PR #5470, opened 2022-11-23
Enable adding a new content directly from Content Selector #5436

ashklianko:


alansemenov (changes requested):
1. It looks like this after I have added a new content to a _single-select_ selector:
image

"+" icon should've been replaced with "x" once an option is selected.

For a _multi-select_ selector everything looks fine:
image

2. Add a tooltip to the "+" icon: "`Add new <display name of the selector>`", f.ex. "`Add new Author`" (should be added to i18n property files). If it's difficult to get a hold of display name, make it simply "`Add new`".

Otherwise it looks great!

ashklianko:
Updated!

alansemenov (approved):

PR #5475, opened 2022-11-25
Display an icon for variants

alansemenov:
Default look of variants:

image

For inherited variants inside layers the blue "cloud" icon (indicating that the item is inherited) will be replaced with "variant" icon with blue background

image


sigdestad:
FYI - while we're fixing this, maybe change the normal "inherited" icon from a cloud to just a round blue backround as well? Will be more consistent - and less intrusive than the cloud icon.

alansemenov:
> FYI - while we're fixing this, maybe change the normal "inherited" icon from a cloud to just a round blue backround as well? Will be more consistent - and less intrusive than the cloud icon.

Just a blue circle? How does it tell user that this content is inherited? Or a blue circle with white cloud inside?

sigdestad:
Blue circle with down arrow inside.

ashklianko (approved):

Issue #5505, opened 2022-12-05
Text input with value validation


ashklianko:
Rename content dialog has an input with content name availability check, however this input with it's styling may be reused in other places (Will be used in variants create dialog).

PR #5506, opened 2022-12-05
Text input with value validation #5505

ashklianko:


alansemenov:
Moved to https://github.com/enonic/app-contentstudio/pull/5535

PR #5507, opened 2022-12-05
Support filtering by regular expressions in `schema/filter/contentTypes`

ashklianko:


alansemenov (changes requested):
Rename `skipValidation` to `displayAsNew`.

When set to `true`:
* skip highlighting of invalid fields
* give a notification about created item

alansemenov (approved):

Issue #5497, opened 2022-12-01
Publishing Wizard improvements


alansemenov:
There are three types of item state which can prevent an item from being published:
* the item is invalid
* the item is in progress
* current user doesn't have publishing permissions for the item

We show notifications about all three but they are not always consistent and obvious.

Do the following:
* Add an artificial delay (500 ms) to show the checks are being executed. Show a spinner + "_Running checks_" text over grey background during this delay. This should be done in parallel with actual resolution of items in the list. "Publish" button should be disabled while the checks are being run.

![image](https://user-images.githubusercontent.com/11612490/205064238-853d697a-29a2-4e4f-8fa8-6bf270471380.png)

* If there are publishing blockers:
  * Use red(ish) background for the notification section - same color as we used in the Archive dialog.
  * We already show icons for invalid and in-progress items, but for insufficient publishing permissions we show a toast notification (and hide the Publish button or something like that). Remove this notification and show icon + "_Permissions (x)_" text + link to exclude these items from publishing (if they can be excluded). Use the same red "X" icon as used in the Archive dialog. 
  * All three notifications should be consistent and behave in the same way.

![image](https://user-images.githubusercontent.com/11612490/205063395-e795f6bd-5ad2-4b99-99f0-d74adc393639.png)

* If there are no blockers:
  * Use light-green background for the notification section and show green "V" icon + text "_Ready for publishing_"

![image](https://user-images.githubusercontent.com/11612490/205065980-05da0e8d-c069-4cc6-89e2-44610caa8da9.png)

PR #5539, opened 2022-12-06
Publishing Wizard improvements #5497

edloidas:
Replaced a single warning line with more complex issues state bar, that will be universal among multiple dialogs.
Replaced issues state bar in publish dialog.
Made Content resource return number of not publishable content.
Made Content resource return number of not publishable content.
Most of the markChecking() calls are made for smoother UI transitions.


alansemenov:
@vbradnitski please review the Java part

vbradnitski (approved):

ashklianko:
When opening a publishing dialog with many items multiple items it may show for a short moment that all checks are passed and green:

https://user-images.githubusercontent.com/4255706/206406227-c5369af9-947a-48c5-97b0-9e5b34550bf9.mp4

ashklianko:
Don't we need the same status bar in ContentUnpublishDIalog?

ashklianko:
Looks good @edloidas !

modules/lib/src/main/resources/assets/js/app/dialog/DialogErrorsStateBar.ts
```diff
@@ -0,0 +1,152 @@
...
+            className: 'resolved-entry',
+            text: resolvedText,
+            iconClass: 'icon-checkmark',
+        });
+
+        this.updateResolvedState();
+    }
+
+    protected updateResolvedState(): void {
+        const isResolved = !this.errorsEntries.some(entry => entry.isActive());
```
ashklianko: `const isResolved: boolean = !this.errorsEntries.some((entry: DialogErrorStateEntry) => entry.isActive());`
I believe we have agreed long ago to always declare types on variables to make it easier to read, even though some types can be inferred, however in this case it requires to go and check what is entry's type

modules/lib/src/main/resources/assets/js/app/dialog/DialogErrorsStateBar.ts
```diff
@@ -0,0 +1,152 @@
...
+        entry.onActiveStateChange(() => this.updateResolvedState());
+        this.updateResolvedState();
+
+        entry.onCheckingStateChange(() => this.updateChecking());
+
+        return entry;
+    }
+
+    setEnabled(enabled: boolean): void {
+        this.errorsEntries.forEach(entry => entry.setActionEnabled(enabled));
```
ashklianko: Same here: `entry` is missing type

modules/lib/src/main/resources/assets/js/app/dialog/DialogErrorsStateBar.ts
```diff
@@ -0,0 +1,152 @@
...
+        this.addClass(Modifiers.FAILED);
+    }
+
+    reset(): void {
+        this.errorsEntries.forEach(entry => entry.reset());
+        this.removeClass(Modifiers.FAILED);
+    }
+
+    markChecking(checking: boolean): void {
+        this.errorsEntries.forEach(entry => entry.markChecking(checking));
```
ashklianko: Same here: `entry` is missing type

modules/lib/src/main/resources/assets/js/app/dialog/DialogErrorsStateBar.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    toggleHideIfResolved(hideIfResolved: boolean): void {
+        if (this.hideIfResolved !== hideIfResolved) {
+            this.hideIfResolved = hideIfResolved;
+            const isResolved = !this.errorsEntries.some(entry => entry.isActive());
```
ashklianko: Types are missing

alansemenov:
> Don't we need the same status bar in ContentUnpublishDIalog?

Correct, but this will be done in a separate task.

modules/lib/src/main/resources/assets/js/app/dialog/DialogErrorsStateBar.ts
edloidas: Oh boy, we haven't even got started. Guess we'll have to solve this with a one-on-one fight with knifes.

edloidas:
Fixed the errors bar state problem.
It appears, that the initial problem behind the invalid errors bar was there for quite a long time and by using the debounced onLoadingFinished handlers just made it worse.
Because of the unholy mix of multiple debounced handlers, events and promises, that are used simultaneously, the handlers were triggered when they must not (visually this problem was affecting loading mask and controls lock state earlier).

Closing the dialog was triggering unnecessary delayed notifications, that had effect if you were quick enough to open it before the close callback were actually done running. Or you can just open the dialog, close it, and open again (in case the background checks will take long enough) to get double the notifications that will break the UI.

If we are to use it somewhere else in the future or add more functionality, I would recommend to write the publish processor from scratch if there'll be time.

alansemenov:
Moved to https://github.com/enonic/app-contentstudio/pull/5562

Issue #5482, opened 2022-11-29
Add visualisation of project structure to the Settings tree


alansemenov:
When selecting "Project" folder in the tree, display project structure in a graph inside the right-hand preview panel. Project nodes should be styled in the same way as in the settings tree.

Don't implement navigation as of now.

reisfmb:
@sigdestad @alansemenov 

Here are some previews of the current implementation:

![image](https://user-images.githubusercontent.com/67838246/206103961-e8516c77-0feb-4fdc-93c2-62f47eac325c.png)

![image](https://user-images.githubusercontent.com/67838246/206103658-1a004c9c-e239-4941-a732-75eaa9eaee7d.png)

![image](https://user-images.githubusercontent.com/67838246/206103438-e87c6f27-d90f-414e-80b8-856106c97bbf.png)

@alansemenov I see that you mentioned [here](https://github.com/enonic/app-contentstudio/issues/5482#issue-1468218008) that the nodes should look the same as in the grid, and I tried to do that... you can see that their structure (flag + display name + id) is the same, but I needed to find a way to wrap something around them because of the edges of the graph, so I added that black pill wrapping that structure.

Waiting for feedbacks to keep improving the design and implementing new features.

alansemenov:
Looking good, Bruno! It's now a matter of a simple CSS tweak to make the nodes look like in the tree - just remove the color fill and keep a thin border.

Personally, I don't see much point in showing weight - these numbers are very confusing and we don't really have a concept of weight, we only have "main" parent and the rest. What we could do is have a thick line between a layer and its main parent and normal/thin lines between the layer and the rest of the parents.

PR #5547, opened 2022-12-08
Project and layers DAG visualization #5482

reisfmb:


reisfmb:
@alansemenov 

PR is in draft because it remains to fix some things... some of those things are code related, while others are feature related, such as refreshing the visualization after adding a new project / layer. Currently there's the need to highlight again the project item in the tree grid.

Feel free to test it locally :)

alansemenov (changes requested):
1. Container of the graph has incorrect dimensions which makes it look bad (see the screenshot below - connecting lines go out of the view). It seems that you are using width of the entire page for `svg-container` but you should use width of the preview panel (even a bit less, so that lines never touch the panel borders). Actually, after checking it a bit more, it seems that the width and height are hardcoded (2000px and 600px), but they should dynamically change based on width/height of the preview panel. See examples using `ResponsiveManager` in the code.

image

2. There's no difference in line thickness based on whether parent is "main" or not. All the lines have the same thickness.

3. Lots of Codacy issues which can be seen [here](https://github.com/enonic/app-contentstudio/pull/5547/files#diff-a304fab231d4b70e9c34b3540c1a00e645605aa64ca46c5285c458cc4b4395f0). Seems like you indented with a tab, but it should be two spaces.

reisfmb:
@alansemenov 

1. They are not actually hardcoded, but calculated. Different dags will have different widths / heights. But the problem is that this calculation was actually wrong. Fixed it.

2. This is working, but I think you couldn't see it because the visualization was out of screen. Paths connecting projects to layers will have `stroke-weight` of 3, while other will have `stroke-weight` of 2.

3. Will check Codacy before converting this PR from draft to ready.

I've implemented some fixes + improvements on node positioning + paths. Here's an example of how it is looking:

image

Could you please test it again?

alansemenov (approved):
Great job, Bruno!

PR #5584, opened 2022-12-13
name and path fix

ashklianko:


modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardHeader.ts
```diff
@@ -233,4 +233,24 @@ export class ContentWizardHeader
             listener.call(this);
         });
     }
+
+    update(displayName: string, name: string, silent?: boolean): void {
```
alansemenov: If we are giving this method a general name like `update` then it should accept `content` as an argument (instead of `displayName` and `name`) to make it easily extendable to update other properties without changing the signature.

If it's intended solely for updating `displayName` and `name`, then it should be renamed to something like `updateNames`.

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardHeader.ts
ashklianko: it's not finished yet, just wanted to run tests before lunch :)

alansemenov (approved):

PR #5640, opened 2022-12-29
New Сontent dialog allows upload of media when child content is not allowed

### LINKED ISSUE #4407: New Сontent dialog allows upload of media when child content is not allowed
1. Add this part in a content type config
```
 <allow-child-content>true</allow-child-content> 
  <allow-child-content-type>base:folder</allow-child-content-type> 
```
2. Create a content.
3. Select the content and open `New Content Dialog`

**BUG**: Upload button should be hidden in the dialohg, because only folder type is allowed for child content

image

Try to upload a content, error notification message appears: 

image

ashklianko:


modules/lib/src/main/resources/assets/js/app/create/NewContentDialog.ts
```diff
@@ -301,6 +304,48 @@ export class NewContentDialog
...
+        this.newContentUploader.setVisible(true);
+        this.newContentUploader.setEnabled(true);
+        this.removeClass('no-uploader-el');
+    }
+
+    private handleUploaderNotAllowed(): void {
+        this.addClass('no-uploader-el');
+        this.newContentUploader.setEnabled(false);
+        this.newContentUploader.setVisible(false);
+    }
```
alansemenov: Can we combine this in one method?

```
private setUploaderAllowed(isAllowed: boolean): void {
    this.toggleClass('no-uploader-el', !isAllowed);
    this.newContentUploader.setEnabled(isAllowed);
    this.newContentUploader.setVisible(isAllowed);
}
```

modules/lib/src/main/resources/assets/js/app/create/NewContentDialog.ts
```diff
@@ -301,6 +304,48 @@ export class NewContentDialog
...
+
+    private handleUploaderByContentType(parentContentType: ContentType): void {
+        if (parentContentType.getAllowChildContentType()?.length > 0) {
+            if (parentContentType.getAllowChildContentType().some((allowedType: string) => this.isMediaType(allowedType))) {
```
alansemenov: This logic belongs in the `ContentTypeSummary` class. We also have static `ContentTypeName.MEDIA` and `ContentTypeName.isDescendantOfMedia` which should be used.

Here it should be something very simple, like

```Typescript
private updateUploaderState(): void {
  this.newContentUploader.reset();
  
  if (!this.parentContent) {
      this.setUploaderAllowed(true);
  } else if (this.isTemplateFolderSelected()) {
      this.setUploaderAllowed(false);
  } else {
      new GetContentTypeByNameRequest(this.parentContent.getType())
  .sendAndParse()
  .then((contentType: ContentType) => this.setUploaderAllowed(contentType.isMediaChildContentAllowed()))
  .catch(DefaultErrorHandler.handle);
  }
}
```

alansemenov (changes requested):

alansemenov (approved):

PR #5681, opened 2023-01-04
Bad handling of a site with a page template that returns a rendering …

### LINKED ISSUE #5678: Bad handling of a site with a page template that returns a rendering error
When a site is renderable (has a page template) but preview returns an error, Content Studio will behave in a strange way: the editor panel will simultaneously show error message and controller dropdown, and for some reason the content will be endlessly updated.

ashklianko:
…error #5678

test


alansemenov (approved):
Great job!

PR #5719, opened 2023-01-10
 Delay page editor initialization on demand by a page inside it #5718

### LINKED ISSUE #5718: Delay page editor initialization on demand by a page inside it
React does hydration on the client where it checks that html matches that one rendered on the server.
Page editor starts its initialization on window load and makes changes to page html that may break the above comparison.

I propose the following:
- page editor suspends it's initialization for 100 msec, if nothing happens it continues as before
- There are 2 utility functions in @enonic/nextjs-adapter that throw events to content studio: **notifyRenderingStarted()** and **notifyRenderingFinished()**
- If you trigger first one in the first 100 msec after the window load event, page editor will not start loading until the second function is called or 3 sec passed (in case that function is never called)
- If 2nd function is triggered before 3 sec have passed, page editor continues it's initialization

pmi:


alansemenov (changes requested):
Changes as discussed:
1. Change event names to `ContentRendering` and `ContentRendered`
2. Pass timeout as an optional argument of `ContentRendering` event object. If it's not passed, PageEditor won't be initialised.
3. Remove event handlers when they are no longer needed.

pmi:
Opted to go for disabling hydration completely in edit mode

PR #5724, opened 2023-01-11
Uploaded media doesn't get selected in Content Selector #5691

### LINKED ISSUE #5691: Uploaded media doesn't get selected in Content Selector
1. Open wizard for new `Selector` in `Features`.
2.  click on `Add new` icon

image

3. Click on `Upload` button then select an image, for example

**BUG** -  Uploaded image don't get selected in  `Allow all items from this site`, `Allow all content items`  inputs.
           The  same issue in `Shoertcut selector`

ashklianko:


modules/lib/src/main/resources/assets/js/app/inputtype/selector/ui/NewContentButtonUploaderHandler.ts
```diff
@@ -0,0 +1,127 @@
+import {ProgressBar} from '@enonic/lib-admin-ui/ui/ProgressBar';
+import {UploadItem} from '@enonic/lib-admin-ui/ui/uploader/UploadItem';
+import {Content} from '../../../content/Content';
+import {Element} from '@enonic/lib-admin-ui/dom/Element';
+import {FormItemEl} from '@enonic/lib-admin-ui/dom/FormItemEl';
+
+export class NewContentButtonUploaderHandler {
```
alansemenov: This class is quite generic and its main purpose is to visualise upload progress (not necessarily for new content and not necessarily inside a button), so it should be renamed to something like "`UploadProgressBar`". I would also make it even more generic by replacing `Content` type with generic `<T>` (`export class UploadProgressBar<T>`) but it's not important right now.

modules/lib/src/main/resources/assets/js/app/inputtype/selector/ui/NewContentButtonUploaderHandler.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    private itemUploadedHandler?: (content: Content) => void;
+
+    private itemUploadFailedHandler?: (item: UploadItem<Content>) => void;
+
+    constructor(parentElement: Element) {
+        this.progressBar = new ProgressBar(0);
+        this.parentElement = parentElement;
+
+        this.parentElement.appendChild(this.progressBar);
```
alansemenov: Maybe this should be done in `handleItemProgress` and then removed from `this.parentElement` once the progress is either finished or failed? We don't need to keep it in the DOM unless it's visible. Then we won't need `UPLOADING_CLASS` and CSS tweaks either.

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/inputtype/selector/ui/NewContentButtonUploaderHandler.ts
ashklianko: We need CSS class at least to hide other elements contained in a parent element

alansemenov (approved):

PR #5732, opened 2023-01-13
Content properties in the "Details" widget should be lazy-loaded #5543

### LINKED ISSUE #5543: Content properties in the "Details" widget should be lazy-loaded
Whenever an item in the Content grid is selected, we send several requests to fetch data that will be displayed in the "Details" widget. This happens regardless of whether the "Details" widget is selected or not. Some of these requests, like `effectivePermissions`, can be quite expensive.

Change the logic so that we send requests only when the widget gets selected the first time. Show spinners inside each widget section (only those where we fetch the data) until the data is loaded and displayed.

image

ashklianko:
-changed z-index for preview panel's load mask so it doesn't overlap open widget


alansemenov (approved):
Looking good!

PR #5792, opened 2023-01-27
Move "Mark as Ready" action in the Publishing Wizard #5788

### LINKED ISSUE #5788: Move "Mark as Ready" action in the Publishing Wizard
"Mark as Ready" is now displayed as a menu-item of the Publish menu button. Make it a part of the "Exclude items in progress" link instead. It should still be displayed or hidden based on the config value, as now.

image

edloidas:


alansemenov (changes requested):
After clicking the "Mark as ready" link the top section with checks first changes to "Running checks", then changes _back_ to the red background with "In progress (X)", then changes to green "ready for publishing". See attached video (I throttled down browser speed to make it very obvious).

We should fix this to be smoother. It should never go back to the original "In progress / Mark as ready" state if the link has already be clicked. It should be like this: "Mark as ready" clicked -> change to grey background with "_Marking content as ready_" text -> start "Running checks..." -> new state of checks (either back to red or change to green).


https://user-images.githubusercontent.com/11612490/215528430-3834be80-455b-4047-bc83-84b0096ca9e5.mov

alansemenov (approved):

Issue #5801, opened 2023-01-30
`allow-content-type` config of ContentSelector is not working inside site config


alansemenov:
`allow-content-type` config of ContentSelector in site.xml doesn't seem to work - it returns all elements.

site.xml:
```XML
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<site xmlns="urn:enonic:xp:model:1.0">
  <form>
    <input type="ContentSelector" name="searchPage">
      <label>Search page</label>
      <occurrences minimum="0" maximum="1"/>
      <config>
        <allow-content-type>landing-page</allow-content-type>
      </config>
    </input>
  </form>
</site>
```

![image](https://user-images.githubusercontent.com/11612490/215503323-841eea7c-f1c4-4fab-98d7-520ffc673080.png)

Same input works correctly inside a content type:

![image](https://user-images.githubusercontent.com/11612490/215503669-679e92d2-d693-4965-b8a0-5619a7377993.png)

PR #5815, opened 2023-01-31
allow-content-type config of ContentSelector is not working inside si…

ashklianko:
…te config #5801

### DIFF
--- a/modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentSelectorQueryJsonToContentQueryConverter.java
+++ b/modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentSelectorQueryJsonToContentQueryConverter.java
@@ -1,10 +1,26 @@
 package com.enonic.xp.app.contentstudio.rest.resource.content;
 
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import com.google.common.base.Preconditions;
+
+import com.enonic.xp.app.ApplicationKey;
 import com.enonic.xp.app.ApplicationWildcardMatcher;
 import com.enonic.xp.app.contentstudio.rest.resource.content.json.ContentSelectorQueryJson;
-import com.enonic.xp.content.*;
+import com.enonic.xp.content.Content;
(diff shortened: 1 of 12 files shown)


modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentSelectorQueryJsonToContentQueryConverter.java
```diff
@@ -63,27 +74,53 @@ public ContentQuery createQuery()
...
-                contentTypeService.getAll().stream().map( ContentType::getName ).filter( filter ).collect( Collectors.toList() ) );
+            return this.filterContentTypeNames( applicationKey );
         }
 
         return ContentTypeNames.from( contentTypeNames );
     }
 
+    private ApplicationKey getApplicationKey()
+    {
+        if ( this.contentQueryJson.getApplicationKey() != null )
```
vbradnitski, line 96: It would be nice to have it tested

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/json/ContentSelectorQueryJson.java
```diff
@@ -36,7 +39,8 @@ public ContentSelectorQueryJson( @JsonProperty("queryExpr") final String queryEx
                                      @JsonProperty("inputName") final String inputName,
                                      @JsonProperty("contentTypeNames") final List<String> contentTypeNamesString,
                                      @JsonProperty("allowedContentPaths") final List<String> allowedContentPaths,
-                                     @JsonProperty("relationshipType") final String relationshipType )
```
vbradnitski, line 39: there is a chance that `relationshipType` is not used any more

vbradnitski (changes requested):

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/json/ContentSelectorQueryJson.java
alansemenov: this is not a part of this task

Issue #5818, opened 2023-01-31
Changing a component config reloads the entire page


alansemenov:
See the attached video. Change of a component should only reload the changed component (and also don't make the Inspect panel blink several times).

https://user-images.githubusercontent.com/11612490/215780517-65bccbed-e19b-43fc-8d52-c80d31f05ddd.mov

PR #5821, opened 2023-02-01
Changing a component config reloads the entire page #5818

ashklianko:


modules/lib/src/main/resources/assets/js/app/page/region/Regions.ts
```diff
@@ -15,31 +15,25 @@ import {ComponentPropertyChangedEventHandler} from './Component';
 export class Regions
     implements Equitable {
 
-    public static debug: boolean = false;
+    public static debug: boolean = true;
```
alansemenov: This should be `false`.

PR #5958, opened 2023-02-28
ContentEventsProcessor: fix edit urls generation #5957

### LINKED ISSUE #5957: ContentEventsProcessor: fix edit urls generation
`ContentEventsProcessor` is located in `lib-contentstudio`, thus it might be used throughout any app (and it is used in _studio.plus_ currently), however when the link for content edit is generated it fetches local value of `toolUri` config variable and will generate incorrect edit url.

Add `studioToolUri` to configs with generated link to the studio app and add a new method into `UrlHelper` class that will use this variable. Make `ContentEventsProcessor` use this new method. Thus if any side app will use `ContentEventsProcessor` without `studioToolUri` var in a config then edit won't open and there will be an error in the dev console

ashklianko:


modules/app/src/main/resources/services/config/config.js
```diff
@@ -22,10 +23,8 @@ function handleGet() {
             assetsUri: portal.assetUrl({
                 path: ''
             }),
-            toolUri: admin.getToolUrl(
-                app.name,
-                'main'
-            ),
+            toolUri: studioToolUri,
```
alansemenov: Why do we need two properties with the same value?

alansemenov (approved):

PR #5975, opened 2023-03-02
Move online from/to fields to a separate box in the Details widget #5966

### LINKED ISSUE #5966: Move online from/to fields to a separate box in the Details widget
1. Move "Online from" and "Online to" fields to a separate box. The box should be hidden when content is not in the master branch.
2. Link for opening the dialog should say "Edit Schedule" (new i18n key)
3. Rename "Edit Properties" link to "Edit Settings". It should no longer dependend on whether the content is in master anymore, only on user role (if current role has permissions to change language/owner)
4. Change behaviour of the "Edit properties" dialog so that its header is configurable and that we show either "Schedule" or "Settings" step, based on which box it's opened from.
5. When the dialog is opened from the "Properties" box, the title should be "Edit Settings" (same i18n key as for the link)
6. When the dialog is opened from the "Schedule" box, the title should be "Edit Schedule" (same i18n key as for the link)
7. Save confirmations should be different, based on which "mode" the dialog is on: "`Settings for <content name> are updated`" or "`Schedule for <content name> is updated`"

image

Dialog when opened via "Edit Settings" link:
image

Dialog when opened via "Edit Schedule" link:
image

ashklianko:


alansemenov (changes requested):
1. Empty title in both dialogs
2. We don't need tooltip (bottom right corner)

3. We don't need `properties-wizard-step-form-header` ("Settings" and "Schedule") since it's now only one group of properties in each dialog.

Screenshot 2023-03-02 at 17 48 11

alansemenov (approved):

PR #6038, opened 2023-03-10
Link dialog - Media option default value is 'Open File' #6028

### LINKED ISSUE #6028: Incorrect preselection of media option in the Insert Link dialog
If existing link to a content item in HtmlArea is changed to link to a media content, then selection in "Media options" radio-group won't be correct. It should always have "Open file" as default option but it's preset to "Link to page". See the attached video.

https://user-images.githubusercontent.com/11612490/223761537-e32f62cd-7e58-4b48-b89f-ab00e62fc9cf.mov

reisfmb:


modules/lib/src/main/resources/assets/js/app/inputtype/ui/text/dialog/LinkModalDialog.ts
```diff
@@ -887,6 +891,8 @@ export class LinkModalDialog
         }
 
         if (selectedContent.getType().isDescendantOfMedia()) {
+            const mediaRadio: RadioGroup = <RadioGroup>this.getFieldById('contentMediaRadio');
```
alansemenov: This should be extracted to a separate method `getMediaRadioGroup` which should be used everywhere you are now using `const mediaRadio: RadioGroup = <RadioGroup>this.getFieldById('contentMediaRadio');`

```Typescript
private getMediaRadioGroup(): RadioGroup {
    return this.getFieldById('contentMediaRadio');
}
```

alansemenov (changes requested):

alansemenov (approved):

PR #6118, opened 2023-03-24
Don't auto-show preview panel for a site #6040

### LINKED ISSUE #6040: Don't auto-show preview panel for a site without preview
For a site, don't auto-expand the preview panel (the "Monitor" icon should also be "off" by default), if there's no preview or no apps are selected or none of the selected apps provide controllers (basically, if no controller can be selected in the preview panel).

Now:
image

Should be:
image

ashklianko:


alansemenov (approved):
Good stuff 👍

PR #6419, opened 2023-06-14
Enable automatic exclusion of non-required dependencies in the Publishing Wizard #6415

### LINKED ISSUE #6415: Enable automatic exclusion of non-required dependencies in the Publishing Wizard
When Publishing Wizard opens, it automatically fetches all dependencies (both mandatory and non-mandatory) of the "main" content:
image

We will add an option to automatically exclude non-required dependencies by introducing a new app config property called `publishingWizard.excludeDependencies`. If this property is present and set to `true` in the app config, PubWiz should still fetch all dependencies, but then exclude all non-required dependencies before rendering the dependency list for enduser. 

There should be no visible blinking (when we first show the full list, then remove dependencies). It should still be possible to manually include the automatically excluded dependencies, if needed.

In the case of the screenshot above, this is how it should look if the new config property is `true`:
image

Note that removal of non-mandatory dependencies and applying changes resulted in removal of the two dependencies that were originally required (because they were parents of the non-mandatory dependencies).

edloidas:


ashklianko (approved):

sgauruseu:
Issue - Excluded item appears in dependencies after reopening the dialog 
Steps - 
1. Open a issue, exclude an item(click on Apply button)
2. Close the dialog, wait for 3 sec
3. Reopen the issue in the dialog 


image

alansemenov (approved):

Issue #6492, opened 2023-07-03
Page Components View: decouple page data from rendered live edit view


ashklianko:
PCV relies on a live edit html to build a structure, so even though descriptor has 2 regions and just one was rendered, then we are now showing just 1 of them that was rendered.
Fix it so that PCV doesn't rely on rendered HTML.

PR #6512, opened 2023-07-05
Don't expand Inspect panel for a selected component with no descripto…

### LINKED ISSUE #6393: Don't expand Inspect panel for a selected component with no descriptor
When a new component (part or layout) is added to a page we instantly open the Inspect panel which will in most cases cover the newly added component. You get one descriptor dropdown inside the component on the page and one inside the Inspect panel, covering each other.

Fix it so that in this case (when no descriptor is selected) we keep the Inspect panel collapsed (or collapse it if it was expanded).

image

ashklianko:
…r #6393


sgauruseu:
**Issue 1:  Inspect Panel closes after saving/autosaving**
Steps:
1. Open a site, insert a  `layout component`
2. Open `Inspect Panel`
3. Select a option in layout dropdown in Inspect Panel

image
 
The site has been saved automatically after selecting a option and `Insect Panel` closes

ashklianko:
Will update after #6492 is done

PR #6686, opened 2023-08-21
Content studio inline preview + nextjs #6390

### LINKED ISSUE #6390: Adding a new component triggers refresh of the editor page
Content studio can refresh the entire page if a new part is inserted into a region. 

We have setup content studio to use the nextjs to render the frontend. 

If you add a new part to a page controller it will refresh the entire page you are on.
This gives a really bad user experience.

ashklianko:


ashklianko:
Tested on local site enonic next setup and asked Persijn to test it also, works good

PR #6820, opened 2023-09-15
Locate target node in the grid after move #6764

### LINKED ISSUE #6764: Locate target node in the grid after move
1. Change notification text after successful move to include path of the target item: "_X items were moved_" -> "_X items were moved to `target/content/path`_".
2. Make the target path a link. If this doesn't work, add an action (`NotificationMessage` has `actionList`).
3. The link should pass id of the target to the special mode in the grid, similar to how we do it for displaying inbound/outbound dependencies of an item.
4. In this mode, grid should find target content by id and expand it (it will always be expandable since child items were moved under it).

ashklianko:


alansemenov (changes requested):
One thing we forgot about. It's now possible to move content from inside the Content Wizard as well. There we DON'T need a link to locate the target node (since we're not in the grid)

ashklianko:
> One thing we forgot about. It's now possible to move content from inside the Content Wizard as well. There we DON'T need a link to locate the target node (since we're not in the grid)

Updated

alansemenov (approved):

PR #6827, opened 2023-09-18
Improve navigation experience in Content Grid

### LINKED ISSUE #6653: Preview panel should not wait for IsRenderable response
When an item is selected in the grid, it seems that preview panel won't start loading until Preview button is activated. These two requests should not depend on each other.

### LINKED ISSUE #6654: Improve navigation experience in Content Grid
We now have a delay on selection of a grid item to make sure we don't send a preview request when item selection changes rapidly (for example, when navigating the grid with arrow buttons). This means that after you selected a new item, it takes a few seconds before preview for the new one starts loading. Combined with https://github.com/enonic/app-contentstudio/issues/6653 (when preview is slow to render), it may result in user looking at preview of previously selected item for quite a long time after a new one has been selected, before the new preview loads.

How it should be:
1. When an item B is selected, we should immediately remove preview of the previously selected item A from the preview panel and
2. immediately show the spinner over the preview panel, but not start actual loading of preview for B yet.
3. When the navigation delay has passed and we know that selection ended up on B, we'll send requests for IsRenderable and preview panel (which should be asynchronous as described in https://github.com/enonic/app-contentstudio/issues/6653).
4. If user navigates away from B while preview is still being loaded, we should follow steps 1-2 - preview for B should not appear if B is no longer selected.

alansemenov:
Needs https://github.com/enonic/lib-admin-ui/pull/3259

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseItemPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowseItemPanel.ts
@@ -18,4 +18,8 @@ export class ContentBrowseItemPanel
     setStatisticsItem(item: ContentSummaryAndCompareStatus): void {
         super.setStatisticsItem(!!item ? item.clone() : item);
     }
+
+    getItemStatisticsPanel(): ContentItemStatisticsPanel {
+        return this.itemStatisticsPanel as ContentItemStatisticsPanel;
+    }
 }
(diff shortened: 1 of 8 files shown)


gradle.properties
```diff
@@ -6,5 +6,5 @@ appProjectName=contentstudio
 libProjectName=lib-contentstudio
 restProjectName=rest
 xpVersion=7.13.0
-libAdminUiVersion=4.8.0-SNAPSHOT
+libAdminUiVersion=4.8.0-3258-SNAPSHOT
```
ashklianko: Need to set the correct version

ashklianko (changes requested):

gradle.properties
alansemenov: Yeah, I will remove this commit.

ashklianko (approved):

PR #6851, opened 2023-09-20
Inspect panel expands when text component is in edit mode #6828

### LINKED ISSUE #6828: Inspect panel expands when text component is in edit mode
While editing a text component, any click inside it will cause the Inspect panel to expand. When the panel is in floating mode and covers the text being edited, it creates a terrible user experience.

Change it so that Inspect panel will expand only if the component was not yet selected on click.

https://github.com/enonic/app-contentstudio/assets/11612490/7aeaf1a1-a6b4-46b0-bdbd-80a7c29a68ea

ashklianko:
- removed static fields from ContextSplitPanel.ts since it doesn't belong there (class is not singleton for example and using other instance will break the logic) 
- hiding context panel if text mode is on and if context panel is in floating mode

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/event/InspectEvent.ts
+++ b/modules/lib/src/main/resources/assets/js/app/event/InspectEvent.ts
@@ -8,8 +8,8 @@ export class InspectEvent
 
     private readonly showPanel: boolean;
 
-    constructor(showWidget: boolean, showPanel: boolean, name?: string) {
-        super(name);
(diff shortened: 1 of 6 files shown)


ashklianko:
@alansemenov, I've changed behavior of showing/hiding inspect panel for empty part/layout: now inspect panel is going to be hidden ONLY if it is in `FLOATING` mode, thus when it is docked it is not covering live view so no need to hide it; Same thing for text component

alansemenov (approved):

PR #6859, opened 2023-09-25
Newly created item is not added to the filtered grid #6818

### LINKED ISSUE #6818: Newly created item is not added to the filtered grid
**Case 1:** 
1. Select a checkbox in Filter Panel (work in progress for example )
2. Open a wizard add  new content, close the wizard

**BUG** - New added content is not displayed in the filtered grid

image

**Case 2:**
1. Add a site , expand the site.
2. Select `Last modified - 1day` checkbox

image
3. Delete `templates` folder
image

**BUG** - Deleted  `Templates` folder remains visible in the grid after deleting 

image

ashklianko:


modules/lib/src/main/resources/assets/js/app/browse/ContentTreeGrid.ts
```diff
@@ -624,24 +625,32 @@ export class ContentTreeGrid
...
+                childrenIds.some((childId: ContentId, index: number) => {
+                    if (contentId.equals(childId)) {
+                        insertPosition = index;
+                        return true;
+                    }
 
-                        return false;
-                    });
+                    return false;
+                });
```
alansemenov: Can be replaced with
```Typescript
const insertPosition: number = childrenIds.findIndex((childId: ContentId) => contentId.equals(childId));
````

alansemenov (changes requested):

alansemenov (approved):

Issue #6069, opened 2023-03-14
DateTime picker opens on any click inside the Schedule Publish dialog


sgauruseu:
1. Select a published content 
2. Click on `Edit Schedule` button in `Details Panel`
3. `Edit Schedule` dialog  is opened - OK 

**BUG** -  click in any place opens Date Time Picker dialog

![image](https://user-images.githubusercontent.com/3728712/225002743-209da537-389f-4ff0-8c32-d39d90db651e.png)

ashklianko:
This is how modal dialog behaves when focus moves out of a dialog - it brings it back to the dialog and focuses first focusable element, in this case it is a date time picker that always shows calendar on focus. @alansemenov not sure if there's anything to do

PR #6886, opened 2023-10-02
DateTime picker opens on any click inside the Schedule Publish dialog…

ashklianko:
… #6069

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
@@ -13,6 +13,9 @@ import {PropertiesWizardStepForm} from './PropertiesWizardStepForm';
 import {UpdateContentRequest} from '../../../../resource/UpdateContentRequest';
 import {ObjectHelper} from '@enonic/lib-admin-ui/ObjectHelper';
 import {InputView} from '@enonic/lib-admin-ui/form/InputView';
+import {Element} from '@enonic/lib-admin-ui/dom/Element';
+import {DateTimeRangePicker} from '@enonic/lib-admin-ui/ui/time/DateTimeRangePicker';
+import {DateTimePicker} from '@enonic/lib-admin-ui/ui/time/DateTimePicker';
 
 export class ScheduleWizardStepForm
     extends PropertiesWizardStepForm {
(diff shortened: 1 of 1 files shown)


ashklianko:
Can't add a corresponding task to the Development section

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
```diff
@@ -66,6 +84,20 @@ export class ScheduleWizardStepForm
         });
     }
 
+    private findFirstChildWithClassIn(parent: Element, className: string): Element {
```
alansemenov: why does the class name end with "`In`"? "`findFirstChildWithClass`" should do just fine.

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
```diff
@@ -58,6 +61,21 @@ export class ScheduleWizardStepForm
                 this.notifyFocused(event);
             });
 
+            this.formView.onRendered(() => {
+                const dateTRP = this.findFirstChildWithClassIn(this.formView, 'date-time-range-picker') as DateTimeRangePicker;
```
alansemenov: `dateTRP` -> `firstDateRangePicker`

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
```diff
@@ -66,6 +84,20 @@ export class ScheduleWizardStepForm
...
+            if (child.hasClass(className)) {
+                return child;
+            } else {
+                let found = this.findFirstChildWithClassIn(child, className);
+                if (found) {
+                    return found;
+                }
+            }
+        }
+        return null;
```
alansemenov: This can be replaced with 
```Typescript
return parent.getChildren().find((child) =>
    child.hasClass(className) || this.findFirstChildWithClass(child, className));
```

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
alansemenov: No, probably not exactly. The second part should be the element returned by `findFirstChildWithClass`

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
ashklianko: Since we are passing the item we search in I added "In" postfix to diverse between inherited element functions that search within current element

modules/lib/src/main/resources/assets/js/app/view/context/widget/details/ScheduleWizardStepForm.ts
ashklianko: Won't work: will return top most parent containing required element, not element itself

alansemenov: Ok, but then let's at least simplify it as this is a classic recursion:
```Typescript
private findFirstChildWithClass(parent: Element, className: string): Element {
    for (const child of parent.getChildren()) {
        if (child.hasClass(className)) {
            return child;
        }

        return this.findFirstChildWithClass(child, className);
    }
    
    return null;
}
```

ashklianko: Won't work also - will look result always only in first child

alansemenov (approved):

PR #6933, opened 2023-10-11
Composite status for moved modified items

### LINKED ISSUE #6912: Composite status for moved modified items
When a published item has been changed, its status will be shown as "_Modified_". If the same item has also been moved, then its status will be shown as "_Moved_" and it's impossible to see that it's also been modified. In this case we should show both statuses separated with a comma: "_Modified, Moved_".

alansemenov:


modules/lib/src/main/resources/assets/js/app/browse/ContentRowFormatter.ts
```diff
@@ -82,16 +82,31 @@ export class ContentRowFormatter {
...
+        });
 
-            return status.toString();
+
+        return statusElements;
+    }
+
+    private static doStatusFormat(data: ContentSummaryAndCompareStatus): string {
+
+        if (data && data.getContentSummary()) {
```
ashklianko: `if (data?.getContentSummary()) {` is enough

ashklianko (approved):

sgauruseu:
Content status is not updated in the Deatils Widget

image

PR #6935, opened 2023-10-11
Status toolbar: fix flickering while saving modified content #6934

### LINKED ISSUE #6934: Status toolbar: fix flickering while saving modified content
When published content has been modified and Save button is pressed then we see that status toolbar changes it's state to PUBLISHED for a moment and switches back to modified after while. Fix it so status doesn't change

ashklianko:


alansemenov (approved):
Nice fix 👍

PR #7052, opened 2023-11-13
Assets of menu-item widget remain in DOM when the widget is no longer…

### LINKED ISSUE #7047: Assets of menu-item widget remain in DOM when the widget is no longer active
1. Install CS+
2. Open Content Studio
3. Switch to the Archive widget. Note that assets (JS/CSS) of the widget were added to the DOM
4. Switch back to the main Content Navigator widget.
 
Bug: Note that the assets of the Archive widget remained in the DOM.

Since menu-item widgets can be created by external developers, this bug makes it very easy to screw up JS/CSS of the Content Studio. It should be the same as for context widgets - their assets should be removed from the DOM when user switches to another widget.

Ideally we should try to isolate widgets better by using Shadow DOM (but then we'll have to switch from stylesheets to inline styles).

ashklianko:
… active #7047

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
+++ b/modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
@@ -20,15 +20,22 @@ import {Router} from './Router';
 import {UrlAction} from './UrlAction';
 import {ContentAppBar} from './bar/ContentAppBar';
 import {WidgetsSidebar} from './widget/WidgetsSidebar';
+import {WidgetInjectionResult} from './util/WidgetInjectionResult';
+
+interface MenuWidgetInjectionResult extends WidgetInjectionResult {
+    isEnabled: boolean;
+}
 
 export class AppWrapper
     extends DivEl {
+
+    private static INTERNAL_WIDGET_PREFIX: string = 'com.enonic.app.contentstudio';
 
     private sidebar: WidgetsSidebar;
 
     private widgets: Widget[] = [];
 
-    private widgetElements: Map<string, Element> = new Map<string, Element>();
+    private widgetElements: Map<string, MenuWidgetInjectionResult> = new Map<string, MenuWidgetInjectionResult>();
 
     private toggleSidebarButton: Button;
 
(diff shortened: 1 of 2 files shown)


ashklianko:
1) Disabling CSS stylesheets from non-internal widgets via using of `disabled` attribute to remove potential CSS interference from custom widgets
2) Doing nothing with JS scripts since they are already loaded and we can't remove them from the memory

@alansemenov if you still would like to remove those html tags totally (though it has no practical sense) then I'll remove them

alansemenov (changes requested):
This won't work. As far as I see, when widget contents get injected into the DOM, its `script` & `link` tags not only get added to the main document's `<head>` but they remain inside the `widget` element as well. When you disable the `link` element inside the `<head>`, the one inside `widget` keeps on working.

The problem is that we don't remove `script` and `link` tags from inside `widget` element when embedding it, which results in them being added to the DOM twice, and never removed.

This is also the case for context panel widgets now, but I'm pretty sure this used to work originally. It seems that this has been changed during the latest changes related to widgets a year ago.

ashklianko:
Updated

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -310,6 +327,36 @@ export class AppWrapper
...
+        widgetData.widgetContainer.show();
+
+        if (!this.isInternalWidget(key) && !widgetData.isEnabled) {
+            widgetData.linkElements.forEach((linkElement: HTMLLinkElement) => {
+                linkElement.disabled = false;
+            });
+
+            widgetData.isEnabled = true;
+        }
+    }
```
alansemenov: This can be replaced with one method accepting a boolean parameter for active/inactive. It should also be renamed since it doesn't only shows/hides but also enables/disables elements.

```Typescript
private setWidgetActive(key: string, widgetData: MenuWidgetInjectionResult, active: boolean): void {
    widgetData.widgetContainer.setVisible(active);

    if (!this.isInternalWidget(key) && widgetData.isEnabled !== active) {
        widgetData.linkElements.forEach((linkElement: HTMLLinkElement) => {
            linkElement.disabled = !active;
        });

        widgetData.isEnabled = active;
    }
}
```

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
ashklianko: Updated

alansemenov (approved):

PR #7086, opened 2023-11-20
Incorrect resolution of dependencies in the Publish Dialog #7081

### LINKED ISSUE #7081: Incorrect resolution of dependencies in the Publish Dialog
1. Resolution of dependencies is slow, we should find out why. Apparently it might even result in the list of dependencies not being rendered at all. We should not try to resolve dependencies for items that are not selected for publishing.
2. Number of dependencies is incorrect. In the attached video there are 2 dependencies loaded but it shows "All (11)" above the list.

https://github.com/enonic/app-contentstudio/assets/11612490/02f16713-f6cb-4765-9216-ecea3e7bab0b

edloidas:
Fixed an unoptimized call to get comparison results via resolvePublishDependencies that resulted in additional requests for each outbound dependency and increased execution time by a factor less than or equal to the number of outbound dependencies.

### DIFF
--- a/modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
+++ b/modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
@@ -708,7 +708,7 @@ public final class ContentResource
             : ContentIds.from( fullPublishList.stream().filter( publishNotAllowedCondition ).collect( Collectors.toList() ) );
 
         //check that not all contents are pending delete
-        final Boolean isAllPendingDelete = getNotPendingDeletion( fullPublishList, compareResults ).getSize() == 0;
+        final Boolean isSomePublishable = fullPublishList.getSize() > 0;
 
         //filter required dependant ids
         final ContentIds requiredDependantIds = ContentIds.from(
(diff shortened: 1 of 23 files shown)


alansemenov:
@vbradnitski please review the [Java part](https://github.com/enonic/app-contentstudio/pull/7086/files?file-filters%5B%5D=.java&show-viewed-files=true)

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
```diff
@@ -794,11 +789,11 @@ private ContentIds problematicDependantsOnTop( final ContentIds dependentContent
     private ContentIds filterOutStatus( final ContentIds contentIds, final CompareContentResults compareResults,
                                         final CompareStatus status )
     {
-        return ContentIds.from( compareResults.stream()
-                                    .filter( result -> result.getCompareStatus() != status )
-                                    .map( CompareContentResult::getContentId )
-                                    .filter( contentIds::contains )
-                                    .collect( Collectors.toList() ) );
+        final ContentIds unorderedIdsWithoutStatus = ContentIds.from( compareResults.stream()
```
vbradnitski: I would recommend not passing `compareResults` to the method, but just call `contentService.compare()` here. Right now it is a bit unclear which contentIds should be in the return list. The answer is " from `contentIds` param " but its hard to get it by reading and requires additional code investigation. And thats the reason why the mistake was done before.

vbradnitski (changes requested):

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
edloidas: Well, not quite. `filterOutStatus()` is just a common method, created to avoid duplication in `getNotPendingDeletion()` and `getNotEqual()`, that are both utility methods, accepting CompareResult, that is different in the context they are called from.
`getNotEqual()` is used in in new relatevely new `getOutboundDependenciesIds()` method, where `contentService.compare()` is passed.

vbradnitski: `PendingDelete` - is deprecated status and the related code can be removed from the CS.
 different context of the calls - is the problem. Passed `CompareResults` contains more data than `contentIds` and its not obvious from the method signature what should be filtered from what.

vbradnitski (approved):

alansemenov (approved):

PR #7096, opened 2023-11-21
Fragment components are not inserted in layout components after DnD i…

### LINKED ISSUE #7095: Fragment components are not inserted in layout components after DnD in PCV
1. Open a site , insert `3-col` and `2-col` layouts
2. Insert a text component in `3-col` left region
3. Save the text component as fragment

image

4. Try to move the fragment using DnD to the left region in `2-col` layout

**BUG**  - Fragment components are not inserted in the 2-col layout components after using DnD

ashklianko:
…n PCV #7095

-Not letting fragment with layout to be moved to be moved to other layout

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsGridDragHandler.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsGridDragHandler.ts
@@ -21,7 +21,6 @@ import {ComponentPath} from '../page/region/ComponentPath';
 export interface PageComponentsHelperFunctions {
     getPathByItem: (item: ComponentsTreeItem) => ComponentPath;
     getItemByPath: (path: ComponentPath) => ComponentsTreeItem;
-    isDropAllowed: (draggedItem: ComponentsTreeItem) => boolean;
 }
 
 export class PageComponentsGridDragHandler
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsGridDragHandler.ts
```diff
@@ -173,6 +164,30 @@ export class PageComponentsGridDragHandler
...
+
+    private isOverLayout(parentTreeItem: ComponentsTreeItem): boolean {
+        const grandParentPath: ComponentPath = this.helperFunctions.getPathByItem(parentTreeItem)?.getParentPath();
+        const grandParent: ComponentsTreeItem = grandParentPath ? this.helperFunctions.getItemByPath(grandParentPath) : null;
+        return grandParent?.getType() instanceof LayoutComponentType;
+    }
+
+    private isFragmentWithLayout(item: ComponentsTreeItem): boolean {
+        return item.getComponent()?.getDescription() === LayoutComponentType.get().getShortName();
+    }
```
alansemenov: This method's logic is not limited to fragments, so it should have a different name (`isComponentWithLayout`?). If we absolutely want this to return `true` only for fragments, then it should have `instanceof FragmentComponentType` check inside.

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsGridDragHandler.ts
ashklianko: Updated

alansemenov (approved):

PR #7136, opened 2023-12-04
Content changes are not saved along with meta data updates #7128

### LINKED ISSUE #7128: Content changes are not saved along with meta data updates
To reproduce:
1. Open a published content
2. Make some changes inside the content form. "Save" button becomes enabled.
3. Open the Details widget
4. Click "Edit schedule" in the Details widget
5. Change Published From date to something else
6. Click Update button
7. Content gets saved, "Save" button becomes disabled.

Bug: Content changes are not saved (can be verified with a page refresh), only the new Online From. Same problem with meta-data ("Edit settings" link above).

ashklianko:
-Problem with considering equal empty value and no value. After form layout is done, some new empty property arrays may be added, and thus persisted item becomes not equal to the viewed one

Fixing it with ignoring empty values until new item is modified

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,6 +5,6 @@ projectName=rootcontentstudio
 appProjectName=contentstudio
 libProjectName=lib-contentstudio
 restProjectName=rest
-xpVersion=7.13.0
+xpVersion=7.13.5-SNAPSHOT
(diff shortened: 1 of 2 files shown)


ashklianko:
https://github.com/enonic/xp/pull/10357 required to remove failing tests

alansemenov (approved):

PR #7158, opened 2023-12-07
[Firefox] Entire top region is selected after DnD#7154

### LINKED ISSUE #7154: [Firefox] Entire top region is selected after DnD
After a component is moved inside a page, the entire top region will be selected on the page. In PCV it's still the moved component that is selected, but inside the Inspect panel (and on the page itself) it's the region that is selected.

This happens only in Firefox :(

https://github.com/enonic/app-contentstudio/assets/11612490/aeadd285-fced-40da-8e8f-bd635e9622af

ashklianko:
-FF generates an extra click event on Layout or Region after drag and drop was made, handling by skipping this event

### DIFF
--- a/modules/lib/src/main/resources/assets/js/page-editor/DragAndDrop.ts
+++ b/modules/lib/src/main/resources/assets/js/page-editor/DragAndDrop.ts
@@ -21,7 +21,6 @@ import {RegionItemType} from './RegionItemType';
 import {ItemView} from './ItemView';
 import {PageView} from './PageView';
 import {LayoutItemType} from './layout/LayoutItemType';
-import {Component} from '../app/page/region/Component';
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/page-editor/DragAndDrop.ts
```diff
@@ -55,6 +54,8 @@ export class DragAndDrop {
 
     private wasDropped: boolean = false;
 
+    private justDropped: boolean = false;
```
alansemenov: Can we rename this to `newlyDropped`?

ashklianko: Done!

alansemenov (approved):

PR #7161, opened 2023-12-07
Content form is auto-collapsed when a component is selected in PCV #7159

### LINKED ISSUE #7159: Content form is auto-collapsed when a component is selected in PCV
We should distinguish between when a component is selected in the Page Editor and inlined PCV. When working with the Page Editor, it's logical to collapse the form to give more room to the Editor. But when user works inside the form (selects a component inside the PCV) we should not auto-collapse it.

ashklianko:
- Tracking source of the select/inspect event, so if it is triggered from Live Edit then don't collapse

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/event/InspectEvent.ts
+++ b/modules/lib/src/main/resources/assets/js/app/event/InspectEvent.ts
@@ -1,5 +1,6 @@
 import {Event} from '@enonic/lib-admin-ui/event/Event';
 import {ClassHelper} from '@enonic/lib-admin-ui/ClassHelper';
+import {PageNavigationEventSource} from '../wizard/PageNavigationEventData';
 
 export class InspectEvent
     extends Event {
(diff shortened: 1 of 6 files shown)


modules/lib/src/main/resources/assets/js/app/event/InspectEvent.ts
```diff
@@ -8,10 +8,14 @@ export class InspectEvent
 
     private readonly showPanel: boolean;
 
-    constructor(showWidget: boolean, showPanel: boolean) {
+    private readonly fromEditor: boolean;
```
alansemenov: Change this to store PageNavigationEventSource instead

```suggestion
    private readonly source: PageNavigationEventSource;
```

modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -940,31 +949,33 @@ export class LiveFormPanel
...
+        const path: ComponentPath = data.getPath();
         this.lastInspectedItemPath = path;
         const currentPage: Page = PageState.getState();
         const item: PageItem = currentPage?.getComponentByPath(path);
+        const isTriggeredFromEditor = data.isFromLiveEditor();
 
         if (item instanceof Component) {
             if (force) { // force inspecting component
-                this.openComponentInspect(item);
+                this.openComponentInspect(item, isTriggeredFromEditor);
```
alansemenov: Pass `data.source` everywhere, instead of `isTriggeredFromEditor`

alansemenov (changes requested):

alansemenov (approved):

PR #7190, opened 2023-12-13
Support multi-inheritance config #7149

### LINKED ISSUE #7149: Support multi-inheritance config
Ref https://github.com/enonic/xp/issues/10351

Content Studio should read this config from XP and switch the Parent Projects dropdown to single-select if multi-inheritance is disabled.

ashklianko:


modules/lib/src/main/resources/assets/js/app/settings/dialog/project/create/data/ProjectParentDialogStepData.ts
```diff
@@ -3,14 +3,18 @@ import {Project} from '../../../../data/project/Project';
...
         return this;
     }
 
-    getParentProject(): Project {
-        return this.parentProject;
+    getParentProjects(): Project[] {
+        return this.parentProjects;
+    }
+
+    getFirstParentProject(): Project {
```
alansemenov: Let's rename this to `getMainParentProject`. In the future we might allow user to select the main parent.

alansemenov (changes requested):
These changes should've been done on the [multi-inheritance branch](https://github.com/enonic/app-contentstudio/tree/issue-7103) which already has support for multiple parents.

PR #7306, opened 2024-01-31
X-data is not shown for fragments #7284

### LINKED ISSUE #7284: X-data is not shown for fragments
Instead of fixing https://github.com/enonic/app-contentstudio/issues/7282 and https://github.com/enonic/app-contentstudio/issues/7281 we will fix it so that x-data is shown for fragments. Those who want to hide it can use regex in site.xml to hide it for fragment content type.

We of course need to verify that https://github.com/enonic/app-contentstudio/issues/7282 no longer reproduces after x-data is shown.

ashklianko:
- Enabling Save button in Inspections panel only if page obj changed
- After saving component (via Page update) have to trigger component inspection (if inspection panel is open) to make component form to be re-rendered with a new config value (config may have been updated on backend with default values etc.)

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -1284,7 +1284,13 @@ export class ContentWizardPanel
     }
 
     private createFragmentSteps(): ContentWizardStep[] {
-        return [this.initPageComponentsWizardStep()];
+        const steps = [this.initPageComponentsWizardStep()];
+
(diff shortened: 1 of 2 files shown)


alansemenov (changes requested):
1. Open a fragment with an x-data
2. Add a text component to the page
3. Change something inside the text. Apply button gets enabled, looking good. Click Apply to save the text. The button becomes disabled again.
4. Now change some fields inside the x-data. Don't save.
5. Now change some thing inside the text component - "Apply" button will remain disabled.

https://github.com/enonic/app-contentstudio/assets/11612490/80080d77-c7bb-463b-bbf9-77a2e3ef75ef

alansemenov (approved):

PR #7311, opened 2024-02-01
Too much indentation in Page Components View

### LINKED ISSUE #7291: Too much indentation in Page Components View
When root node is not collapsible there should be less indentation in the PCV:
image

ashklianko:
#7291

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsTreeGrid.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsTreeGrid.ts
@@ -70,6 +70,15 @@ export class PageComponentsTreeGrid
         PageState.getEvents().onComponentUpdated((event: ComponentUpdatedEvent) => {
             this.updateItemByPath(event.getPath());
         });
+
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/app/wizard/PageComponentsTreeGrid.ts
```diff
@@ -70,6 +70,15 @@ export class PageComponentsTreeGrid
...
         });
+
+        const isRootItemEmptyChecker = this.toggleRootItemHasChildren.bind(this);
+
+        this.onLoaded(isRootItemEmptyChecker);
+        this.onDataChanged(isRootItemEmptyChecker);
+    }
+
+    private toggleRootItemHasChildren(): void {
+        this.toggleClass('empty', !this.isRootItemWithChildren());
```
alansemenov: It works fine, but `empty` name for a CSS class is a) too generic, b) in this case not entirely correct. It should be `root-leaf` or something like that.

alansemenov (changes requested):

alansemenov (approved):

PR #7319, opened 2024-02-02
InspectionsPanel: don't enable 'Apply' button until changes made #7315

### LINKED ISSUE #7315: Disable "Apply" button in the Inspections panel when there are no changes to apply
Currently the button is always enabled, no matter if any changes to the page state present

ashklianko:


ashklianko:
To be checked after https://github.com/enonic/app-contentstudio/pull/7318

alansemenov (approved):
Looking good but we need tests

PR #7399, opened 2024-03-04
Split AiAssistant config #7398

### LINKED ISSUE #7398: Split AiAssistant config
* Setup config with data like service URL must be passsed upon dialog creation, as it is required for other parts of the AiAssistant to work properly.
* Non-critical data, that may be updated, e.g. user name, must be passed via event and updated when dialog opened.

edloidas:


modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardToolbar.ts
```diff
@@ -140,7 +141,12 @@ export class ContentWizardToolbar
 
         const aiAssistantContainer = new DivEl('ai-assistant-container');
         this.addElement(aiAssistantContainer);
-        AI.renderLaunchButton(aiAssistantContainer.getHTMLElement());
+
+        const config: EnonicAiSetupData = {
```
ashklianko: 'config' variable name creates ambiguity with EnonicAiConfigData, could use 'setupData' or similar

ashklianko (approved):

PR #7459, opened 2024-03-21
Move button opening AI Assitant to the right bottom corner #7454

### LINKED ISSUE #7454: Move button opening AI Assitant to the right bottom corner
Should be here:
image

ashklianko:


ashklianko:
Requires https://github.com/enonic/app-saga/pull/160

alansemenov (approved):

edloidas:
IMHO, bottom right action button is actually a cancer of Google material design.
The button came from mobile-first idea, where that button located next to big finger in ugly Google apps. Then it disappeared as quickly as appeared. The location of the button is not "most logical", but rather "most annoying". It takes up more space than it needs and than it could.

I highly recommend returning it to where the rest of the active elements are located between the pages. 
The argument that it will be difficult to find stops working immediately after the first click and is corrected by the initial presentation.

ashklianko:
@edloidas what is your thought on the most suitable place for this button then? Same as it was before?

ashklianko:
Leaving the button as it is until a better spot is found

PR #7484, opened 2024-04-04
Publish of a modified item should not refresh preview #7471

### LINKED ISSUE #7471: Publishing should not refresh preview/editor
If an item is "Modified" (publish fields are already set), publishing it shouldn't refresh the preview panel.

ashklianko:


ashklianko:
Done in https://github.com/enonic/app-contentstudio/pull/7499

PR #7490, opened 2024-04-04
Consistent styling of applications for a site and a project

### LINKED ISSUE #7489: Consistent styling of applications for a site and a project
**Site wizard:**

Right after application is stopped:
image

After page reload:
image

**Project wizard:**
image

Make it consistent:

A stopped app:
* 0.5 opacity
* Normal background
* Red "(!)" icon
* Keep display name
* Description: "_com.app.id is stopped_"

An uninstalled app:
* Normal background
* Red "(!)" icon
* Display name: appId (as now)
* Description: "_com.app.id is not available_"

alansemenov:
Needs https://github.com/enonic/lib-admin-ui/pull/3524

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/ContentFormContext.ts
+++ b/modules/lib/src/main/resources/assets/js/app/ContentFormContext.ts
@@ -9,6 +9,7 @@ import {ContentId} from './content/ContentId';
 import {ContentPath} from './content/ContentPath';
 import {Project} from './settings/data/project/Project';
 import {ApplicationKey} from '@enonic/lib-admin-ui/application/ApplicationKey';
+import {Application} from '@enonic/lib-admin-ui/application/Application';
 
 export class ContentFormContext
     extends FormContext {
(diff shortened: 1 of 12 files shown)


modules/lib/src/main/resources/assets/js/app/ContentFormContext.ts
```diff
@@ -74,6 +77,15 @@ export class ContentFormContext
         return this.applicationKey;
     }
 
+    setStoppedApplications(applications: Application[]): ContentFormContext {
+        this.stoppedApplications = applications;
+        return this;
+    }
+
+    getStoppedApplicationByKey(appKey: ApplicationKey): Application {
+        return this.stoppedApplications.find(app => app.getApplicationKey().equals(appKey));
```
ashklianko: `return this.stoppedApplications?.find(app => app.getApplicationKey().equals(appKey));`

`stoppedApplications` is optional, need to rather use question mark or to init value with an empty array (then `setStoppedApplications` also needs to be updated:
`this.stoppedApplications = applications || [];`
)

ashklianko (changes requested):

ashklianko (approved):

PR #7499, opened 2024-04-09
Don't reload content preview/live edit if no data changed #7485

### LINKED ISSUE #7485: "Mark as ready" should not refresh preview/editor
We should also not refresh preview OR editor if only `workflow.state` and `modifiedTime` have been changed. This is a "mark as ready" operation which cannot affect preview in any way, so no need to refresh. This could be fixed in the same task.

image

_Originally posted by @alansemenov in https://github.com/enonic/app-contentstudio/issues/7471#issuecomment-2034003820_

ashklianko:
Both #7471 and #7485

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/Diff.ts
+++ b/modules/lib/src/main/resources/assets/js/app/Diff.ts
@@ -1,0 +1,15 @@
+export interface Diff {
+    [key: string]: boolean | Diff;
+}
+
+export const isEqual = (diff: Diff) => {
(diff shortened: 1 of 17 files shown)


ashklianko:
NB: Made content preview to not reload on certain conditions, however the content object available for diff check is of type `ContentSummary` which means that we don't have `Page`,` Data`, `ExtraData` info to detect what triggered update, unlike if it was 
 `Content` type

ashklianko:
Tests error is not reproduced locally, I am going to recreate a pull to recheck if the error is present

PR #7503, opened 2024-04-10
Optional X-Data header: prevent jumping on enable/disable form #7502

### LINKED ISSUE #7502: Optional X-Data header: prevent jumping on enable/disable form
X-Data form still jumps when opening/closing optional x-data form, styles are located in lib-admin, fixing it here

ashklianko:


ashklianko:
NB: gray area around disabled x-data header will be a little smaller; if that is fine then the problem is fixed

alansemenov (approved):

PR #7504, opened 2024-04-10
Issue 7485 upd

### LINKED ISSUE #7471: Publishing should not refresh preview/editor

### LINKED ISSUE #7485: "Mark as ready" should not refresh preview/editor

ashklianko:


alansemenov (approved):
Great stuff. Makes a big difference, too.

PR #7510, opened 2024-04-12
Controller returning null breaks part/layout placeholder #7497

### LINKED ISSUE #7497: Controller returning null breaks part/layout placeholder
1. Create a part with a controller that returns `null`
2. Add a part to a page and select the controller from the previous step in the part dropdown

Bug: The result will be a part placeholder which still looks like no controller is selected, although it is and is correctly shown in the Inspect panel on the right. There's the same problem with layouts.

image

We should fix it so that:
1. Controller dropdown is not shown
2. A text is shown over the placeholder saying that the part (or layout) is empty

Suggestion:
image

ashklianko:


alansemenov (changes requested):
Right after descriptor is selected, placeholder shows "`Part "Part" is empty`", even though descriptor's display name is already resolved. After page refresh it shows the display name correctly.

image

ashklianko:
> Right after descriptor is selected, placeholder shows "`Part "Part" is empty`", even though descriptor's display name is already resolved. After page refresh it shows the display name correctly.
> 
> image

Updated!

alansemenov (approved):

PR #7526, opened 2024-04-17
Duplicate message about failed rendering in the editor panel #7482

### LINKED ISSUE #7482: Duplicate message about failed rendering in the editor panel
1. Create a site which is rendered by some controller from an app (f.ex. Superhero)
2. Stop the application the controller is in

Duplicate error message in the editor panel:

image

ashklianko:


modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -2736,10 +2742,13 @@ export class ContentWizardPanel
...
 
     private updateMissingOrStoppedApplications() {
+        const before = this.missingOrStoppedAppKeys?.length ?? 0;
+
         this.fetchMissingOrStoppedAppKeys().then((missingApps: ApplicationKey[]) => {
             this.missingOrStoppedAppKeys = missingApps;
+            const after = missingApps.length;
 
-            if (missingApps.length > 0) {
+            if (before !== after) { // can also sort arrays and check array equality
```
alansemenov: This is too philosophical. Can we rename this to `currentAppCount` and `fetchedAppCount`?

alansemenov (changes requested):
The problem now is that it stopped warning about missing apps at all. It says "check the logs for errors" but there are no errors in the logs.

The logic should be like this:

1. If HEAD returns 418 AND there are missing apps, then the message about missing apps should have higher priority than other messages.
2. If HEAD returns 200 AND there are missing apps, then we don't need to care about missing apps and show the message about them.
3. If HEAD returns 418 AND there are no missing apps, then show relevant message (like now).

image

modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -987,4 +1009,29 @@ export class LiveFormPanel
...
+                    this.frameContainer?.hide();
+                    this.placeholder.setPageIsNotRenderableMode();
+                    this.placeholder.show();
+                }
+            } else { // no page
+                this.clearErrorMissingApps();
+                this.placeholder?.setHasControllersMode(this.hasAvailableControllers);
+            }
+        }
+    }
```
alansemenov:
```suggestion
        if (this.isRenderable) {
            this.clearErrorMissingApps();
            return;
        }
        
        if (!this.hasPage) {  // no page
            this.clearErrorMissingApps();
            this.placeholder?.setHasControllersMode(this.hasAvailableControllers);
            return;
        }

        if (this.hasMissingApps) { // not renderable due to missing apps
          this.setErrorMissingApps();
          return;
        }
                
        // not renderable due to rendering error
        this.clearErrorMissingApps();

        if (!this.placeholder) {
            this.createLiveEditPagePlaceholder();
        }

        this.frameContainer?.hide();
        this.placeholder.setPageIsNotRenderableMode();
        this.placeholder.show();
    }
```

alansemenov: Can probably be refactored as well, since `this.clearErrorMissingApps()` is called in 3 out of 4 cases, so there's no point in repeating it.

alansemenov (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
alansemenov: Functionality now works fine, btw. Great job

ashklianko: Updated

alansemenov (approved):

PR #7544, opened 2024-04-24
Enable marking image as decorative in the Insert Image dialog #7542

### LINKED ISSUE #7542: Enable marking image as decorative in the Insert Image dialog
https://www.w3.org/WAI/tutorials/images/decorative/

Replace "Alternative text" input with a radio group with "Accessibility" label and two radio buttons:
* Decorative image
* Alternative text, with a text input underneath

One of the two options must be selected to insert an image. If "Alternative text" is selected, user must fill in value in the text input (make sure to trim the value before validating).

If "Decorative image" is selected, add "`alt`" attribute with an empty value (`alt=""`) to the `img` element.
If "Alternative text" is selected, add "`alt`" attribute with the value from the text input (make sure to trim).

For an existing image presence (or absence) and value of "`alt`" attribute will determine which of the radio buttons will be preselected.

Remember to localise all the new texts.

image

ashklianko:


alansemenov (changes requested):
Almost there. We shouldn't automatically preselect "Alt text" radio button if the image binary has alt text, user still has to explicitly make a choice. Alt text should be preset in the input, but the radio button should not be preselected.

image

ashklianko:
> Almost there. We shouldn't automatically preselect "Alt text" radio button if the image binary has alt text, user still has to explicitly make a choice. Alt text should be preset in the input, but the radio button should not be preselected.
> 
> image

Updated

alansemenov (approved):

PR #7546, opened 2024-04-28
Controller-dropdown should not be displayed in Fragment wizard #7543

### LINKED ISSUE #7543: Controller-dropdown should not be displayed  in Fragment wizard
1. Open `Features` site
2. `Save as Fragment` the existing text component  
3. Switch to the fragmnet wizard and clck on `Hide Page Editor` toggler
4. Click on the `Show Context Panel` toggler

**BUG** - Controller-dropdown should not be displayed  in `Context Panel`

image

ashklianko:
- Adding inspection panels and inseret panel depending on type of page item; text and part fragments for example only going to have one available inspection panel and no insert panel at all


alansemenov (changes requested):
We need to preselect layout/part/text of the fragment when nothing is selected in the Page Component view of a fragment. If it's selected and then unselected, then controller stays correctly selected in the Inspect panel.  

image

ashklianko:
Updated!

alansemenov (approved):

PR #7552, opened 2024-04-30
Improved styling of regions inside Page Components view

### LINKED ISSUE #7290: Improved styling of regions inside Page Components view
Regions should be styled differently than other nodes in the Page Components view: 
* Remove subtitle, keep only display name
* Change color
* Make a double-click expand/collapse (any) node

image

alansemenov:


ashklianko:
Couple issues I see in this updated design:

1. Region name is not aligned with expand/collapse arrow, an eye catches it immediately:
image

2. Empty regions look alien, while page components have icon and subtitle it only has display name, nothinbg tells what it is:
image

ashklianko (changes requested):
Need to discuss the look of the updated region

alansemenov:
Alignment is fixed
image

alansemenov:
Updated

image

PR #7666, opened 2024-06-18
Update widgets url generation for XP8 #7665

### LINKED ISSUE #7665: Update widgets url generation for XP8
After latest changes in XP8 widgets in CS are no longer located under `/admin/site/.../_/widgets`.

To fix it use `/admin/tool/_/widgets/...`.

Issue in XP https://github.com/enonic/xp/issues/10596

anatol-sialitski:


modules/lib/src/main/resources/assets/js/app/rendering/UriHelper.ts
```diff
@@ -40,4 +40,8 @@ export class UriHelper {
         return adminUrl + (adminUrl.charAt(adminUrl.length - 1) === Path.DEFAULT_ELEMENT_DIVIDER ? '' : Path.DEFAULT_ELEMENT_DIVIDER) +
                baseUrl;
     }
+
+    public static buildWidgetUri(widgetUrl: string): string {
+        return `/admin/tool/${widgetUrl}`;
+    }
```
alansemenov: This will not work when admin is mounted on root via vhost config

```
mapping.admin.host = localhost
mapping.admin.source = /
mapping.admin.target = /admin
mapping.admin.idProvider.system = default
```

alansemenov (changes requested):

alansemenov (approved):

PR #7679, opened 2024-06-20
Fragment: inspection panel doesn't trigger update events #7676

### LINKED ISSUE #7676: Fragment: inspection panel doesn't trigger update events
https://enonic.zendesk.com/agent/tickets/7536

Open a fragment, the one that comprises a part with a configuration. Upon initiating the fragment wizard, the inspection panel for the part opens immediately. Interaction with the part's configuration in the inspection panel does not update the content state, and the Save button remains disabled. However, selecting the part in the PageComponentsView (which consists of only one element—the part itself) resolves the issue, allowing the interaction to function correctly.

In version 5.1.4, the inspection panel did not open immediately upon opening the content wizard, necessitating the selection of a fragment component first.

ashklianko:
- Fix: Using page state component to set into the inspection panel, thus events start firing
- Fix: Context window should inspect non-null object

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
@@ -724,26 +724,25 @@ export class LiveFormPanel
             this.inspectPage({showPanel: false, showWidget: true});
         });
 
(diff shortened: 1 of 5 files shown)


alansemenov (approved):
Works fine

PR #7712, opened 2024-07-03
Animate text input that is being translated #7670

### LINKED ISSUE #7670: Animate text input that is being translated
Text inputs (TextLine, TextArea, HtmlArea) should have a new state which will show that their value is being generated (fetched etc.):
* the input should be disabled (+ some kind of a mask/different background to show that the input is locked)
* Saga icon should be animated (we can start with having a blue circle circling around the Saga icon, this will most likely change later on)
* a tooltip over the locked input saying "`<label> is being translated...`"

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3634

edloidas (approved):

PR #7940, opened 2024-10-09
Incorrect behaviour inside selector inputs #7938

### LINKED ISSUE #7938: Incorrect behaviour inside selector inputs
Problems:
1. All selectors now have "OK" button that has to be clicked to confirm selection
2. "OK" button replaces the dropdown handle button, so it's impossible to close the dropdown unless you click outside
3. Broken styles (height is too high) inside the dropdown

image

How it should be (the way it was before):
1. In single-select dropdown a click on an option should instantly select it.
2. In multi-select dropdowns "Apply" button should appear to the left of the dropdown handler, once smth is selected
3. Styles should be correct 

image

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3748

alansemenov (approved):

PR #7943, opened 2024-10-10
PCV Item doesn't get selection after clicking on the Edit menu item #…

### LINKED ISSUE #7930: PCV Item  doesn't get selection after clicking on the Edit menu item
**Case 1** 

1. Add 3 text componsnts in a site
2. Select the third item in PCV
3. Expand the menu in the first item and click on `Edit` 

image

**BUG** -  The first text component item should be automatically selected in PCV and  switched to `edit-mode`

**Actually** - no items are selected in the PCV after clickin on `Edit`. 

image

**Case 2** 
**The same issue with Inspect menu item** 
 The part is selected in `PCV` , but  `layout-item` has been inspected

image

ashklianko:
…7930

### DIFF
--- a/modules/lib/src/main/resources/assets/js/page-editor/ComponentView.ts
+++ b/modules/lib/src/main/resources/assets/js/page-editor/ComponentView.ts
@@ -226,11 +226,17 @@ export class ComponentView
     }
 
     select(config?: ItemViewSelectedEventConfig, menuPosition?: ItemViewContextMenuPosition) {
-        Element.fromHtmlElement(window.frameElement as HTMLElement).giveFocus();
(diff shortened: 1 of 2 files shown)


alansemenov (changes requested):
Tested the fix. Case 2 is not completely fixed. It does correctly change the Inspect panel, but it should also shift focus from the previous component to the new one, both in view and edit mode. It should basically change the selected component to the new one (as Edit does). 

Also remove `console.log`

ashklianko:
Then we don't need Inspect event if does the same thing as Select.

alansemenov (approved):

PR #7988, opened 2024-10-23
fix for local dev build

ashklianko:
SWC's latest versions started processing src/output dirs differently, need to update build script accordingly


alansemenov (approved):
Awesomeness

PR #8056, opened 2024-11-13
Live view

### LINKED ISSUE #8053: Merge live view with master

pmi:
Depends on Widget config missing fields
 https://github.com/enonic/lib-admin-ui/issues/3765


modules/lib/src/main/resources/assets/js/app/view/toolbar/EmulatorDropdown.ts
```diff
@@ -0,0 +1,151 @@
...
+    constructor() {
+        super(new EmulatorListBox(), {
+            className: 'emulator-dropdown preview-toolbar-dropdown',
+            maxSelected: 1,
+            filter: (item: EmulatorDevice, searchString: string): boolean => {
+                return item.getName().toLowerCase().indexOf(searchString.toLowerCase()) > -1 ||
+                       item.getDeviceType().toString().indexOf(searchString.toLowerCase()) > -1 ||
+                       item.getWidth().toString().indexOf(searchString.toLowerCase()) > -1 ||
+                       item.getHeight().toString().indexOf(searchString.toLowerCase()) > -1;
+            }
```
alansemenov: This dropdown doesn't have a filtering input, so we don't need `filter` config. If for some reason we do then we shouldn't repeat `searchString.toLowerCase()` 4 times.

modules/lib/src/main/resources/assets/js/app/view/toolbar/PreviewWidgetDropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+    private selectedOption: PreviewWidgetOptionViewer;
+
+    constructor() {
+        super(new WidgetSelectorListBox(), {
+            className: 'preview-toolbar-dropdown',
+            maxSelected: 1,
+            filter: (item: Widget, searchString: string): boolean => {
+                return item.getWidgetDescriptorKey().getName().toLowerCase().indexOf(searchString.toLowerCase()) > -1 ||
+                       item.getDescription().toLowerCase().indexOf(searchString.toLowerCase()) > -1;
+            }
```
alansemenov: Same here - no need for `filter` since the dropdown is not filterable.

modules/app/src/main/resources/admin/widgets/preview-json/preview-json.xml
```diff
@@ -0,0 +1,11 @@
+<widget xmlns="urn:enonic:xp:model:1.0">
+  <display-name>JSON</display-name>
+  <description>Display in JSON format</description>
```
alansemenov: All of the preview widgets must have localised `display-name` and `description` (via i18n attribute). It's enough to add the localisation keys only to `phrases.properties`.

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -216,106 +244,49 @@ export class ContentItemPreviewPanel
...
-                    setTimeout(() => {
-                        this.item = null; // we don't have ref to content under contentPreviewPath and there is no point in figuring it out
-                        this.skipNextSetItemCall = false;
-                        this.frame.setSrc(clickedLinkRelativePath);
-                    }, 500);
+                    // setTimeout(() => {
+                    //     this.item = null; // we don't have ref to content under contentPreviewPath and there is no point in figuring it out
+                    //     this.skipNextSetItemCall = false;
+                    //     this.fetchPreviewForPath(clickedLinkRelativePath);
+                    // }, 500);
```
alansemenov: Remove the comments

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -154,38 +164,56 @@ export class ContentItemPreviewPanel
...
+            switch (this.frame.getClass()) {
+            case 'image':
+                this.applyImageStyles(frameWindow);
+                break;
+            case 'text':
+                try {
+                    frameWindow.addEventListener('click', (event) => this.frameClickHandler(frameWindow, event));
+                } catch (error) { /* error */ }
+                break;
+            }
```
alansemenov: Should have default fallback

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -216,106 +244,49 @@ export class ContentItemPreviewPanel
...
+        case PREVIEW_TYPE.WIDGET: {
+            this.getEl().addClass('widget-preview');
+            break;
+        }
+        case PREVIEW_TYPE.EMPTY: {
+            this.showPreviewMessages(messages || [i18n('field.preview.notAvailable')]);
+            break;
+        }
+        case PREVIEW_TYPE.FAILED:
+        case PREVIEW_TYPE.MISSING: {
```
alansemenov, line 273: Maybe replace with `default:`?

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -441,4 +366,40 @@ export class ContentItemPreviewPanel
...
+        if (img) {
+            img.style.margin = '0 auto';
+            img.style.height = '100%';
+        } else {
+            img = frameWindow.document.querySelector('body > img');
+        }
+        if (img) {
+            img.style.maxWidth = '100%';
+            img.style.maxHeight = '100%';
+        }
```
alansemenov, line 398: Can't we set this via CSS? If not, then we need to do `new ElementHelper(body)`, then use setters from ElementHelper (`setDisplay`, `setMaxWidth` etc)

alansemenov (changes requested):
We need to have a bit more space both around and inside the dropdowns, it feels too crampled now. Also, the icons inside the toolbar are a bit too small - they should be of the same size as the preview icon.

Now (14x14):
image

Should be (20x20):
image

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
pmi: No, it can't be done with css because we modify contents of iframe that is populated by the browser.
`ElementHelper` doesn't work either unfortunately, because it doesn't have access to fields like `justifyContent` or `alignItems`

alansemenov (approved):
Great job, Pavel!

PR #8059, opened 2024-11-13
Update AI helper config #8058

### LINKED ISSUE #8058: Update AI helper config
Make AiHelper use new config format.

edloidas:
Updated AiHelper to use new format.
Added null check for after expression match in DisplayName, as it may return null in some cases, which resulted in console errors on tabbing through page.


edloidas:
Requires https://github.com/enonic/lib-admin-ui/pull/3786

ashklianko (approved):

PR #8102, opened 2024-11-27
Juke icon for inputs should be togglable #8089

### LINKED ISSUE #8089: Juke icon for inputs should be togglable
By default, the icon should be in "inactive" state. Use the following properties in css:
```css
filter: grayscale(0.50);
opacity: 0.5;
```

In the "active" state `filter` should be removed and `opacity` should become 1 (the icon will look as now).

The icon will be in "active" state on hover and when the input is set inside the Operator dialog's context.

Make sure there's no blinking (the icon remains in the active state) between when the icon is clicked and when the input is set in the context.

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3803 and https://github.com/enonic/app-ai-content-operator/pull/150

edloidas (approved):

PR #8105, opened 2024-11-28
Fix highlight and scroll to the input from AI Dialog #8101

### LINKED ISSUE #8101: Fix highlight and scroll to the input from AI Dialog
Highlight and scroll to the element in AI dialog stopped working after removing 'data-path' attribute, need to implement a new approach for this behavior.

- Send corresponding events to Studio and do highlighting/scrolling from there.
- Item-set and option set option should also work

ashklianko:


ashklianko:
Requires https://github.com/enonic/app-ai-content-operator/pull/153

modules/lib/src/main/resources/assets/js/app/ai/ui/InteractionType.ts
```diff
@@ -0,0 +1 @@
+export type InteractionType = 'click';
```
edloidas: Not sure that it deserves to be in a separate file (even more so under `/ui`). We won't probably use it as import, so it can be just an inner type in AiContentOperatorInteractionEvent

modules/lib/src/main/resources/assets/js/app/ai/AI.ts
```diff
@@ -238,6 +241,26 @@ export class AI {
         }
     };
 
+    private handleInteractionEvent = (event: AiContentOperatorInteractionEvent) => {
+        if (event.interaction === 'click') {
```
edloidas: It's better to have switch here, because that's what you do, right? If you have switch without the `default` clause, it will show error, if you add new interaction, but forget to add the handler. So it's kinda handy, IMHO.

edloidas (changes requested):

edloidas (approved):

PR #8120, opened 2024-12-03
Show Juke icons for parent items of an active input #8106

### LINKED ISSUE #8106: Show Juke icons for parent items of an active input
NB: Requires  [#8089](https://github.com/enonic/app-contentstudio/issues/8089)

Inputs can be located within item sets or option set options. When the context is set, users should be able to navigate up through context levels not only via the AI dialog but also directly from the content form.

So if the context is set:

- Show Juke icon not only for currently active/context input but also show a greyed out icon of Juke for every item up in the hierarchy.
- Add AIHelper with icons for item-sets and option-set options (put icon near action menu)

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3812

alansemenov (approved):

edloidas (approved):

PR #8126, opened 2024-12-03
Live View Panel is not updated after updating its content in wizard #…

### LINKED ISSUE #8082: Live View Panel is not updated after updating its content in wizard
1. Select Features site , click on Edit button 
2. Change the text in `main region` or insert an image
3. Switch to Browse Panel again

**BUG**  -  Live View Panel is not updated after updating the content in wizard

image

### LINKED ISSUE #8117: Preview button is disabled after selecting a site in the second time
1. Select a site with a controller(Features)  - Preview button is enabled - OK
2. Click on the same row and unhighlit the site
3. Click on the row again .

**BUG**  - `Preview` button is disabled : 

image

pmi:
…8082

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
@@ -36,7 +36,7 @@ export class ContentItemPreviewPanel
     protected debouncedSetItem: (item: ViewItem) => void;
     protected readonly contentRootPath: string;
     private previewHelper: PreviewActionHelper;
-    private renderable: Promise<boolean>;
+    private renderable: Q.Promise<boolean>;
 
     constructor(contentRootPath?: string) {
         super('content-item-preview-panel');
@@ -111,30 +111,13 @@ export class ContentItemPreviewPanel
         this.fetchPreviewForPath(contentSummary);
     }
 
-    private defer<T>() {
-        const deferred: {
-            promise: Promise<T>,
-            resolve: (value: T | PromiseLike<T>) => void,
-            reject: (reason?: unknown) => void,
-        } = {
-            promise: null,
-            resolve: null,
-            reject: null,
-        };
-        deferred.promise = new Promise<T>((resolve, reject) => {
-            deferred.resolve = resolve;
-            deferred.reject = reject;
-        })
-        return deferred;
-    }
-
-    public isRenderable(): Promise<boolean> {
+    public isRenderable(): Q.Promise<boolean> {
         return this.renderable;
     }
 
     private async fetchPreviewForPath(summary: ContentSummary): Promise<void> {
 
-        const deferred = this.defer<boolean>();
+        const deferred = Q.defer<boolean>();
         this.renderable = deferred.promise;
 
         const previewWidget = (this.toolbar as ContentItemPreviewToolbar).getWidgetSelector().getSelectedWidget();
--- a/modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
@@ -36,7 +36,7 @@ export class ContentItemPreviewPanel
     protected debouncedSetItem: (item: ViewItem) => void;
     protected readonly contentRootPath: string;
     private previewHelper: PreviewActionHelper;
-    private renderable: Q.Promise<boolean>;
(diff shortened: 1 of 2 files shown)


ashklianko (changes requested):
- Let's stick with Q.Promises for now and switch entire code to the ES6 promises at some point
- You can use Q.Deferred for now, implementing your own defer function looks like overkill here and like a code that doesn't belong here

Need to mention that using Deferred is considered a bad practice , but here is how it should be:

```
const deferred = Q.defer<boolean>()
        this.renderable = deferred.promise;
```

pmi:
Using deprecated Q.Promise is also a bad practice :) 

I know that Deferred is not advised, but it is needed here because the renderable promise is determined by the result of the fetch promise. 
At the same it can't be simple boolean, because its value is unknown until fetch is finished.

ashklianko:
Yes, so let's stick to Q.defer for now to eliminate redundant custom implementation of deferred

ashklianko (approved):

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -110,7 +111,15 @@ export class ContentItemPreviewPanel
         this.fetchPreviewForPath(contentSummary);
     }
 
+    public isRenderable(): Q.Promise<boolean> {
```
alansemenov: This method should be renamed. What is renderable here, entire ContentItemPreviewPanel? More fitting name is for example `isPreviewRendered`. 

Also why is it public?

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -76,19 +77,19 @@ export class ContentItemPreviewPanel
...
         }
 
         this.toolbar.setItem(item);
         this.item = item;
     }
 
-    public isPreviewUpdateNeeded(item: ContentSummaryAndCompareStatus, force?: boolean): boolean {
-        return !this.skipNextSetItemCall && this.isItemAllowsUpdate(item, force);
+    public async isPreviewUpdateNeeded(item: ContentSummaryAndCompareStatus, force?: boolean): Promise<boolean> {
+        return !this.skipNextSetItemCall && (this.isItemAllowsUpdate(item, force) || this.isRenderable());
```
alansemenov: Values of different types should not be combined in one logical expression. The method used to return `boolean && boolean`, now you changed it to return `Promise<boolean>` but the condition inside is `boolean && (boolean ||  Promise<boolean>)`.

First of all, Promise should be replaced with Q.Promise as we are not on native promises yet.
Then the condition should be split in several so that the method always returns a promise, not a boolean.

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -36,6 +36,7 @@ export class ContentItemPreviewPanel
     protected debouncedSetItem: (item: ViewItem) => void;
     protected readonly contentRootPath: string;
     private previewHelper: PreviewActionHelper;
+    private renderable: Q.Promise<boolean>;
```
alansemenov: This property should also be renamed.

alansemenov (changes requested):

ashklianko:
One more thing to discuss about using `await` on this level: do we really need blocking main loop until we get a response? We use though async-await on the top-most levels where it is required but we should avoid it on this level and stick to promises.

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
pmi: This tells if currently displayed item is renderable because is renderable request is now dropped, will rename to isItemRenderable.
Public in case other components want to know if item is renderable.

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
pmi: Logical operations work correctly with native promises returning a promise if one of the operands is one.

But I am not sure about Q.Promise, so will change it since we're going back to Q.

pmi: Which I am not a fan of btw, because we have plans to drop it

alansemenov (approved):

PR #8131, opened 2024-12-04
AI: validation is not triggered after insert into HtmlArea #8130

### LINKED ISSUE #8130: AI: validation is not triggered after insert into HtmlArea

ashklianko:


alansemenov (approved):
the fix helped

PR #8140, opened 2024-12-06
Show translation status for component config inputs #8092

### LINKED ISSUE #8092: Show translation status for component config inputs
- Items that are being translated in component config and shown in inspection panel must have a spinner as the fields in main content panel. Same for the error status
- Inputs in component inspection panel must not have Juke icon that opens AI dialog

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3814

modules/lib/src/main/resources/assets/js/app/ai/AI.ts
```diff
@@ -384,14 +400,28 @@ export class AI {
...
+            (helper: AiHelper) => this.isHelperForXData(helper, xData)) : null;
+    }
+
+    private getAiHelperByPage(path: string): AiHelper | undefined {
+        if (path.indexOf(AI.CONFIG_PREFIX) > -1) {
+            const dataPath = path.replace(AI.PAGE_PREFIX, '').split(`/${AI.CONFIG_PREFIX}`)[1];
+            return AiHelper.getAiHelpersByGroup('page').find((helper: AiHelper) => helper.getDataPath() === dataPath);
+        }
+
+        return null;
```
edloidas: Our TS config is not strict and won't actually detect if this function actually returns undefined. But, when we finally make it strict, `undefined` and `null` are not of the same type and result in error. Let it just work as "find" and return `undefined`, like you assumed in return value. 
```suggestion
    private getAiHelperByPage(path: string): AiHelper | undefined {
        if (path.indexOf(AI.CONFIG_PREFIX) > -1) {
            const dataPath = path.replace(AI.PAGE_PREFIX, '').split(`/${AI.CONFIG_PREFIX}`)[1];
            return AiHelper.getAiHelpersByGroup('page').find((helper: AiHelper) => helper.getDataPath() === dataPath);
        }

        return undefined;
```

modules/lib/src/main/resources/assets/js/app/inputtype/text/HtmlArea.ts
```diff
@@ -622,8 +623,11 @@ export class HtmlArea
         super.updateInputOccurrenceElement(textAreaEl, property, unchangedOnly);
     }
 
-    isAiEditable(): boolean {
-        return true;
+    getAiConfig(): AiConfig {
+        return {
+            group: this.getContext().formContext?.getName(),
+            aiTools: this.getContext().formContext?.getAiTools() ?? [],
```
edloidas: Don't forget moving tools from Array to Set, if you'll make corresponding changes to libAdminUI.

edloidas (approved):

PR #8200, opened 2025-01-09
Refresh page if components with page contributions are added #2034

### LINKED ISSUE #2034: Refresh page if components with page contributions are added
XP and Content Studio ignores PageContributions when components are inserted into the Page Editor. This means that react components will lack any styling the first time, until you reload the Page Editor (or add components that already are within the Page Editor before it loaded the first time). Enonic's planned solution for this is to just reload the Page Editor whenever a component uses PageContributions, a rather blunt solution.

We could instead check for `req.mode == 'edit'` and then instead of adding our things as PageContribution we'll inline scripts and styles.

This would require the PR enonic/lib-react4xp#32 to be merged first. Gjensidige has a working example of this running, will tweak it and send a PR.

ashklianko:


ashklianko:
Requires https://github.com/enonic/xp/pull/10852

modules/lib/src/main/resources/assets/js/page-editor/LiveEditPage.ts
```diff
@@ -515,66 +518,73 @@ export class LiveEditPage {
...
-
-                if (parentItemView instanceof RegionView) { // PageView for a fragment|
-                    newComponentView.getParentItemView().registerComponentViewListeners(newComponentView);
+        return fetch(componentUrl)
+            .then(response => {
+                const hasContributions = response.headers.has('X-Has-Contributions');
+
+                // reloading entire live page if the component has contributions and there are no equal components on the page
+                if (hasContributions && !this.hasSameComponentOnPage(componentView.getPath())) {
+                    NotifyManager.get().showFeedback(i18n('live.view.component.has.contributions'));
```
alansemenov: We don't need to show the notification. Editors should not/will not care about page contributions and it's too much text (in addition to notification about saved changes) which they will not have time to read. Remove the new i18n key as well.

alansemenov (changes requested):

alansemenov (approved):
Good stuff!

PR #8219, opened 2025-01-14
Image editor upload button has no aria-label #8215

### LINKED ISSUE #8215: Image editor upload button has no aria-label
image

skjulestad:


modules/lib/src/main/resources/i18n/phrases.properties
```diff
@@ -235,6 +235,7 @@ field.rich.text = Rich Text Editor
 field.rich.text.help = Write text and embed images
 field.fragment=Fragment
 field.fragment.help=Reusable components
+field.image.help = Upload or use existing image
```
alansemenov: You cannot select an existing image with the upload button.

alansemenov:
```suggestion
field.image.help = Upload an image file
```

alansemenov: We also need to rename the key. The way it is now is too generic, and it doesn't have anything to do with field either.

We have a separate section of keys that start with "tooltip", so I suggest to add the new key there and call it "`tooltip.button.uploadImage`".

alansemenov:
```suggestion
tooltip.button.uploadImage = Upload an image file
```

alansemenov (approved):
Nice work, Mathias!

PR #8224, opened 2025-01-16
Change status from "Publishing scheduled" to "Scheduled" #8169

### LINKED ISSUE #8169: Change status from "Publishing scheduled" to "Scheduled"
The current text is too verbose and breaks UI

Skjermbilde 2024-12-19 kl  10 08 35

NOTE: Also look at updating translations in a similar way

skjulestad:
Updating from Publishing Scheduled/Expired to Scheduled/Expired
Old:

Image

Image

New:

Image

Image


alansemenov (changes requested):
As you can see, we already have `status.scheduled` and `status.expired` in the same file, which have the texts we need and which are also translated to other languages.

So what we need to do here is:
a) Replace `status.published.expired` and `status.published.scheduled` with `status.expired` and `status.scheduled` respectively everywhere in the code. As far as I can see, they are only used in `ContentSummaryAndCompareStatus`.

b) delete `status.published.expired` and `status.published.scheduled` from `phrases.properties`

alansemenov (approved):

PR #8243, opened 2025-01-21
Refactor live view to support archive #8242

### LINKED ISSUE #8242: Refactor live view to support archive

pmi:


alansemenov (approved):

ashklianko (approved):

modules/lib/src/main/resources/assets/js/app/action/PreviewActionHelper.ts
```diff
@@ -16,13 +16,17 @@ interface OpenedWindow {
 
 export class PreviewActionHelper {
 
+    private archive: boolean;
+
     private notifyBlocked: () => void;
 
-    constructor() {
+    constructor(archive: boolean = false) {
```
alansemenov: We should not have tight coupling between CS and CS+, the former should not know anything about the latter. And what if we want to send even more parameters to the preview url in the future?

Instead of having `isArchive` method and `archive` property passed all over we should do this via inheritance.

We should add `WidgetArchivePreviewAction` to CS+ which inherits from `WidgetPreviewAction` and passes optional `additionalUrlParams` object to `PreviewActionHelper` constructor. `PreviewActionHelper` will then merge whatever params are passed in the `additionalUrlParams` with "base" params in its getUrl method.

alansemenov (changes requested):

alansemenov (approved):
Nice!

ashklianko (approved):

PR #8298, opened 2025-02-05
Content links in Preview panel must navigate to items in the Grid #7998

### LINKED ISSUE #7998: Content links in Preview panel must navigate to items in the Grid

ashklianko:


modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
```diff
@@ -296,14 +297,51 @@ export class ContentBrowsePanel
...
 
     private getPathFromInlinePath(contentPreviewPath: string): string {
         return UriHelper.getPathFromPortalInlineUri(contentPreviewPath, RenderingMode.INLINE);
     }
 
+    private expandToListElementByPath(list: ContentsTreeGridList, itemPath: ContentPath, targetPath: ContentPath): void {
+        let loadCount = 0;
+
+        const itemFinder = () => {
+            const listElement = list.getListElementByPath(itemPath)
```
alansemenov: Missing ";"

modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
```diff
@@ -296,14 +297,51 @@ export class ContentBrowsePanel
...
+
+            if (listElement) {
+                list.unItemsAdded(itemFinder);
+
+                listElement.whenRendered(() => {
+                    listElement.getHTMLElement().scrollIntoView();
+
+                    if (itemPath.equals(targetPath)) {
```
alansemenov, line 323: Result of the `itemPath.equals(targetPath)` check should be assigned to a const outside of the loop, as it won't change with iterations.

modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
```diff
@@ -296,14 +297,51 @@ export class ContentBrowsePanel
...
+                        this.selectionWrapper.deselectAll();
+                        this.selectionWrapper.select(listElement.getItem());
+                    } else {
+                        listElement.expand(); // if was loaded but the children list collapsed
+                        this.expandToListElementByPath(listElement.getList(), targetPath.getPathAtLevel(itemPath.getLevel() + 1), targetPath);
```
alansemenov, line 328: Same here, `targetPath.getPathAtLevel(itemPath.getLevel() + 1)` can be calculated once.

modules/lib/src/main/resources/assets/styles/wizard/collaboration.less
```diff
@@ -4,7 +4,7 @@
 
 @space-between: 2px;
 
-@min-width: @unit-size + @border-width;
+@min-width: 26px; // @unit-size + @border-width;
```
alansemenov: Isn't it still the same as `@unit-size + @border-width`? :)

alansemenov (changes requested):

modules/lib/src/main/resources/assets/styles/wizard/collaboration.less
ashklianko: yes, but codacy was complaining :)

modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
ashklianko: it will be invoked just once, when the searched element is finally there

modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
ashklianko: same here, invoked just once

alansemenov (approved):

PR #8303, opened 2025-02-05
Minimal AVIF and WEBP support #8278

### LINKED ISSUE #8278: Basic AVIF and WEBP support
XP7.15 adds avif and webp media support. https://github.com/enonic/xp/issues/9038

However, the support is minimal: image service cannot convert to or from these formats.
CS uses ImageService for thumbnail generation - and currently fails, as well as image rotation and cropping in edit mode.
Similar to how these features are disabled for gif and svg - avif and webp should be always served pure - without any transformation.

ashklianko:


alansemenov (changes requested):
Not entirely there. webp and avif files should follow the same behaviour we have for svg files. Yes, they are images but they cannot be edited, so we should hide the toolbar with editing buttons for them.

Here's how Content Wizard looks for svg:

image

and here's how it looks (in this branch) for webp:

image

ashklianko:
Updated, but need to make couple notes:
- ImageEditor is being created for image selectors, though it is not used there, just hidden in the dom
- We show a download link for svg image content form and now for avif and webp, need to consider showing it above image editor toolbar for other images too.

alansemenov (approved):
Great job!

PR #8331, opened 2025-02-11
Wrong button inside the aggregation selector #8321

### LINKED ISSUE #8321: Wrong button inside the aggregation selector
Current "Apply" button is too wide (even worse in non-english languages) and covers the placeholder text. Also in previous version it used to be placed over the dropdown handle button, which left even more space for the filter.

Also note that there's a problem with the list of selected aggregations (above the filter). Long checkbox names push the checkbox itself too far left. Can probably be solved by setting min-width on the label's `before` pseudo-element. Finally, ellipsis overflow doesn't seem to work for long checkbox names.

**Before:**
Image

**Now:**
Image

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3879

PR #8387, opened 2025-02-19
Path is not updated after Juke inserts translated display name #8362

### LINKED ISSUE #8362: Path is not updated after Juke inserts translated display name
After content is translated by Juke Translator, content path should be updated according to the new translated display name.

Image

ashklianko:
- Implemented a workaround: emptying a path before inserting a generated display name, thus path gets generated always


alansemenov (approved):

sigdestad:
Just FYI, if you did not consider this. Path must not be made empty for a published item ofc.

alansemenov:
> Just FYI, if you did not consider this. Path must not be made empty for a published item ofc.

It's not possible to localise a published item.

PR #8412, opened 2025-02-24
Context panel icon reacts to clicks inside the Archive #8382

### LINKED ISSUE #8382: Context panel icon reacts to clicks inside the Archive
We have context panel (the panel with context widgets) both in the "main" content section and the archive, and there's two separate toggle icons, for each of the panels. However, when the toggle icon is clicked inside the Archive, the toggle icon inside the Content Studio reacts to that click, changes its state and expands its panel.
Both icons/panels should be totally separate and don't react to each other's events.

https://github.com/user-attachments/assets/0eff894b-13f7-411f-8c42-ef87b61c14fc

ashklianko:
- A global event listener is not suitable here since it can be invoked from any other built-in app, like Archive, thus all the apps get ToggleContextPanelEvent no matter which app fired it


ashklianko:
NB: need to update Plus app also

alansemenov (approved):

PR #8462, opened 2025-03-03
Content path is cleared after Juke translation #8456

### LINKED ISSUE #8456: Content path is cleared after Juke translation
After localising a content, its path gets cleared and HtmlArea input won't get translated.

https://github.com/user-attachments/assets/c1f6ab11-1204-4160-b390-9d78c2fd5906

ashklianko:
- If generated DisplayName is the same as it was then no need to reset name/path


modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
```diff
@@ -61,6 +63,10 @@ export class AiContentDataHelper {
         return CompareStatusChecker.isNew(this.compareStatus);
     }
 
+    private isTextEqualToDisplayName(text: string): boolean {
+        return ObjectHelper.stringEquals(text?.trim(), this.contentHeader.getDisplayName());
```
alansemenov: Move this check inside `isAllowedToChangeName` and get rid of the new function

alansemenov (changes requested):

alansemenov (approved):

PR #8464, opened 2025-03-03
Juke: force data update when applying results #8463

### LINKED ISSUE #8463: Juke: force data update when applying results
Currently if Translator or Operator try to update a dirty field (a field with it's value different from the saved one ) then it doesn't get updated.

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3910

alansemenov (approved):

PR #8468, opened 2025-03-03
Use widgetApiUrl from Config to build Widget and Widget Icon urls #8467

### LINKED ISSUE #8467: Use `widgetApiUrl` from Config to build Widget and Widget Icon urls
Use `widgetApiUrl` from Config to build Widget and Widget Icon urls

anatol-sialitski:


modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -149,7 +150,7 @@ export class AppWrapper
     }
 
     private updateUrl(widget: Widget): void {
-        if (widget.getUrl() === UrlAction.BROWSE.toString()) {
+        if (`${UrlHelper.getWidgetBaseUrl()}/${widget.getUrl()}` === UrlAction.BROWSE.toString()) {
```
pmi: There are many places where you create a string `${UrlHelper.getWidgetBaseUrl()}/${widget.getUrl()}`

Maybe create a `UrlHelper.getWidgetUrl(widget)` method instead that would return `${UrlHelper.getWidgetBaseUrl()}/${widget.getUrl()}` ?

anatol-sialitski: I thought about it, we have to use it for  `iconUrl` as well. Actually, `widget.getUrl()` and `iconUrl` return a part of path, not url

modules/lib/src/main/resources/assets/js/app/util/UrlHelper.ts
```diff
@@ -49,4 +49,16 @@ export class UrlHelper {
...
+        return CONFIG.getString('widgetApiUrl');
+    }
+
+    static buildWidgetUrl(path: string): string {
+        return `${UrlHelper.getWidgetBaseUrl()}/${path}`;
+    }
+
+    static buildWidgetIconUrl(iconPath: string): string {
+        return `${UrlHelper.getWidgetBaseUrl()}/${iconPath}`;
+    }
```
alansemenov: I suggest we move this to the [Widget](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/js/content/Widget.ts#L45) class in lib-admin-ui (create a separate issue), as it better belongs there. Then we can also call it directly from any app using lib-admin-ui.

```Typescript
    public getUrl(): string {
        return this.url;
    }

    public getIconUrl(): string {
        return this.iconUrl;
    }

// Add this:

    public getFullUrl(): string {
        return `${CONFIG.getString('widgetApiUrl')}/${this.url}`;
    }

    public getFullIconUrl(iconPath: string): string {
        return `${CONFIG.getString('widgetApiUrl')}/${this.iconUrl}`;
    }
```

Then replace all `UrlHelper.buildWidgetUrl(widget.getUrl())` with `widget.getFullUrl()`.

alansemenov (changes requested):

rymsha (approved):

modules/lib/src/main/resources/assets/js/app/util/UrlHelper.ts
anatol-sialitski: I do not agree. Not so long ago, I removed `adminUrl` from `lib-admin-ui`. Let's not create a strong coupling.

alansemenov: Well, current implementation makes it very hard to use widget's url. Before it was enough to call `widget.getUrl` or `widget.getIconUrl` to get a fully functional url. But in this implementation those methods are unusable by themselves. Just count in how many places you have to repeat exactly the same call of `UrlHelper.buildWidgetUrl()`. 

There's already coupling via Widget class in any app that is using widgets, so I don't really see a problem here.

anatol-sialitski: I disagree with this solution because I really do not want to create strong coupling between `lib-admin-ui` and the client apps that use it. For now, I’ve created a pull request https://github.com/enonic/lib-admin-ui/pull/3992, but we will figure out a better solution in the near future.

alansemenov (approved):
Beautiful

PR #8504, opened 2025-03-11
Multiple bug fixes

### LINKED ISSUE #8492: Icons should not be shown in the content preview toolbar
The icons distract attention and have too little value, so keep them only inside the dropdowns.

Also make sure that we set correct `aria` attributes for both dropdowns

Image

### LINKED ISSUE #8493: Help icons should not be shown for x-data fields
We have removed help icons for data fields as those might interfere with the Juke icon. But for x-data they are still shown. Remove them for x-data as well.

### LINKED ISSUE #8503: Status should not be shown when Inspect panel is narrow
When Inspect panel is < 360px we should not show status in Content Selectors (both selected options and inside the dropdowns)

Image

alansemenov:


alansemenov:
Needs https://github.com/enonic/lib-admin-ui/pull/3925 to test WCAG changes, but builds without it

ashklianko (approved):

PR #8536, opened 2025-03-17
Live Edit is reloaded when applying config changes #8521

### LINKED ISSUE #8521: Live Edit is reloaded when applying config changes
Live Edit is reloaded when applying config changes for component with contributions.

Reproduced with a part/layout with a config that contains an HtmlArea input: applying changes from Inspection panel makes entire live page reload, but only the component itself must be affected and reloaded

ashklianko:


modules/lib/src/main/resources/assets/js/app/wizard/page/LiveFormPanel.ts
```diff
@@ -516,9 +516,17 @@ export class LiveFormPanel
         saveAction.onExecuted(() => {
             const selectedItem = this.lastInspectedItemPath ? PageState.getComponentByPath(this.lastInspectedItemPath) : null;
 
+            if (selectedItem instanceof DescriptorBasedComponent) {
+                this.contentWizardPanel.setMarkedAsReady(false);
```
alansemenov, line 520: Why?

ashklianko: To signal that we are not loading a new component but reloading existing, thus even though it has page contributions we won't reload entire page

alansemenov (approved):

PR #8597, opened 2025-03-25
Support Language of Parts in the RichText editor

### LINKED ISSUE #8596: Support Language of Parts in the RichText editor
[WCAG 3.1.2 Language of Parts](http://www.w3.org/TR/UNDERSTANDING-WCAG20/meaning-other-lang-id.html)

Plugin for CKE4: https://ckeditor.com/cke4/addon/language (license OK).

The list of languages should ideally be configured on the project level. In the first version can be hardcoded to the same languages we support in Content Studio.

Update the [list of supported plugins](https://developer.enonic.com/docs/xp/stable/cms/schemas/input-types/htmlarea#usage) when done.

alansemenov:


alansemenov:
NB! Plugin is not a default one. To enable it, it should be specified in the HtmlArea's config (see below):

```xml
    <input name="htmlarea_text" type="HtmlArea">
      <label>HTML Area with customized toolbar</label>
      <occurrences minimum="0" maximum="2"/>
      <config>
        <exclude>*</exclude>
        <include>JustifyLeft JustifyRight | Bold Italic | Language</include>
      </config>
    </input>
```

ashklianko (approved):

PR #8598, opened 2025-03-25
Create Issue modal dialog: issue item is not selected in the list #8576

### LINKED ISSUE #8576: Create Issue modal dialog: issue item is not selected in the list
1. Select a content and open `Create Issue` modal dialog
2. Click on node toggler in the dialog and switch  to tree mode

Image

3. Close the dialog and reopen it again 
4. Expand the dropdown list 

**BUG**  - The selected item is not checked in the expanded list , Tree mode should be switched to flat mode after closing the dialog 

Image

ashklianko:


modules/lib/src/main/resources/assets/js/app/inputtype/selector/ContentTreeSelectorDropdown.ts
```diff
@@ -278,4 +279,12 @@ export class ContentTreeSelectorDropdown
 
         return true;
     }
+
+    isInTreeMode(): boolean {
+        return this.treeMode;
+    }
+
+    setMode(mode: 'tree' | 'flat'): void {
```
alansemenov, line 287: 1. We should use enums instead of hardcoded strings. Especially if we're going to use it in many classes.

2. It's a bit strange that in `isInTreeMode` we check value of `this.treeMode` property, while `setMode` does something completely different and doesn't change `treeMode` at all. Either `setMode` should also modify `treeMode`, or `isInTreeMode` should check whether `this.modeButton` is active.

ashklianko:
Updated in https://github.com/enonic/app-contentstudio/pull/8603

PR #8624, opened 2025-03-28
Move WebSocket URL to config #8353

### LINKED ISSUE #8353: Move WebSocket URL to config
Use `portal.apiUrl()` to generate URL from shared-socket

edloidas:
Moved WebSocket URL from hardcoded variable to 'init' message. Moved from assets and hardcoded URL to portal's apiUrl in config.

### DIFF
--- a/modules/app/src/main/resources/assets/shared-socket/websocket/init.ts
+++ b/modules/app/src/main/resources/assets/shared-socket/websocket/init.ts
@@ -1,5 +1,5 @@
 import {computed, map} from 'nanostores';
-import {WS_PROTOCOL, WS_URL} from './constants';
+import {WS_PROTOCOL} from './constants';
 import {isWebSocketMessage, WebSocketMessage} from './data';
 
 type WebSocketState = 'connecting' | 'connected' | 'disconnecting' | 'disconnected';
@@ -51,7 +51,7 @@ let reconnectTimeout: number;
 
 let customMessageHandler: (message: Record<string, unknown>) => void;
 
-export function connect(customHandler: (message: Record<string, unknown>) => void): void {
+export function connect(url: string, customHandler: (message: Record<string, unknown>) => void): void {
     const {state, connection} = $websocket.get();
 
     if (state === 'connecting' || state === 'connected') {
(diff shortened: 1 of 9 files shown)


modules/app/src/main/resources/assets/shared-socket/worker/init.ts
```diff
@@ -1,6 +1,5 @@
-import {$isConnected, connect as connectWebSocket} from "../websocket/init";
-import {isMessageWithMetadata} from "./data";
-import {addPort, broadcast, sendTo, sendToId} from "./ports";
+import {$isConnected} from "../websocket/init";
+import {addPort, broadcast, sendTo} from "./ports";
```
alansemenov: We're not using double quotes anywhere. Strange that Linter didn't complain.

modules/app/src/main/resources/assets/shared-socket/websocket/init.ts
```diff
@@ -64,7 +66,13 @@ export function connect(customHandler: (message: Record<string, unknown>) => voi
         cleanup(connection);
     }
 
-    const ws = new WebSocket(WS_URL, [WS_PROTOCOL]);
+    if (url == null) {
```
rymsha: Why would url be optional in connect method? IMO error must be thrown

modules/app/src/main/resources/assets/shared-socket/worker/init.ts
edloidas: Ah, yeah. Will fix. It's CS, so we have less strict rules here.

modules/app/src/main/resources/assets/shared-socket/websocket/init.ts
edloidas: Hmm, well, I found a better approach and fixed it.

Here are a few comments regarding the previous code and suggestions:
- `Optional<T>` here is our custom type that effectively means `T | undefined | null`. In this case, the URL is not optional as it would be in `url?: string`. Instead, it must explicitly be provided each time. The `Optional<T>` type is simply indicating that the value can be `null`, so it's important to handle it explicitly. This works well in strict TypeScript when transitioning from unknown results (such as from a `JSON.parse` call) to our own code, where we're 100% certain about nullish values.
- Regarding throwing errors, it's generally better to avoid doing this in JavaScript because we cannot gracefully handle them as easily as in Java. Errors in JavaScript usually indicate that you've f***ed up somewhere. Having Go-style error returns (like we do in some of our server-side JS code in other apps) would add complexity at this stage. Therefore, try-catch blocks should be kept for situations where the implementation is hidden from us and cannot be controlled, such as with JSON.parse, engine-native functions, Java beans, etc.

alansemenov (approved):

rymsha (approved):

PR #8668, opened 2025-04-07
Set parent lists for tree-based list elements #8667

### LINKED ISSUE #8667: Set parent lists for tree-based list elements
Couple classes are missing an obligatory param, that breaks tree list level calculation

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3982

alansemenov (approved):

PR #8676, opened 2025-04-08
Related to lib-admin #3977

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/3990

PR #8681, opened 2025-04-08
Fix improper usage of promise void returns #8680

### LINKED ISSUE #8680: Fix improper usage of promise void returns
The current codebase contains several instances of improper Promise usage where functions return `Q(null)`. These pattern result in the return type being inferred as `any` instead of `void`, which is problematic, especially as we move toward enabling TypeScript strict mode and eliminating any.

Additionally, there are unnecessary usages of `Q(null)`, `Q.resolve()` where simply calling `Q()` (or nothing at all) would suffice.

edloidas:
Further refactoring related to #8677 and #8679 fixes. 
Makes code a little bit secure by returning void explicitly, instead of any. Also removes invalid or redundant code, that is not type checked due to current TSConfig.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
@@ -183,7 +183,7 @@ export class CompareWithPublishedVersionDialog
         if (updateDiff) {
             return this.displayDiff();
         } else {
-            return Q.resolve();
+            return Q();
         }
     }
 
@@ -193,7 +193,7 @@ export class CompareWithPublishedVersionDialog
         if (updateDiff) {
             return this.displayDiff();
         } else {
-            return Q.resolve();
+            return Q();
         }
     }
 
--- a/modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
@@ -182,9 +182,9 @@ export class CompareWithPublishedVersionDialog
 
         if (updateDiff) {
             return this.displayDiff();
-        } else {
-            return Q();
         }
+
+        return Q();
     }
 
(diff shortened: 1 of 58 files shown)


modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
```diff
@@ -193,7 +193,7 @@ export class CompareWithPublishedVersionDialog
         if (updateDiff) {
             return this.displayDiff();
         } else {
-            return Q.resolve();
+            return Q();
```
ashklianko: Same here, no need 'else'

modules/lib/src/main/resources/assets/js/app/dialog/CompareWithPublishedVersionDialog.ts
```diff
@@ -183,7 +183,7 @@ export class CompareWithPublishedVersionDialog
         if (updateDiff) {
             return this.displayDiff();
         } else {
```
ashklianko: Don't need 'else', can make it even simpler:

```
if (updateDiff) {
            return this.displayDiff();
        } 
       
 return Q();
        
```

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -943,15 +942,15 @@ export class ContentWizardPanel
 
     private handleNonRenderablePage(): Q.Promise<void> {
         this.liveEditModel = null;
-        return Q.resolve();
+        return Q();
     }
 
     private unloadPage(): Q.Promise<void> {
```
ashklianko: No need in promise here at all

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
ashklianko: And in `unloadPage` too

alansemenov (approved):

ashklianko (approved):

pmi (approved):

PR #8712, opened 2025-04-22
Add SharedWorker fallback #8301

### LINKED ISSUE #8301: Add SharedWorker fallback
For the devices, that can not use SharedWorker, we must use WebSocket directly.
Consider having a wrapper code, that will either init and start a SharedWorker, or execute a script directly and establish WebSocket connection in the Tab.

edloidas:


alansemenov (approved):
Looking good! I assume you've tested it on a browser with no SharedWorker support :)

PR #8738, opened 2025-04-30
New Permissions UI #7506

### LINKED ISSUE #7506: New Permissions UI
The new permission management will be steered by the modal wizard dialog that consists of three steps + the Summary step. User can apply the changes on every step and skip the remaining ones.

Key points:
* No more "inherit permissions" checkbox, all permissions are explicitly set on a content
* New "Copy from parent" button to copy permissions from the direct parent (if there's any) 
* Principal selector is always in "edit" mode, never readonly
* Simplified permission management for the Everyone role
* No more "Overwrite child permissions" checkbox - it will be done on a separate step where user can select one of three update policies
* "Next" button is disabled if no changes have been made
* "Apply" button inside the "Next" button menu, to apply changes on any step and close the dialog

ashklianko:


alansemenov:
@ashklianko this should not be a PR to master. The task is an epic, so we need an epic branch and all PRs will go there.

alansemenov:
Or, if this is already a PR of the epic branch, then it should have a different title (same as the epic) and it should be in Draft until all of the tasks are done.

alansemenov (approved):

PR #8739, opened 2025-04-30
Websocket improvements #8719

### LINKED ISSUE #8300: Websocket improvements
Our Websockets implementation is unstable. Content Studio opens a new websocket connection in every tab, and so do Collaboration and Juke apps as well. Combined with limits on Websocket connections that browsers and dns providers have, users may experience error message notifying that websockets stopped working. Without websockets functioning Content Studio becomes blind to server events.

Read more here: https://enonic.atlassian.net/wiki/spaces/TD/pages/3258220545/2025-01-08+Websockets+improvements

The purpose of this epic is to have a stable solution where Content Studio opens just one websocket connection and lets its instances (CS opened in other browser tabs) and tenants (Collaboration, Juke apps) reuse it without opening new connections.

edloidas:


alansemenov (approved):
Great job, @edloidas!

PR #8791, opened 2025-05-15
Update/Remove redundant XP methods #8784

### LINKED ISSUE #8784: Update/Remove redundant XP methods
Studio won't build with changes in https://github.com/enonic/xp/pull/11074/files#diff-01e2f22294929ba0105209cde29a42d0d2b23265bf4441ad15e80aaa4df0cf95, fix all the issues

ashklianko:
Required for https://github.com/enonic/xp/pull/11074


modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/project/ApplyProjectReadAccessPermissionsCommand.java
```diff
@@ -51,7 +51,6 @@ private TaskResultJson doExecute()
                 params( ApplyContentPermissionsParams.create().
                     permissions( newList ).
                     contentId( contentRoot.getId() ).
-                    overwriteChildPermissions( true ).
```
rymsha, line 54: `applyPermissionsScope` must be used instead

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/query/ContentQueryWithChildren.java
```diff
@@ -52,13 +53,13 @@ private QueryExpr constructExprToFindChildren()
 
         for ( ContentPath contentPath : contentsPaths )
         {
-            ConstraintExpr likeExpr =
-                CompareExpr.like( fieldExpr, ValueExpr.string( ContentHelper.getContentRoot() + "/" + contentPath.asRelative() + "/*" ) );
+            ConstraintExpr likeExpr = CompareExpr.like( fieldExpr, ValueExpr.string(
+                ContentHelper.getContentRoot() + "/" + ContentPathHelper.toRelative( contentPath ) + "/*" ) );
```
rymsha: use `ContentPath.from` instead

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentPathHelper.java
```diff
@@ -0,0 +1,27 @@
+package com.enonic.xp.app.contentstudio.rest.resource.content;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.enonic.xp.content.ContentPath;
+
+public class ContentPathHelper
```
rymsha: use `ContentPath.from` instead

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ComponentNameResolverImpl.java
```diff
@@ -53,58 +52,58 @@ else if ( component instanceof TextComponent )
     }
 
 
-    public ComponentName resolve( final PartComponent component )
+    public String resolve( final PartComponent component )
     {
         final PartDescriptor partDescriptor =
             component.getDescriptor() != null ? partDescriptorService.getByKey( component.getDescriptor() ) : null;
-        return partDescriptor != null ? ComponentName.from( partDescriptor.getDisplayName() ) : component.getName();
+        return partDescriptor != null ?  partDescriptor.getDisplayName() : component.getType().toString();
```
rymsha: it is not 1:1 fix.  Look how `CreateFragmentCommand` is changed in XP. But maybe it is OK here
Actualy there is no such thing as ComponentName in XP8. Look how to avoid using this class alltoghether

modules/app/src/test/java/com/enonic/xp/app/contentstudio/LiveEditInjectionTest.java
```diff
@@ -87,7 +87,7 @@ private void injectAndAssert(final String templateName)
         assertEquals( readResource( templateName ).trim(), result.trim() );
     }
 
-    @Test
+ //   @Test
```
rymsha: Fix the tests

rymsha (changes requested):

modules/app/src/test/java/com/enonic/xp/app/contentstudio/LiveEditInjectionTest.java
ashklianko: Wasn't supposed to be commited, tests are working

modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentPathHelper.java
ashklianko: Won't work, sole purpose of this helper is to replace a function from ContentPath -  asRelative() that was removed, and it was meant to set 'absolute' param of ContentPath to 'false', but 'ContentPath.from' copies the value of 'absolute' from parent, without possibility to override it. Also it is only possible to create a builder class for ContentPath with empty constructor, would be nice to have a builder 'ContentPath.create( final ContentPath source)', then the helper would lool like:
```

public static ContentPath toRelative( final ContentPath source )
    {
        return new ContentPath.Builder(source).absolute( false ).build();
    }
```

rymsha: it was removed because relative content paths are evil
what is wrong with
`public static ContentPath from( final ContentPath parent, final String name )` ?

And why not using builder where it is still possible to configure absolute/relative.

For sure - constructing content path via string is the worst option

rymsha (approved):

PR #8840, opened 2025-06-10
Support baseUrl in Content Studio #8479

### LINKED ISSUE #8479: Support baseUrl in Content Studio
`baseUrl` will be edited via a text input field (TextLine) in the site and project wizard forms and persisted in site config for `portal` app. Example of persisted property:

```
"data": {
    "siteConfig": [
      {
        "applicationKey": "portal",
        "config": { "baseUrl": "https://www.example.com:8443/my-site" }
        ...
```

We should manually build data object so that value of the baseUrl input field gets added to (or removed from, if empty) the list of site configs on save and make sure "portal" app is not visible in the UI (as it is not a "regular" app).

Image
Image

ashklianko:
- Added SiteContentWizardStepForm.ts for sites to intercept content form creation to handle a new "Base URL" field below "Description". It acts as a proxy for a hidden SiteConfig property so need to add/remove according to changes in Base URL field

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/inputtype/siteconfigurator/SiteConfigProviderRegistry.ts
+++ b/modules/lib/src/main/resources/assets/js/app/inputtype/siteconfigurator/SiteConfigProviderRegistry.ts
@@ -1,0 +1,18 @@
+import {ApplicationConfigProvider} from '@enonic/lib-admin-ui/form/inputtype/appconfig/ApplicationConfigProvider';
+
+export class SiteConfigProviderRegistry {
+
+    private static siteConfigProvider: ApplicationConfigProvider;
+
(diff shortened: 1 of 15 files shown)


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/4054

ashklianko:
One thing to update: make the Portal app be always last in Site Configs

alansemenov (changes requested):
It works fine, but we need to disable the Base Url field in the UI if user doesn't have permissions to change it (should use the same permissions check as for disabling app config in the Applications dropdown underneath). There's a backend check which prevents saving in case of insufficient permissions, but we need to update the UI as well.

alansemenov (approved):

PR #8919, opened 2025-07-15
Replace "show changes" icon with checkboxes #8909

### LINKED ISSUE #8909: Use checkboxes for comparing versions in Version History
- Show the checkbox only on hover
- Don't show the checkbox on hover if 2 versions are selected
- Don't show the checkbox when only one version is comparable

ashklianko:


ashklianko:
Added https://github.com/enonic/app-contentstudio/issues/9011 here

alansemenov (changes requested):
When "Compare" button docks to the top on scroll there should be padding on top of the list so that it doesn't scroll under the button. See video

https://github.com/user-attachments/assets/9d1f1f19-0c47-4eb5-99d5-937b24f166ab

ashklianko:
> When "Compare" button docks to the top on scroll there should be padding on top of the list so that it doesn't scroll under the button. See video
> 
>  bug-version-history-scroll.mov

Updated

alansemenov (approved):

PR #8924, opened 2025-07-16
Version History based on node API part 2 #8923

### LINKED ISSUE #8923: Version History based on node API part 2
See https://github.com/enonic/app-contentstudio/issues/8812
Moving more content versions related classes from XP to Studio, including ContentVersions class that needs to have incorrect sorting removed

ashklianko:


ashklianko:
Fixes the sorting issue, now all versions will be sorted by the version's timestamp in descending order, no other rearrangements will be applied

vbradnitski (approved):

PR #8978, opened 2025-08-05
Optimise project fetching request

### LINKED ISSUE #8977: Optimise project fetching request
Currently we make a rest call to fetch the projects that are available for the current user, and then if some of those projects have parents that did not come in the response, we separately fetch projects graph in 'getTree' calls.

Those missing projects can be found and fetched during the first 'list' call and retuned immediately without extra requests.

ashklianko:


modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/project/ProjectResource.java
```diff
@@ -163,11 +167,69 @@ public boolean delete( final DeleteProjectParamsJson params )
...
     public ProjectsJson list()
     {
-        final List<ProjectJson> projects = this.projectService.list().stream().map( this::doCreateJson ).collect( Collectors.toList() );
+        final List<ProjectJson> projects = this.projectService.list().stream().map( this::doCreateJson ).collect( toList() );
 
         return new ProjectsJson( projects );
     }
 
+    @GET
+    @Path("listWithMissing")
```
alansemenov: This is not a good name. list with missing what? if this is a complete list (as opposed to `list` being limited to only projects available for the current user), then it should be called `all` or `allProjects`. Or this one should be renamed to `listAll` and the other one to `listAvailable`

alansemenov (approved):

vbradnitski (approved):

PR #8990, opened 2025-08-08
Upgrade target ES version and optimize build

### LINKED ISSUE #8984: Migrate to newer version of JS
After migrating LibAdminUI to a newer version of JS, we need to make it compatible with it and migrate all code to ES2015+.

### LINKED ISSUE #8991: Optimize build
As soon as #8984 is ready, we should upgrade other parts of the build:
- Migrate from SWC + Webpack to Vite, like we did in LibAdminUI to simplify build and make it faster.
- Move to PNPM to handle dependencies properly.
- Enable caching in Gradle to make consecutive builds in Gradle faster.
- Add `.npmrc` files to fail on missing peer dependencies.
- Remove unused dependencies, like `concurrently`.

### LINKED ISSUE #8995: Move build to pnpm
We need to migrate our dependency resolution to pnpm, because npm handles complex projects poorly — especially those that include links to local dependencies (e.g., `file:` references).
We also discovered that some dependencies were missing from our own project and were instead reused from other projects, e.g.:
  - `jquery` is not listed in the app’s dependencies and is implicitly reused from lib-admin-ui.
  - Bunch of loaders and PostCSS plugins.
  - `@swc/core` was cached in CI and installed incorrectly.

pnpm ensures that the application uses its own installed dependencies rather than implicitly resolving them from other packages, unless explicitly defined as peer dependencies.

edloidas:
Migrated from ES5 (as of swc builds) to ES2023.
    Set `useDefineForClassFields` of TSConfig and SWC to false to keep old classes specs to prevent app from breaking.
Renamed usage of `lib.lite.css` to `lib-lite.css` from lib-admin-ui. 
Added missing npm dependencies.
Replaced and renamed some types in Config and classes props.
Upgraded to the latest LTS node version.

---

Enabled and configured the Gradle configuration cache, parallel execution, and build cache to speed up local and CI builds.
Refactored build scripts to be compatible with the configuration cache by removing dynamic file checks, `afterEvaluate` blocks, and fixing cross-build dependencies.
Upgraded Gradle dependencies.
Copied LICENSE to modules folder for enonicDefaults to work properly.
Updated deprecated features like archivesBaseName to newer syntax.

---

Moved from npm to pnpm.
Further optimized Gradle build, caching checks and moving to modern syntax.


alansemenov (approved):
Wonderful job!

PR #9060, opened 2025-08-27
figcaption element should not be added if image caption is empty #9057

### LINKED ISSUE #9057: `figcaption` element should not be added if image caption is empty
We made a first attempt on fixing the issue [here](https://github.com/enonic/app-contentstudio/issues/7907) but had to roll it back because if affected image alignment in the dialog and decided to hide `figcaption` [with css](https://github.com/enonic/app-contentstudio/issues/8824) instead.

Unfortunately, screen readers still find `figcaption` in the DOM and treat such an image element as normal, not decorative.

We need to find a proper solution to the problem so that empty `figcaption` is not added to the DOM.

ashklianko:


alansemenov (approved):
good stuff

PR #9082, opened 2025-09-04
Enonic UI: Icon image-handling #9077

### LINKED ISSUE #9077: Enonic UI: Icon image-handling
CheckableListItemWithStatus does not render images.

Image

Update ContentIcon and CheckableListItemWithStatus to handle and render images and svg's instead of icons where url is present.

If url is not present, use icon as fallback.

skjulestad:
Added image rendering to ContentIcon.tsx with fallback for error handling.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
+++ b/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
@@ -5,7 +5,6 @@ import {DialogDependantItemsList, ObserverConfig} from '../dialog/DialogDependan
 import {ContentServerChangeItem} from '../event/ContentServerChangeItem';
 import {ContentServerEventsHandler} from '../event/ContentServerEventsHandler';
 import {CheckableListItemWithStatus} from '../ui2/list/CheckableListItemWithStatus';
-import {ContentSummaryViewer} from '../content/ContentSummaryViewer';
 
 
 export class PublishDialogDependantList
@@ -41,15 +40,9 @@ export class PublishDialogDependantList
 
     createItemView(content: ContentSummaryAndCompareStatus, readOnly: boolean): CheckableListItemWithStatus {
         const className = this.isItemHidden(content) ? 'hidden' : undefined;
-        const viewer = new ContentSummaryViewer();
-        const summary = content.getContentSummary();
-        if (summary) {
-            viewer.setObject(summary);
-        }
 
         return new CheckableListItemWithStatus({
             content,
-            contentViewer: viewer,
             className,
             readOnly: readOnly || !this.isItemExcludable(content),
             checked: this.mustSelectItem(content),
--- a/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
+++ b/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
@@ -5,6 +5,8 @@ import {DialogDependantItemsList, ObserverConfig} from '../dialog/DialogDependan
 import {ContentServerChangeItem} from '../event/ContentServerChangeItem';
 import {ContentServerEventsHandler} from '../event/ContentServerEventsHandler';
 import {CheckableListItemWithStatus} from '../ui2/list/CheckableListItemWithStatus';
+import { ContentSummaryViewer } from '../content/ContentSummaryViewer';
+
 
 export class PublishDialogDependantList extends DialogDependantItemsList {
 
@@ -39,8 +41,13 @@ export class PublishDialogDependantList extends DialogDependantItemsList {
 
     createItemView(content: ContentSummaryAndCompareStatus, readOnly: boolean): CheckableListItemWithStatus {
         const className = this.isItemHidden(content) ? 'hidden' : undefined;
+        const viewer = new ContentSummaryViewer();
+        const summary = content.getContentSummary();
+        if (summary) viewer.setObject(summary);
+
         return new CheckableListItemWithStatus({
             content,
+            contentViewer: viewer,
             className,
             readOnly: readOnly || !this.isItemExcludable(content),
             checked: this.mustSelectItem(content),
--- a/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
+++ b/modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
@@ -5,13 +5,14 @@ import {DialogDependantItemsList, ObserverConfig} from '../dialog/DialogDependan
 import {ContentServerChangeItem} from '../event/ContentServerChangeItem';
 import {ContentServerEventsHandler} from '../event/ContentServerEventsHandler';
 import {CheckableListItemWithStatus} from '../ui2/list/CheckableListItemWithStatus';
-import { ContentSummaryViewer } from '../content/ContentSummaryViewer';
+import {ContentSummaryViewer} from '../content/ContentSummaryViewer';
 
 
-export class PublishDialogDependantList extends DialogDependantItemsList {
+export class PublishDialogDependantList
+    extends DialogDependantItemsList {
 
     private requiredIds = ContentIds.empty();
-    private visibleIds  = ContentIds.empty();
+    private visibleIds = ContentIds.empty();
 
     private listChangedListeners: (() => void)[] = [];
     private visibleUpdatedListeners: (() => void)[] = [];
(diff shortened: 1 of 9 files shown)


modules/lib/src/main/resources/assets/js/app/publish/PublishDialogDependantList.ts
```diff
@@ -36,11 +39,17 @@ export class PublishDialogDependantList extends DialogDependantItemsList {
...
 
 
-
     createItemView(content: ContentSummaryAndCompareStatus, readOnly: boolean): CheckableListItemWithStatus {
         const className = this.isItemHidden(content) ? 'hidden' : undefined;
+        const viewer = new ContentSummaryViewer();
+        const summary = content.getContentSummary();
+        if (summary) {
+            viewer.setObject(summary);
+        }
```
edloidas: This isn’t quite correct. You’re mixing two presentational components here. On top of that, the **Viewer** you’re using is the legacy component we’re actually trying to remove at the moment.

_Think of it like this: you had an old Windows laptop that could play MKV video files Now you’ve switched to a new Mac - modern, sleek, but it doesn’t natively support MKV yet. Instead of installing the right codec, you’re setting up your old laptop next to it just to play those files. That’s clunky and exactly what we should avoid._

We must not use the Viewer here.
What you actually need is just the `resolveIconUrl()` method. It’s relatively small, so you can call the utility class directly: `ContentIconUrlResolver`.

Even better, you could add a method in the `ContentIcon` file that uses `UriHelper` directly, like in the resolver mentioned above:

```js
${UriHelper.trimUrlParams(url)}?${UriHelper.encodeUrlParams(params)}
```

You’ll just need to pass the right parameters like `size` or `crop`.

* `crop` can actually be omitted.
* `size` should be set reasonably - something like double the intended icon size (e.g. `64`) works well.

modules/lib/src/main/resources/assets/js/app/ui2/list/CheckableListItemWithStatus.tsx
```diff
@@ -7,13 +8,16 @@ import {ContentIcon} from './ContentIcon';
...
 
-const CheckableListItemWithStatusComponent = ({content, ...props}: Props): JSX.Element => {
+const CheckableListItemWithStatusComponent = ({content, contentViewer, ...props}: Props): JSX.Element => {
 
     const label = content.getPath().toString();
+    const summary = content.getContentSummary();
+    const imageUrl = summary ? contentViewer.resolveIconUrl(summary) : null;
     const status = CompareStatusFormatter.formatStatusText(content.getCompareStatus());
-    const icon = <ContentIcon contentType={content.getType()}/>;
+    const icon = <ContentIcon contentType={content.getType()} imageUrl={imageUrl}/>;
```
edloidas: Looking at this, I think it makes more sense to handle the image inside ContentIcon, since it may be used in multiple places. We don’t want to duplicate the URL-resolving logic.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -24,61 +24,136 @@ import {
...
+    case String(ContentTypeName.IMAGE):
+        return <FileImage/>;
+    case String(ContentTypeName.MEDIA_IMAGE):
+        return <ImageIcon/>;
+    case String(ContentTypeName.MEDIA_VECTOR):
+        return <SplinePointer/>;
+    default:
+        return <FileIcon/>;
+    }
+};
```
edloidas: It can stay here, but I don’t see any reason why it should look like a regular function instead of a component:

```tsx
const FallbackContentIcon = ({type}: {type: ContentTypeName}): JSX.Element => {
```

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -24,61 +24,136 @@ import {
...
-    default:
-        return <FileIcon/>;
 
+    default:
+        return (
+            <ImgWithFallback
+                src={imageUrl}
+                fallback={fallbackIconFor(contentType)}
+            />
+        );
```
edloidas: Are you returning image with fallback four times? It might be reasonable just to remove the case block for `ContentTypeName.MEDIA_VECTOR`, `ContentTypeName.MEDIA_IMAGE`, `ContentTypeName.IMAGE` then. Also, for such a small icon size, `object-fit: cover;` might be the universal choice. `contain` might look awful for wide and long images.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -24,61 +24,136 @@ import {
...
+    alt?: string;
+    fallback?: JSX.Element;
+    imgClassName?: string;
+}): JSX.Element => {
+    const [errored, setErrored] = useState(false);
+    if (!src || errored) {
+        return fallback;
+    }
+    return <img className={imgClassName} alt={alt} src={src} onError={() => setErrored(true)}/>;
+};
```
edloidas: Move this to a separate file and name it simply `<Image>`. This might become our default way of rendering images. There could be a lot going on here, but let’s keep it simple for now:

```tsx
<img
  className={className}
  alt={alt}
  width={size}        // optional
  height={size}       // optional
  loading={lazy ? 'lazy' : 'eager'} // eager by default
  decoding="async"
  referrerPolicy="no-referrer"
  src={src}
  srcSet={srcSet}     // omit for now, but we may add it later
  onError={handleError}
/>
```

`src` must always be set and treated as non-null.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
skjulestad: We need contain for external icons like author and cover for svg/images as contain is clipping the icons. Svg's / images need cover to look good and square along other icons. All icons are rendered as either svg og img so we do need to separate by type to avoid miss-matching cover and cointain.

Without it we will get either of these end results:
image
image

modules/lib/src/main/resources/assets/js/app/ui2/list/CheckableListItemWithStatus.tsx
```diff
@@ -1,33 +1,41 @@
...
 const CheckableListItemWithStatusComponent = ({content, ...props}: Props): JSX.Element => {
-
     const label = content.getPath().toString();
     const status = CompareStatusFormatter.formatStatusText(content.getCompareStatus());
-    const icon = <ContentIcon contentType={content.getType()}/>;
+
+    const icon = (
+        <ContentIcon
+            contentType={content.getType()}
+            summary={content.getContentSummary?.()}
```
edloidas: I'm sure that we don't need optional chaining here, as `getContentSummary` function is present in objects of this type. You don't need to verify it exists.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -1,84 +1,94 @@
...
+    imgClassName?: string;
+};
+
+const buildIconUrl = (summary: ContentSummary, size = 64, crop?: boolean): string => {
+    const url = new ContentIconUrlResolver().setContent(summary).resolve();
+    const params: Record<string, string | number | boolean> = {size};
+    if (crop) {
+        params.crop = true;
+    }
+    return `${UriHelper.trimUrlParams(url)}?${UriHelper.encodeUrlParams(params)}`;
```
edloidas: Why are you using both `ContentIconUrlResolver` and `UriHelper` here? I see that the first line there is doing exactly the same as the rest of the code here, but inside the `ContentIconUrlResolver` class. 
Ideally, we must create a utility function in a new file with no defaults and use it here. Just a smaller and smarter version of `ContentIconUrlResolver`:
```ts
interface ImageUrlParams {
    size?: number | null;
    crop?: boolean | null;
}

export function createImageUrl(url: string, params: ImageUrlParams): string {
    return `${UriHelper.trimUrlParams(url)}?${UriHelper.encodeUrlParams(params)}`;
}
```

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -1,84 +1,94 @@
...
+        return tryImage(imgClassName) ?? icon ?? <FileIcon/>;
     }
-}
+
+    if (icon) {
+        return icon;
+    }
+
+    return tryImage() ?? <FileIcon/>;
+};
```
edloidas: While it's generally correct, it's hard to read, as it mixes variables, function call and components.
Also, regarding rendering logic there:
- If the content is image, you try image -> icon -> FileIcon (obsolete, but done for type safety, as your map has string keys)
- You try icons instead with fallback to image, if type is unknown. File won't probably be used here also.
To simplify to simplify the flow, we can write it like this:
```tsx
if (isImageType) {
    return tryImage(imgClassName) ?? icon ?? <FileIcon />;
}

return icon ?? tryImage() ?? icon ?? <FileIcon />;
```
I think that `imgClassName` should not be called like this. Actually, class must be set here, in this component and not from the outside.
So, we use image, if content type is unknown, or if it's image type.

edloidas (changes requested):

PR #9106, opened 2025-09-10
Controlled x-data sorting #6450

### LINKED ISSUE #6450: Incorrect order of X-data in the content form
From: https://discuss.enonic.com/t/wrong-xdata-sorting/3045

We need to implement our standard approach to ordering xData by combining a priority + order of apps (if priority is equal).
Also, "page" should then be injected, based on a specific and reasonably high priority, so it will appear early in the list.

ashklianko:
- Removed unordered toMap() call and replaced with more accurate and effective code
- Moved the only method in XDataResource to XDataContextResource and removed the former


vbradnitski (approved):

alansemenov (approved):

alansemenov:
Great job!

PR #9121, opened 2025-09-12
Enonic UI: Add ContentItem #9093

### LINKED ISSUE #9093: Enonic UI: Add ContentItem
- Renamed `CheckableListItemWithStatus` ➜ `ContentItemCheckable`
- Create `ContentItem`
  - Should look similar to `ContentItemCheckable` but use `ListItem` from `@enonic/ui`, instead of `SelectableListItem`.
  - Needs to support `onClick` event.
  - Needs to accept children. Place them before status badge in the right part.
- Replace `ArchiveCheckableItem` and `ArchiveSelectableItem` with `ContentItem`.
- Customize with buttons "Show references" in place (list), if needed.
  - We should keep button as separate part to use it when replacing `TogglableStatusSelectionItem` in the future.

skjulestad:
Renamed CheckableListItemWithStatus ➜ ContentItemCheckable
Created ContentItem with support for onClick event and children.
Replaced ArchiveCheckableItem and ArchiveSelectableItem with ContentItem.
Added separate show ref button for ContentItem.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/dialog/ArchiveDialogHelper.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/ArchiveDialogHelper.ts
@@ -1,7 +1,6 @@
 import {Action} from '@enonic/lib-admin-ui/ui/Action';
 import {ActionButton} from '@enonic/lib-admin-ui/ui2/ActionButton';
 import {i18n} from '@enonic/lib-admin-ui/util/Messages';
-import {LinkIcon} from 'lucide-react';
 import {DependencyParams} from '../browse/DependencyParams';
 import {DependencyType} from '../browse/DependencyType';
 import {ShowDependenciesEvent} from '../browse/ShowDependenciesEvent';
@@ -9,17 +8,34 @@ import {ContentId} from '../content/ContentId';
 import {Branch} from '../versioning/Branch';
 
 export class ArchiveDialogHelper {
-
     static createShowReferences(contentId: ContentId, branch?: Branch): ActionButton {
         const action = new Action(i18n('action.showReferences')).onExecuted(() => {
             const params = DependencyParams.create()
-                    .setContentId(contentId)
-                    .setDependencyType(DependencyType.INBOUND)
-                    .setBranch(branch)
-                    .build();
+                .setContentId(contentId)
+                .setDependencyType(DependencyType.INBOUND)
+                .setBranch(branch)
+                .build();
             new ShowDependenciesEvent(params).fire();
         });
 
-        return new ActionButton({action, startIcon: LinkIcon, className: 'show-ref'});
+        return new ActionButton({action, className: 'show-ref'});
+    }
+
+    static mountShowReferences(
+        host: HTMLElement,
+        contentId: ContentId,
+        branch?: Branch
+    ): () => void {
+        const btn = ArchiveDialogHelper.createShowReferences(contentId, branch);
+        const el = btn.getHTMLElement();
+        host.appendChild(el);
+
+        return () => {
+            try {
+                btn.remove();
+            } catch {
+                el.parentNode?.removeChild(el);
+            }
+        };
     }
 }
--- a/modules/lib/src/main/resources/assets/js/app/dialog/ArchiveDialogHelper.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/ArchiveDialogHelper.ts
@@ -20,22 +20,4 @@ export class ArchiveDialogHelper {
 
         return new ActionButton({action, className: 'show-ref'});
     }
-
-    static mountShowReferences(
-        host: HTMLElement,
-        contentId: ContentId,
-        branch?: Branch
-    ): () => void {
-        const btn = ArchiveDialogHelper.createShowReferences(contentId, branch);
-        const el = btn.getHTMLElement();
-        host.appendChild(el);
-
-        return () => {
-            try {
-                btn.remove();
-            } catch {
-                el.parentNode?.removeChild(el);
-            }
-        };
-    }
 }
--- a/modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsWithReferencesDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsWithReferencesDialog.ts
@@ -5,7 +5,6 @@ import {ContentWithRefsResult} from '../resource/ContentWithRefsResult';
 import {ContentId} from '../content/ContentId';
 import {i18n} from '@enonic/lib-admin-ui/util/Messages';
 import {ContentSummaryAndCompareStatus} from '../content/ContentSummaryAndCompareStatus';
-import {ListBox} from '@enonic/lib-admin-ui/ui/selector/list/ListBox';
 import {ContentServerChangeItem} from '../event/ContentServerChangeItem';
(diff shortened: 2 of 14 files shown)


modules/lib/src/main/resources/assets/js/app/dialog/ArchiveDialogHelper.ts
```diff
@@ -1,25 +1,41 @@
...
+        host.appendChild(el);
+
+        return () => {
+            try {
+                btn.remove();
+            } catch {
+                el.parentNode?.removeChild(el);
+            }
+        };
     }
```
edloidas: It's a hack, that you use to make regular Element to render in component.
Show reference `ActionButton` is a JSX Preact component, that was wrapped into Element in lib-admin-ui. Then you'll pass it to JSX in ShowReferencesButton just to put it back to back to virtual dom.
We don't actually need it.

modules/lib/src/main/resources/assets/js/app/ui2/list/ShowReferencesButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+    }, [contentId, target]);
+
+    return (
+        <div
+            ref={mountRef}
+            className={className}
+            onClick={(e) => e.stopPropagation()}
+            onMouseDown={(e) => e.stopPropagation()}
+        />
+    );
```
edloidas: Just use a simple button from Enonic UI instead.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+        if (!isInteractive) return;
+        if (e.key === 'Enter' || e.key === ' ') {
+            e.preventDefault();
+            onClick?.();
+        }
+    };
+
+    return (
+        <ListItem
+            className={['archive-item', isInteractive ? 'clickable' : '', className].join(' ').trim()}
```
edloidas: Better use our `cn` utility from `@enonic/ui`, than this.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+            selected={selected}
+            onClick={isInteractive ? onClick : undefined}
+            onKeyDown={onKeyDown}
+            role={isInteractive ? 'button' : undefined}
+            aria-disabled={isInteractive ? undefined : true}
+            tabIndex={isInteractive ? 0 : undefined}
```
edloidas: After discussion, we aren't going to make the whole item clickable, only the content part. We need to find a way to wrap it into button. I'll see what needs to be adjusted in the Enonic UI. Meanwhile, just remove those hacks.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+        >
+            <ListItem.Content label={label} icon={Icon} />
+            <ListItem.Right>
+                {children}
+                {showReferences && hasInbound && (
+                    <ShowReferencesButton
+                        contentId={contentId}
+                        target={target}
+                        className="show-references-btn"
```
edloidas: We don't need such classes if they are not used or setting some styles.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+        this.props.setKey('selected', selected);
+    }
+
+    setOnClick(callback: () => void): void {
+        this.props.setKey('onClick', callback);
+    }
+
+    setHasInbound(hasInbound: boolean): void {
+        this.hasInboundClass = hasInbound;
+        this.toggleClass('has-inbound', hasInbound);
```
edloidas: Again, such class is used in old archive items to hide some children. It has no use here.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+        this.props.setKey('hasInbound', hasInbound);
+    }
+
+    hasInbound(): boolean {
+        return this.hasInboundClass;
+    }
```
edloidas: Is this code really called anywhere? I doubt that we need it or `hasInboundClass` property.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+        return this.props.get().clickable !== false;
+    }
+
+    setSelected(selected: boolean): void {
+        this.props.setKey('selected', selected);
+    }
+
+    setOnClick(callback: () => void): void {
+        this.props.setKey('onClick', callback);
+    }
```
edloidas: What is the purpose of this method? I can't see that it's used somewhere. Remove it if it's unnecessary. Double check other methods too.

modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsWithReferencesDialog.ts
```diff
@@ -6,16 +6,32 @@ import {ContentId} from '../content/ContentId';
...
+    if (v === null || typeof v !== 'object') return false;
+
+    const m = v as Partial<InboundAwareView>;
+    return typeof m.setHasInbound === 'function'
+           && typeof m.getItem === 'function';
+}
+
+
+
+
```
edloidas: You can replace this one with 
```ts
type InboundAwareItem = ArchiveCheckableItem | ArchiveSelectableItem | ContentItem;
```
After that, you could just remove runtime type checks. Archive items will probably be removed soon. 
And what are those empty lines for?

modules/lib/src/main/resources/assets/js/app/dialog/DependantItemsWithReferencesDialog.ts
```diff
@@ -42,12 +58,18 @@ export abstract class DependantItemsWithReferencesDialog extends DependantItemsW
     protected initListeners(): void {
         super.initListeners();
 
-        const itemsAddedHandler = (items: ContentSummaryAndCompareStatus[], itemList: ListBox<ContentSummaryAndCompareStatus>) => {
-            if (this.resolveDependenciesResult) {
-                this.updateItemViewsWithInboundDependencies(items.map(item => itemList.getItemView(item) as ArchiveCheckableItem));
-            }
+        const itemsAddedHandler = (
+            items: ContentSummaryAndCompareStatus[],
+            itemList: ListBox<ContentSummaryAndCompareStatus>
```
edloidas: We already know what types of lists expected here. Replace that generic type with the exact one. You won't need to do manual type checks with `isInboundAwareViewOf` in the code later.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItem.tsx
edloidas: As I said, this will be done in Enonic UI. You don't need to change anything here, but keep in mind that this approach is not entirely correct and will be changed in the future.

edloidas (approved):

PR #9143, opened 2025-09-17
Enable automatic translation of content on language change #9140

### LINKED ISSUE #9140: Enable automatic translation of content on language change
Currently our Juke Translator kicks in and suggests to translate content only after Localise. We want to enable it also when a language is set on the content (either changed or set from nothing to something).

When "Apply" button is clicked inside the "Edit Settings" dialog and the value of language field has changed, perform exactly the same checks as for localisation (that the Juke Translator app is active and properly configured, that content has the fields that Translator supports etc.) and, if all conditions are met, open exactly the same confirmation dialog asking whether the content should be translated into the new language. The content should be saved after translation, just like after localisation.

Also, while we're on this, fix it so that the translation progress dialog briefly shows "_x of x completed_" before replacing x's with actual numbers - it's better to show something generic like "_Initiating translation..._" than "_x of x_".

Image

ashklianko:


alansemenov (changes requested):
It works as expected, but we need to resolve and show display name of the language ("norsk" instead of "no"):

image

alansemenov:
Here's what we show for the same language after localisation
image

alansemenov (approved):

PR #9164, opened 2025-09-23
Incorrect user shown for Created version in Version History #9161

### LINKED ISSUE #9161: Incorrect user shown for Created version in Version History
When we artificially create a "Created" version in Version History, we incorrectly resolve user name. It should be exactly the same as on "Created by" line of the "Details" widget

ashklianko:
Forcing fetching of a creator's display name

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/dialog/CompareContentVersionsDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/CompareContentVersionsDialog.ts
@@ -1,8 +1,10 @@
+import {DefaultErrorHandler} from '@enonic/lib-admin-ui/DefaultErrorHandler';
 import {Body} from '@enonic/lib-admin-ui/dom/Body';
 import {DivEl} from '@enonic/lib-admin-ui/dom/DivEl';
 import {Element} from '@enonic/lib-admin-ui/dom/Element';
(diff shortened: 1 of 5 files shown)


modules/lib/src/main/resources/assets/js/app/dialog/CompareContentVersionsDialog.ts
```diff
@@ -322,8 +328,16 @@ export class CompareContentVersionsDialog
...
+        const getContentVersions = new GetContentVersionsRequest(this.content.getContentId()).sendAndParse();
+        const getPrincipalByKey = new GetPrincipalsByKeysRequest([this.content.getContentSummary().getCreator()]).sendAndParse().catch(
+            (e) => {
+                DefaultErrorHandler.handle(e);
+                return [];
+            });
+
+        return Q.all([getContentVersions, getPrincipalByKey]).spread((result: GetContentVersionsResult, principals: Principal[]) => {
             this.versions = result.getContentVersions();
+            this.creatorDisplayName = principals[0]?.getDisplayName() ?? this.content.getContentSummary().getCreator().toString();
```
alansemenov: Could assign `this.content.getContentSummary().getCreator()` to a const to avoid calling it twice, on lines 332 and 340.

alansemenov (approved):

PR #9170, opened 2025-09-24
Enonic UI: Create a WorkflowContentIcon #9040

### LINKED ISSUE #9040: Enonic UI: Create a WorkflowContentIcon
We need to render an icon for content with a workflow:
- Wrap the icon from https://github.com/enonic/app-contentstudio/issues/9041
- Display a workflow icon in the top right corner. There are 4 possible states:
  - Ready
  - In Progress
  - Invalid
  - (No workflow data) - just show nothing

Examples:

Image
Image

skjulestad:
Added WorkflowBadge.tsx
Slimmed down list-style
Wrapped ContentIcon.tsx with WorkflowBadge.tsx in:
ContentItem.tsx
ContentItemCheckable.tsx


modules/lib/src/main/resources/assets/js/app/ui2/list/ContentIcon.tsx
```diff
@@ -78,7 +78,7 @@ export const ContentIcon = ({
     if (canShowImage && src) {
         // 2x size for better quality
         return <Image
-            className={isImageType ? 'object-cover' : 'object-contain'}
+            className={`w-[24px] h-[24px] ${isImageType ? 'object-cover' : 'object-contain'}`}
```
edloidas: No custom pixel classes. It's `w-6`. I doubt that  `w-[24px]` will be used elsewhere, but it'll create an additional class in resulted CSS file.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+        status === WorkflowStateStatus.IN_PROGRESS &&
+        cn(
+            'text-[var(--color-warn)] [&>circle]:fill-current [&>path]:stroke-white [&>line]:stroke-white',
+            '[&>line:nth-of-type(1)]:scale-y-[1] [&>line:nth-of-type(1)]:-translate-y-[0.25px]',
+            '[&>line:nth-of-type(2)]:translate-y-[0.5px] [&>line]:[stroke-linecap:round]'
+        ),
+
+        status === WorkflowStateStatus.INVALID &&
+        'text-[var(--color-danger)] [&>circle]:fill-current [&>path]:stroke-white [&>path]:scale-[1.25]',
+    );
```
edloidas: Ideal place to use `cva` here. Might need to install this dependency.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+            : null;
+
+    if (!status) return <>{children}</>;
+
+    const Icon =
+        status === WorkflowStateStatus.READY
+        ? CircleCheck
+        : status === WorkflowStateStatus.IN_PROGRESS
+          ? CircleAlert
+          : CircleX;
```
edloidas: Move to a separate private component. Could name `<StatusIcon>` or something like that. Please, don't use ternaries here, use switch.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+    if (!summary) return <>{children}</>;
+
+    const status: WorkflowStateStatus | null =
+        !summary.isValid()
+        ? WorkflowStateStatus.INVALID
+        : summary.isReady()
+          ? WorkflowStateStatus.READY
+          : summary.isInProgress()
+            ? WorkflowStateStatus.IN_PROGRESS
+            : null;
```
edloidas: Move outside. Create a file `src/assets/js/app/util/content.ts` and name it `calcWorkflowStateStatus(summary: ContentSummary): WorkflowStateStatus`. Use if-else inside.
> What an awful name `WorkflowStateStatus`...

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+    summary?: ContentSummary | null;
+    children?: ReactNode;
+    overlaySize?: number;
+};
+
+export function WorkflowBadge({
+                                  summary,
+                                  children,
+                                  overlaySize = 12,
+                              }: Props): JSX.Element | null {
```
edloidas: You always return `JSX.Element`, not `null`.

alansemenov (commented):
As far as I can see, `overlaySize` already has a default value of 12, so I don't see why we need to pass it every time. Same with `size` for ContentIcon - we are showing it in hundreds of places in UI, so we should stick to default values without having to pass them every time.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
+import type {JSX, ReactNode} from 'react';
+import type {ContentSummary} from '../../content/ContentSummary';
+import {WorkflowStateStatus} from '../../wizard/WorkflowStateManager';
+import {CircleCheck, CircleAlert, CircleX} from 'lucide-react';
+import {cn} from '@enonic/ui';
+
+
+type Props = {
+    summary?: ContentSummary | null;
+    children?: ReactNode;
```
edloidas: - Accept `status` instead of `summary`. Don't make it optional, but allow `null`. 
- `children` must not be optional here, otherwise this component makes no sense.

We should have our props as accurate and minimal as possible. We also should avoid passing objects, especially created with classes, as they tend to change (the reference), while keeping the primitive inside intact. This will lead to unnecessary recalculations and redraws. Also hurt your memo from previous files.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItemCheckable.tsx
```diff
@@ -22,10 +23,15 @@ const ContentItemCheckableComponent = ({
     const label = String(content.getPath());
     const contentType = String(content.getType());
     const url = content.getContentSummary().getIconUrl();
+    const summary = content.getContentSummary();
```
edloidas: Calculate status here and pass it instead:
```
const status = calcWorkflowStateStatus(content.getContentSummary());
```
> The details described in comments to other files.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+        status === WorkflowStateStatus.INVALID &&
+        'text-[var(--color-danger)] [&>circle]:fill-current [&>path]:stroke-white [&>path]:scale-[1.25]',
+    );
+
+    return (
+        <span className="relative inline-flex items-center h-7.5 w-7.5">
+      {children}
+            <span className="absolute -top-0.5 -right-0.5">
+        <Icon className={iconClass} strokeWidth={2} size={overlaySize} />
+      </span>
```
edloidas: Should we really wrap it in `<span>`? I think it's okay to apply classes to the icon directly.

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
edloidas: Since the icon can be moved to a separate component, `cva` can be applied there, not in this component.

modules/lib/src/main/resources/assets/js/app/ui2/list/ContentItemCheckable.tsx
```diff
@@ -22,10 +23,15 @@ const ContentItemCheckableComponent = ({
...
 
     const Icon = useMemo(
-        () => <ContentIcon contentType={contentType} url={url} size={24} />,
-        [contentType, url],
+        () => (
+            <WorkflowBadge summary={summary} overlaySize={12}>
+                <ContentIcon contentType={contentType} url={url} size={24} />
+            </WorkflowBadge>
+        ),
```
edloidas: I think that this one block will be reused more than once. Then there is a "rule": If the component used 3+ times, and they are coupled together (they both operate with content) and not just stateless UI blocks, then it OK to compose them in one component (usually keeping component separate and combine them in parent better).

- Let's move this logic to `<WorkflowContentIcon>`.
- Move `<WorkflowBadge>` and all other code there too. We don't need it as a separate component, just some inner code. I think you don't even need badge itself, you can use `<ContentIcon>` instead of abstract children directly after all those changes described bellow in other comments.

edloidas (approved):

modules/lib/src/main/resources/assets/js/app/ui2/list/WorkflowBadge.tsx
skjulestad: Unavoidable.
I need root element for ContentIcon and statusIcon for classnames.

PR #9189, opened 2025-09-30
content.update - allow to update page #11256

vbradnitski:


modules/rest/src/main/java/com/enonic/xp/app/contentstudio/rest/resource/content/ContentResource.java
```diff
@@ -1606,7 +1470,7 @@ private UpdateContentParams prepareUpdateContentParams( final Content versionedC
             edit.data = versionedContent.getData();
             edit.displayName = versionedContent.getDisplayName();
             edit.extraDatas = versionedContent.getAllExtraData();
-            edit.page = versionedContent.getPage();
+            edit.page = new EditablePage(versionedContent.getPage());
```
rymsha: check for null

rymsha (changes requested):

rymsha (approved):

PR #9190, opened 2025-09-30
Normalize gradlew.bat line endings to CRLF

edloidas:
Looks like updated `gradlew.bat` file had incorrect line endings (LF), even though we are now forcing `CRLF` for all bat files via `.gitattributes`. 
It resulted in permanent "changed" status in git on Mac, that checking out, because:
- File is stored as LF in the repo
- On checkout:
  - **Mac:** Git rewrites the file's line endings to CRLF.
  - **Windows:** always CRLF in the working copy (= no problem).
- On Mac (or Linux with LF default) Git will mark the file as modified.


rymsha (approved):

edloidas:
Fixed in master in different commit. No longer has `LF` in repo.

PR #9192, opened 2025-09-30
Fix build problems on Windows #9191

### LINKED ISSUE #9191: Fix build problems on Windows
We need to do some R&D and fix build performance on Windows with Microsoft Defender.

edloidas:
Added jsondiffpatch as direct dependency of modules/app. In some scenarios, jsondiffpatch is not found in lib-contentstudio during the build time. This is a straightforward attempt to make it available directly. 
Added webpack build caching.
Added JVM args to prevent failing during heavy loads. 
Enabled file system watching to speed up changes detection. Added various flags for pnpm:
  - package-import-method=copy: copy packages from the store instead of hardlinking.
  - side-effects-cache=true: enable caching of build artifacts from postinstall scripts.
  - node-linker: Uses a hoisted node_modules layout (fewer/deeper symlinks vs isolated).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,4 +26,5 @@ testing/allure-results
 .enonic
 .eslintcache
 .bin
-
+.pnpm-store
+**/.webpack-cache
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,6 @@
+# Performance optimizations for Windows
+package-import-method=copy
+side-effects-cache=true
+
+# Reduce defender scanning overhead
+node-linker=hoisted
--- a/modules/app/package.json
+++ b/modules/app/package.json
@@ -25,6 +25,7 @@
     "jquery": "^3.7.1",
     "jquery-simulate": "^1.0.2",
     "jquery-ui": "^1.14.1",
+    "jsondiffpatch": "^0.7.3",
     "lib-contentstudio": "workspace:*",
     "nanostores": "^0.11.4",
     "q": "^1.5.1"
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -26,6 +26,9 @@ importers:
       jquery-ui:
         specifier: ^1.14.1
         version: 1.14.1
+      jsondiffpatch:
+        specifier: ^0.7.3
+        version: 0.7.3
       lib-contentstudio:
         specifier: workspace:*
         version: link:.xp/dev/lib-contentstudio
--- a/modules/app/webpack.config.js
+++ b/modules/app/webpack.config.js
@@ -12,6 +12,13 @@ const path = require('path');
 const isProd = process.env.NODE_ENV === 'production';
 
 module.exports = {
+    cache: {
+        type: 'filesystem',
+        cacheDirectory: path.resolve(__dirname, '.webpack-cache'),
+        buildDependencies: {
+            config: [__filename]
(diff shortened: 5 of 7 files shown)


gradle.properties
```diff
@@ -13,3 +13,5 @@ version=6.0.0-SNAPSHOT
 org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.daemon=true
+org.gradle.jvmargs=-Xms512m -Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -XX:+UseStringDeduplication
+org.gradle.vfs.watch=true
```
rymsha: These settings are unrelated to node problems on windows.
In fact, most of these may make performance worse.

rymsha (changes requested):

gradle.properties
edloidas: _You are absolutely right._

As for the "jvmargs", I just thought it would be great to increase the defaults a bit, since I sometimes (including during the latest test builds) run out of memory and have to either kill my whole java process or add those args to the build manually. I also remember that some of the guys (Pavel for sure) have custom build scripts with increased memory for the same reason.

Not sure about `org.gradle.vfs.watch`. I can't give you metrics, but at least it felt a bit faster after many different builds.

edloidas: Both can be removed. But just try building the app in various configurations (`clean build` with no node_modules and .xp; `clean build` with existing node_modules; regular `build`, etc.).

rymsha: @ In depth 
Xms 512 - has no effect. So much memory allocation happens instantly 
Xmx 2G - is machine dependent. Machines with insane number of cores (virtual) happen to increase memory demands, but on smaller machines just causes out of memory
HeapDump - just slows down shutdown
StringDeduplication - if works (likely not) is a trade off between CPU usage and memory usage. And we don't win here.
vfs.watch is enabled by default - so it is a placebo effect.


Remove, please

edloidas: Removed.

Double-checked the docs and _my_ local configuration. Still want to calrify some things:
- Heap Dump: [Docs](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/clopts001.html) say, that there is no overhead in running with this option. But yeah, not really needed in gradle props, I guess.
- `-Xms`: My -Xms256m (InitialHeapSize=268435456). Isn't it non-optimal? Build in CS usually consume at elast x1.5 of this?
- `-Xmx`: My -Xmx512m / -XX:MaxMetaspaceSize=384m . Isn't it the top limit? 512m sound really like really small value, that sometimes just not enough and results in OOM. Why can't we set the reasonable top value?

rymsha: The principle is very simple - don't configure anything that does not give you any benefit _now_.

Heap dump has no overhead until it does - when jvm actually halts with OOME - in this case it takes a few extra seconds to write a dump file. We do use this option in production. But in build system it gives no value - just noise. 

We also use Xms in production - to make xms and xmx equal. Again in build system this extra config is just noise. You save ~10 ms.

MetaSpace is a  fine tuning config. If metadata of the build does not fit into 384 mb - we have a problem somewhere else.

Xmx is the only "reasonable" option. But it is machine dependent and should be set locally. I consider this setting out of this PR scope.

rymsha (approved):

PR #9232, opened 2025-10-10
revert must not revert child order #9231

### LINKED ISSUE #9231: revert must not revert child order
child order is not data. it is closer to permissions or name (path) that we don't revert.

ashklianko:


ashklianko:
Made sorted versions non-revertable 
image

alansemenov (approved):

PR #9236, opened 2025-10-14
Enonic UI: Create ProjectItem #9223

### LINKED ISSUE #9223: Enonic UI: Create ProjectItem
Create a basic item to render Project data. 
Use `ListItem` for this.
Look at `ContentItemCheckable` to see how it may look.
Don't forget to keep `ListItem` props, so that it can be set to `selected` in some scenarios, when placed inside lists.

edloidas:


modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+  className,
+  href,
+  onClick,
+  role,
+  ...rest
+}: ProjectItemProps): ReactElement {
+  return (
+    <ListItem
+      selected={selected}
+      className="p-0.75 !outline-none"
```
edloidas: Why do you need to disable outline here. More importantly, with `!important` keyword?

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,153 @@
...
+    // Fallbacks: common focusable elements inside the item
+    const fallback = root.querySelector('a[href], button, input, [tabindex]:not([tabindex="-1"])');
+    if (fallback instanceof HTMLElement) {
+      fallback.focus();
+      return true;
+    }
+
+    return super.giveFocus();
+  }
+}
```
edloidas: Don't need this, actually. We'll remove the whole elements, that render this component.

edloidas (commented):
Not actually what I have in mind. 
For the Project Dialog you'll use `<Listbox>` component, that handles selection. You will need to render `<ProjectItem>` there as a child and that's it.

The idea for this component it to make it render:
- Icon with the language
- Project name
- Project description
- Project language 
- Have a fallback to default icon, if we are using default project.

It doesn't need to be clickable or handle selection state itself. Just pass selected from parent later. So, i think it would be enough to implement Icon component and just use `ListItem.Content` with that icon, title, optional description and optional language tag after title. 

**Basic view in dialog:**
Image

**Same item will probably be used in Settings:**
Image

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+export type ProjectIconProps = {
+    projectName: string;
+    language?: string;
+    icon?: ReactNode;
+    hasIcon?: boolean;
+    isLayer?: boolean;
+    className?: string;
+};
+
+export default function ProjectIcon({projectName, language, hasIcon, isLayer, className}: ProjectIconProps): ReactElement {
```
edloidas: Don't export it as default.

edloidas: Also, prefer arrow function here. And set a displayName.

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    try {
+        const fallbackClass = isLayer
+                              ? ProjectIconUrlResolver.getDefaultLayerIcon()
+                              : ProjectIconUrlResolver.getDefaultProjectIcon();
+        return <i className={`h-6 w-6 rounded ${className ?? ''} ${fallbackClass}`} aria-hidden="true"/>;
+    } catch {
+        // Safe fallback if resolver unexpectedly fails
+        return <i className="h-6 w-6 rounded" aria-hidden="true"/>;
+    }
```
edloidas: What code exactly potentially throws an exception that you need to wrap in try-catch?

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+    icon?: ReactNode;
+    hasIcon?: boolean;
+    isLayer?: boolean;
+    className?: string;
+};
+
+export default function ProjectIcon({projectName, language, hasIcon, isLayer, className}: ProjectIconProps): ReactElement {
+    try {
+        if (hasIcon) {
+            const url = new ProjectIconUrlResolver().setProjectName(projectName).setTimestamp(new Date().getTime()).resolve();
```
edloidas: I'm most certain that you can write smaller and better URL resolver without assertion for missing arguments. Just like with previous resolver you made for ContentIcon.

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+    return (
+        <ListItem
+            {...rest}
+        >
+            <ListItem.Left>
+                <ProjectIcon
+                    projectName={projectName}
+                    language={language}
+                    hasIcon={hasIcon}
+                />
```
edloidas: I think we have an icon in `ListItem.DefaultContent`. Are there reasons for not using it? Different sizes?

modules/lib/src/main/resources/assets/js/app/ui2/util/url.ts
```diff
@@ -1,10 +1,20 @@
...
+import {UrlHelper} from '../../util/UrlHelper';
 
 interface ImageUrlParams {
     size?: number | null;
     crop?: boolean | null;
 }
 
+export interface ProjectIconUrlOptions {
+    ts?: number;
+}
```
edloidas: Not used anymore

modules/lib/src/main/resources/assets/js/app/ui2/util/url.ts
```diff
@@ -1,10 +1,20 @@
...
+
 export function createImageUrl(url: string, params: ImageUrlParams): string {
     return `${UriHelper.trimUrlParams(url)}?${UriHelper.encodeUrlParams(params)}`;
 }
+
+export const resolveProjectIconUrl = (projectName?: string): string | undefined =>
+    projectName
+        ? `${UrlHelper.getCmsRestUri('project/icon/')}${encodeURIComponent(projectName)}`
+        : undefined;
```
edloidas: **Future tip:** this won’t be validated properly due to non-strict behavior in our current config, but you should do it like this:

```diff
- (projectName?: string)
+ (projectName: string | undefined | null)
```

When you use `?`, you mark it as an *optional parameter*, allowing it to be called without any arguments at all — like `resolveProjectIconUrl()`.
When you explicitly add `undefined` to the type, you’re essentially saying:

_“I need this value to be present, but I allow it to be `undefined`; I’ll handle the validation inside.”_

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+export class ProjectItem
+    extends LegacyElement<typeof ProjectItemView, ProjectItemProps> {
+    private project: Project;
+
+    constructor(project: Project) {
+        const displayName = project.getDisplayName() || project.getName();
+        const name = project.getName();
+        const language = project.getLanguage();
+        const icon = project.getIcon();
+        console.log(project);
```
edloidas: Not needed

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <ListItem.Content className="flex-1 min-w-0 grid-cols-[auto_1fr] gap-2.5 items-center grid">
+                <ProjectIcon
+                    projectName={projectName}
+                    language={language}
+                    hasIcon={hasIcon}
+                />
+                <div>
+                <h3 className="text-base leading-5.5">
+                    {label}
+                    {language ? <span className="text-sm text-subtle">({language})</span> : null}
```
edloidas: You can actually use `&&` instead of `? :`:
```diff
- {language ? <span className="text-sm text-subtle">({language})</span> : null}
+ {language && <span className="text-sm text-subtle">({language})</span>}
```

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+    try {
+        if (hasIcon) {
+            const url = resolveProjectIconUrl(projectName);
+            if (url) {
+                return <img src={url} alt="" draggable={false} className="h-6 w-6 rounded-full bg-center object-contain"/>;
+            }
+        }
+    } catch {
+        // fall through to flag/default
+    }
```
edloidas: Since you simplified `esolveProjectIconUrl` and moved the assertion out of it, it will no longer throw errors. Therefore, most of the try-catch blocks here are just adding visual noise.

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+                                                            language,
+                                                            hasIcon,
+                                                            isLayer,
+                                                            className
+                                                        }) => {
+    try {
+        if (hasIcon) {
+            const url = resolveProjectIconUrl(projectName);
+            if (url) {
+                return <img src={url} alt="" draggable={false} className="h-6 w-6 rounded-full bg-center object-contain"/>;
```
edloidas: It's always good to have `alt` text. Maybe set it to project name?

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
edloidas: And try to avoid nested closures. You can write just one check for url and move `hasIcon` to ternary operator:
```ts
const url = hasIcon ? resolveProjectIconUrl(projectName) : null;
if (url) {
    return <img />;
}
```
Much cleaner

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+                />
+            );
+        }
+    } catch {
+        // fall through to default icon
+    }
+        const fallbackClass = isLayer
+                              ? ProjectIconUrlResolver.getDefaultLayerIcon()
+                              : ProjectIconUrlResolver.getDefaultProjectIcon();
+        return <i className={`h-6 w-6 rounded text-xl text-center ${className ?? ''} ${fallbackClass}`} aria-hidden="true"/>;
```
edloidas: Use `cn` instead of those manual concatenations. It's cleaner, and you will merge tailwind classes properly

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectIcon.tsx
```diff
@@ -0,0 +1,65 @@
...
+                return <img src={url} alt="" draggable={false} className="h-6 w-6 rounded-full bg-center object-contain"/>;
+            }
+        }
+    } catch {
+        // fall through to flag/default
+    }
+
+    try {
+        const lang = language ? String(language).toLowerCase() : '';
+        if (lang) {
```
edloidas: You can reverse the language check and return the final fallback here to avoid nested scopes (if-if-if-...) — at least by one level. Always aim for a flatter structure.
If you find yourself with two or more nested if statements, consider reorganizing your code using early returns, ternaries, reordering conditions, or moving some logic into separate functions.

modules/lib/src/main/resources/assets/js/app/ui2/list/ProjectItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+    hasIcon?: boolean;
+    href?: string;
+} & Omit<ListItemProps, 'children'>;
+
+export function ProjectItemView({
+                                    label,
+                                    projectName,
+                                    language,
+                                    hasIcon,
+                                    selected,
```
edloidas: Selected is extracted from props, but never assigned to `<ListItem>`. That's why you'll never see selection styles on item. Either keep it in `...props`, or set it manually.

edloidas (commented):
I updated the code. Please, look at the comments to previous commit.

PR #9263, opened 2025-10-21
Enonic UI: ConfirmationDialog

### LINKED ISSUE #9226: Enonic UI: ConfirmationDialog
Confirmation dialog is a modal window shown before performing potentially dangerous actions — like deleting content or removing comments.

### Rationale
- Start with minimal functionality and expand it later to speed up development.
- While the original confirmation dialog is located in the `lib-admin-ui` repo, implement it here for easier iteration.
- Follow a **composite approach** rather than customizing it through numerous props — this ensures better maintainability and flexibility.
- We don’t need a `LegacyElement` wrapper since it won’t be used within legacy code.

### References
- Refer to the designs in the attached screenshots and the Figma project.
- Review how existing dialogs are implemented in `lib-admin-ui` and `app-contentstudio`:
    - [`ConfirmationDialog`](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/js/ui/dialog/ConfirmationDialog.ts) — a basic dialog invoked from various parts of Content Studio.
    - [`ModalDialogWithConfirmation`](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/js/ui/dialog/ModalDialogWithConfirmation.ts) — a dialog that integrates confirmation as a secondary step. We don’t need to follow this pattern, but it’s useful to understand its usage for potential replacement.
    - [`ConfirmValueDialog`](https://github.com/enonic/app-contentstudio/blob/master/modules/lib/src/main/resources/assets/js/app/remove/ConfirmValueDialog.ts) — implementation of the previous dialog within Content Studio. This one will likely need modification.

### Things to Consider
- For easier testing, it may be best to start implementing this component inside `@enonic/ui` first.

### Implementation

> [!IMPORTANT]  
> This is not a step-by-step guide — it’s a functional checklist ordered logically. Make sure all essential parts are implemented.

1. Create a `ConfirmationDialog.tsx` file containing all parts.
2. We need to follow the same pattern and export everything under `ConfirmationDialog` scope. The difference will be that we will provide _some_ new comonents:
```ts
export const ConfirmationDialog = {
  // Just reexport the same part
  Root: Dialog.Root,
  // ...
  // We have 2 parts that must be changed
  Content: ConfirmationDialogContent,
  Footer: ConfirmationDialogFooter,
};
```
3. Start with implementing the `<ConfirmationDialogProvider>` and context with hooks:
    - We should keep it here, in the same file. We won't move it outside, like we did in UI library.
    - Context will have `unlocked: boolean` and `setUnlocked: (unlocked: boolean) => void` properties for now. 
    - We need those properties to control buttons in the footer and validation in more complex dialogs.
4. Create `<ConfirmationDialog.Content>`, that will render `<Dialog.Content>` and wrap it with `<ConfirmationDialogProvider>`. That way we will be able to share state across different parts (`Body`, Footer`) and automatically unmount it on dialog close.
    - We must also accept `defaultUnlocked` (`true` by default) and pass it to context.
5. Create `<ConfirmationDialog.Footer>` with 2 buttons: **Cancel** and **Confirm**.
    - Make those buttons lables configurable via properties `cancelLabel` and `confirmLabel`.
    - Make it possible to add custom handlers `onCancel` and `onConfirm` to those buttons via props.
8. Use the `useI18n()` hook to fetch localized text from `phrases.properties`.

**UI Behavior:**
- Autofocus should follow the dialog default — focus the first non-destructive element (like **Cancel**) or an input if present. Will probably be focused by default.

**Testing steps:**
1. Select 2 or more items with "Work in progress" status (yellow icon)  in the grid.  
2. Click **Mark as Ready** button in the right side of the Toolbar.  
4. The Confirmation dialog should appear.

### Examples

**Basic confirmation**
Image

skjulestad:
Adds basic confirmation dialog with content and footer.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
+++ b/modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
@@ -1,0 +1,95 @@
+import {Button, Dialog, cn} from '@enonic/ui';
+import type {ComponentPropsWithoutRef, ReactElement, ReactNode} from 'react';
+import {createContext, useContext, useMemo, forwardRef, useState} from 'react';
+import {i18n} from '@enonic/lib-admin-ui/util/Messages';
+
+//
+// * ConfirmationDialogContext
+//
+
+type ConfirmationDialogContextValue = {
+    submitEnabled: boolean;
+    setSubmitEnabled: (submitEnabled: boolean) => void;
+};
+
+const ConfirmationDialogContext = createContext<ConfirmationDialogContextValue | null>(null);
+
+export const useConfirmationDialog = (): ConfirmationDialogContextValue => {
+    const ctx = useContext(ConfirmationDialogContext);
+    if (!ctx) {
+        return {submitEnabled: true, setSubmitEnabled: () => void 0};
+    }
+    return ctx;
+};
+
+type ConfirmationDialogProviderProps = {
+    defaultSubmitEnabled?: boolean;
+    children?: ReactNode;
+};
+
+const ConfirmationDialogProvider = ({defaultSubmitEnabled = true, children}: ConfirmationDialogProviderProps): ReactElement => {
+    const [submitEnabled, setSubmitEnabled] = useState<boolean>(defaultSubmitEnabled);
+
+    const value = useMemo<ConfirmationDialogContextValue>(() => ({submitEnabled, setSubmitEnabled}), [submitEnabled]);
+
+    return <ConfirmationDialogContext.Provider value={value}>{children}</ConfirmationDialogContext.Provider>;
+};
+
+//
+// * ConfirmationDialogContent
+//
+
+export type ConfirmationDialogContentProps = {
+    defaultSubmitEnabled?: boolean;
+    className?: string;
+    children?: ReactNode;
+} & ComponentPropsWithoutRef<typeof Dialog.Content>;
+
+const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogContentProps>(
+    ({defaultSubmitEnabled = true, className, children, ...props}, ref): ReactElement => {
+        return (
+            <Dialog.Content ref={ref}
+                            className={cn('gap-3 min-w-170 max-w-fit', className)} {...props}>
+                <ConfirmationDialogProvider defaultSubmitEnabled={defaultSubmitEnabled}>{children}</ConfirmationDialogProvider>
+            </Dialog.Content>
+        );
+    }
+);
+ConfirmationDialogContent.displayName = 'ConfirmationDialog.Content';
+
+//
+// * ConfirmationDialogFooter
+//
+
+export type ConfirmationDialogFooterProps = {
+    onCancel?: () => void;
+    onSubmit?: () => void;
+    className?: string;
+} & ComponentPropsWithoutRef<'footer'>;
+
+const ConfirmationDialogFooter = ({
+    onCancel,
+    onSubmit,
+    className,
+    ...props
+}: ConfirmationDialogFooterProps): ReactElement => {
+    const {submitEnabled} = useConfirmationDialog();
+
+    return (
+        <Dialog.Footer {...props}>
+            <Dialog.Close asChild>
+                <Button label={i18n('action.cancel')} variant='outline' onClick={onCancel}/>
+            </Dialog.Close>
+            <Dialog.Close asChild>
+                <Button label={i18n('action.submit')} variant='solid' onClick={onSubmit} disabled={!submitEnabled}/>
+            </Dialog.Close>
+        </Dialog.Footer>
+    );
+};
+ConfirmationDialogFooter.displayName = 'ConfirmationDialog.Footer';
+
+export const ConfirmationDialog = {
+    ...Dialog,
+    Content: ConfirmationDialogContent,
+    Footer: ConfirmationDialogFooter,
+};
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -275,8 +275,8 @@ importers:
   .xp/dev/lib-contentstudio:
     dependencies:
       '@enonic/ui':
-        specifier: ~0.12.0
-        version: 0.12.0(@radix-ui/react-slot@1.2.3(@types/react@19.2.2)(react@19.2.0))(focus-trap-react@11.0.4(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(lucide-react@0.545.0(react@19.2.0))(preact@10.27.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0)
+        specifier: ~0.12.1
+        version: 0.12.1(@radix-ui/react-slot@1.2.3(@types/react@19.2.2)(react@19.2.0))(focus-trap-react@11.0.4(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(lucide-react@0.545.0(react@19.2.0))(preact@10.27.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0)
       class-variance-authority:
         specifier: ^0.7.1
         version: 0.7.1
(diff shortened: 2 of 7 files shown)


modules/lib/src/main/resources/assets/js/app/browse/action/MarkAsReadyContentAction.ts
```diff
@@ -37,8 +36,49 @@ export class MarkAsReadyContentAction
...
+                                onCancel: onNo,    // “No”
+                            })
+                        )
+                    )
+                ),
+                mount
+            );
+
         }
+
```
edloidas: You can rename the file to *.tsx` and just write the JSX without need of calling Preact's render functions.

modules/lib/src/main/resources/assets/js/app/browse/action/MarkAsReadyContentAction.ts
```diff
@@ -37,8 +36,49 @@ export class MarkAsReadyContentAction
...
+                            h(ConfirmationDialog.DefaultHeader, {
+                                title: i18n('dialog.confirm.title'),
+                                withClose: true,
+                            }),
+                            h(ConfirmationDialog.Body, null,
+                                h('p', {className: 'text-base'}, i18n('dialog.markAsReady.question')),
+                            ),
+                            h(ConfirmationDialog.Footer, {
+                                onConfirm: onYes,  // “Yes”
+                                onCancel: onNo,    // “No”
```
edloidas: Prefer using default prop names and avoid meanless comments:
```ts
{
    onConfirm,
    onCancel,
})
```

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
+import {Button, Dialog, cn} from '@enonic/ui';
+import type {ComponentPropsWithoutRef, ReactElement, ReactNode} from 'react';
+import {createContext, useCallback, useContext, useMemo, useState} from 'react';
+import {i18n} from '@enonic/lib-admin-ui/util/Messages';
+import {forwardRef} from 'react';
```
edloidas: Organize imports. There are 3 react imports.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+//
+// Public API: re-export Dialog parts; override Content/Footer
+//
+
+type ConfirmationDialogNS =
+    Omit<typeof Dialog, 'Content' | 'Footer'> & {
+    Content: typeof ConfirmationDialogContent;
+    Footer: typeof ConfirmationDialogFooter;
+};
```
edloidas: Pretty much meaningless type. Remove.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+import {createContext, useCallback, useContext, useMemo, useState} from 'react';
+import {i18n} from '@enonic/lib-admin-ui/util/Messages';
+import {forwardRef} from 'react';
+
+//
+// Confirmation dialog context
+//
+
+type ConfirmationDialogContextValue = {
+    unlocked: boolean;
```
edloidas: Now I think it would be better to make it more clear: `unlocked` -> `submitEnabled`. We'll need to rename all other related props too.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const [unlocked, setUnlocked] = useState<boolean>(defaultUnlocked);
+
+    const value = useMemo<ConfirmationDialogContextValue>(() => ({unlocked, setUnlocked}), [unlocked]);
+
+    return <ConfirmationDialogContext.Provider value={value}>{children}</ConfirmationDialogContext.Provider>;
+};
+
+//
+// Content: wraps Dialog.Content with provider to keep shared state
+//
```
edloidas, line 40: Remove the description from the separator comments. It must be either part of another JSDoc comment, or removed completely.
Please start the line with the text with `// * <some text>`. `*` is needed here for some IDEs and plugins to treat it as a special kind (info) of comment, visually distinct.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+    defaultUnlocked?: boolean;
+    className?: string;
+    children?: ReactNode;
+} & ComponentPropsWithoutRef<typeof Dialog.Content>;
+
+const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogContentProps>(
+    ({defaultUnlocked = true, className, children, ...props}, ref): ReactElement => {
+        return (
+            <Dialog.Content ref={ref}
+                            className={cn('!outline-none [&_button]:!outline-none gap-3 min-w-170 max-w-fit', className)} {...props}>
```
edloidas: Still keeping fixes for outline de to conflicts with legacy styles? Just ignore it for now.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+        },
+        [onCancel]
+    );
+
+    const handleSubmit = useCallback<ButtonClick>(
+        (e) => {
+            onConfirm?.(e);
+        },
+        [onConfirm]
+    );
```
edloidas: They basically add zero logic right now. `useCallback` is useless, and those functions are fully compatible with `onClick`, so pass them directly to buttons.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+//
+
+type ButtonClick = NonNullable<ComponentPropsWithoutRef<'button'>['onClick']>;
+
+export type ConfirmationDialogFooterProps = {
+    className?: string;
+    cancelLabel?: ReactNode;
+    submitLabel?: ReactNode;
+    onCancel?: ButtonClick;
+    onConfirm?: ButtonClick;
```
edloidas: Just a small cosmetic request. We should now follow this ordering (this is defined in Enonic UI rules, i'll add them to this project soon too):
```ts
{
    // ...
    className?: string;      // ✅ Second-to-last
    children?: ReactNode;  // ✅ Last
}
```

edloidas: Also, maybe click handlers can be simplified? Do we actually need for the consumer element to handle mouse events? I guess `() => void` instead of `ButtonClick` can be enough.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+} & ComponentPropsWithoutRef<'footer'>;
+
+const ConfirmationDialogFooter = ({
+                                      className,
+                                      cancelLabel = i18n('action.cancel'),
+                                      submitLabel = i18n('action.submit'),
```
edloidas: 1. Move initialization to function's body.
2. Use `useI18n` hook instead of that function. Even though the hook is just alias, it will be replaced with an actual hook in the future.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+        },
+        [onConfirm]
+    );
+
+    return (
+        <Dialog.Footer {...props}>
+            <Dialog.Close asChild>
+                <Button type='button' variant='outline' onClick={handleCancel}>
+                    {cancelLabel}
+                </Button>
```
edloidas: - `type='button'` is meaningless, as it's already set in Enonic UI `<Button>`.
- Make `*Label` a `string`. Then you can pass it via `label` property instead of a child. Then it'll automatically set proper `aria-label` value too.

modules/lib/src/main/resources/assets/js/app/browse/action/MarkAsReadyContentAction.ts
```diff
@@ -37,8 +36,49 @@ export class MarkAsReadyContentAction
...
         } else {
-            this.confirmDialog.setYesCallback(() => void this.markAsReadyAndPublish(contentToMarkAsReady, content)).open();
+            // Mount a UI2 ConfirmationDialog on demand
+            const mount = document.createElement('div');
+            document.body.appendChild(mount);
+
+            const cleanup = (): void => {
+                render(null, mount);
+                mount.remove();
+            };
```
edloidas: I fully understand why you did this. I just think we can do better, even for a temporary solution. What if you export `class ConfirmationDialogElement extends LegacyElement` and use that element here? That dialog will already present the root to render dialog. You can implement `open` and `close` functions, like in old modals, and just set the `open` state instead.

edloidas (changes requested):

edloidas (approved):

PR #9270, opened 2025-10-22
Enonic UI: Create Sidebar component #9267

### LINKED ISSUE #9267: Enonic UI: Create Sidebar component
A vertical navigation panel that appears on the left side of the screen in browse view, providing quick access to Content Studio features and widgets.

---

## Rationale

- **Fresh implementation:** This is a foundational UI component that needs to be built from scratch to align with the new Content Studio 6 architecture.
- **Bridge to legacy code:** We're using `LegacyElement` as a wrapper to ensure smooth integration with existing code while allowing dynamic widget button registration.
- **New code structure:** This component introduces our v6 folder structure (`modules/lib/src/main/resources/assets/js/v6`), establishing patterns for future Content Studio 6 development.

---

## References

- [`WidgetsSidebar.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/widget/WidgetsSidebar.ts) — existing implementation to be replaced
- Figma design

---

## Things to Consider

- **Styling approach:** Use Tailwind utility classes for all styling instead of traditional CSS.
- **i18n migration path:** Use `useI18n` hook (currently an alias for `i18n`) for all localized strings to prepare for future React hook migration.
- **Widget state management:** Current approach stores widgets in WidgetsToolbar — consider improving this architecture while keeping local state for now (passed via LegacyElement wrapper props).
- **Testing environment:** Start development in browse view where Sidebar will be most visible.
- **Future extensibility:** Design the `SidebarElement` API to support dynamic button addition from `AppWrapper`.

---

## Implementation

> [!IMPORTANT]  

1. **Create the component structure:**
   - Create `modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.ts`
   - `AppShell` is a new folder to contain all root-level view parts and soon-to-be renamed `AppWrapper`.
   - This establishes the v6 folder pattern for new code

2. **Build the Sidebar component:**
   - Use `<IconButton>` from Enonic UI for all interactive buttons
   - Top icon (Content Studio logo) should be an `IconButton` with no onClick handler yet
   - Render widget buttons dynamically based on passed configuration

3. **Implement the legacy wrapper:**
   - Create `SidebarElement` class extending `LegacyElement`
   - Expose API (some of the old methods) for `AppWrapper` to register new widget buttons dynamically
   - Bridge between React component and imperative API

4. **Apply styling:**
   - Use Tailwind utility classes for layout, colors, spacing
   - Match dark theme from design specifications
   - Ensure proper vertical stacking and spacing of buttons

5. **Localization setup:**
   - Use `useI18n` hook for all text
   - New text can be added to `phrases.properties` file
   - Replace existing `i18n` calls with the new hook

6. **Preserve legacy code:**
   - Keep existing `Widgets*` files intact for now — they'll be migrated later
   - Ensure backward compatibility with current widget registration

**UI Behavior:**

- Sidebar is always visible and cannot be collapsed (different from current behavior)
- All widget buttons must be keyboard-accessible via Tab navigation
- Button tooltips appear on hover/focus
- Active widget button should have visual indication

---

## Acceptance Criteria

- [x] Sidebar renders correctly in browse view with correct positioning and styling
- [x] Visual appearance matches Figma design (colors, spacing, dimensions)
- [x] All widget buttons are clickable and open corresponding views
- [x] Content Studio logo button is present (no action required yet)
- [x] Keyboard navigation works: Tab cycles through buttons, Enter/Space activates
- [x] `SidebarElement` API allows `AppWrapper` to add buttons dynamically
- [x] All text uses `useI18n`

---

## Testing Steps

1. Open Content Studio application
2. Navigate to browse view (should be default view on startup)
3. Locate the dark vertical Sidebar on the left edge of the screen
4. Verify Sidebar is visible without clicking hamburger menu (always-visible behavior)
5. Test keyboard navigation:
   - Press Tab to cycle through widget buttons
   - Press Enter or Space on focused button to activate
6. Click each widget button and verify corresponding view opens
7. Verify visual states: hover, focus, active button highlighting

---

## Examples

| New | Alt | Current |
|-|-|-|
| New Sidebar | Alt Sidebar Design | Current Sidebar |

---

## Out of Scope

- Implementation of widget selection/management UI in `AppWrapper`

---

## Technical Constraints

- Must extend `LegacyElement` for backward compatibility with existing codebase
- Widget state management remains in local component state (not global store yet)

reisfmb:


modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  handler = () => {},
+}: Props): React.ReactElement => {
+  const project = ProjectContext.get().getProject();
+  const version = "v." + CONFIG.getString("appVersion");
+  const mainWidgets = widgets.slice(0, -1);
+  const lastWidget = widgets.at(-1);
+
+  const SidebarHeader = () => (
+    <>
+      <div class="w-[32px] h-[32px]">
```
edloidas: 1. Don’t use custom sizing — we try to stick to native Tailwind sizes.
When you use values like `w-[32px]`, Tailwind generates a new utility class for that exact size, which is one more than needed since `w-8` already represents `32px` and is likely in use.
2. While not required, you can use the `size-*` utility instead of combining `w-*` and `h-*`.
3. `ProjectIcon` already uses `size-8` (32px).
4. There is no need to wrap `<ProjectIcon>` in a `<div>`.
6. The icon is missing `14px` of vertical padding (`py-3.5`) — add it via `className`.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+};
+
+const SidebarComponent = ({
+  name = Store.instance().get("application").getName(),
+  widgets = [],
+  activeWidgetId = "",
+  handler = () => {},
+}: Props): React.ReactElement => {
+  const project = ProjectContext.get().getProject();
+  const version = "v." + CONFIG.getString("appVersion");
```
edloidas: Designer is not right here. Drop the `.` after `v`. Versions are always written as `v0.0.0`

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const baseIcon = CircleQuestionMark;
+  const isWidgetActive = widgetKey === activeWidgetId;
+  const icon = SidebarElement.isMainWidget(widget) ? Pen : baseIcon;
+
+  if (widgetIconUrl) {
+    return (
+      <Tooltip value={widgetDisplayName} side="right">
+        <Button
+          className={cn(
+            "px-[13px]",
```
edloidas: Use natie Tailwind sizes instead: `px-3.25`. Read related comments in Sidebar Header bellow. Replace all px sizing.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+            isWidgetActive
+              ? "bg-surface-primary-selected hover:bg-surface-primary-selected-hover"
+              : ""
```
edloidas: Can be written simpler: `isWidgetActive && 'bg-surface-primary-selected hover:bg-surface-primary-selected-hover'`.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
+import {Widget} from "@enonic/lib-admin-ui/content/Widget";
+import {Store} from "@enonic/lib-admin-ui/store/Store";
+import {LegacyElement} from "@enonic/lib-admin-ui/ui2/LegacyElement";
+import {CONFIG} from "@enonic/lib-admin-ui/util/Config";
+import {IconButton, Tooltip, Button, cn} from "@enonic/ui";
+import {CircleQuestionMark, Pen} from "lucide-react";
+import {AppContext} from "../../../../app/AppContext";
+import {ProjectIcon} from "../../../../app/ui2/list/ProjectIcon";
+import {ProjectContext} from "../../../../app/project/ProjectContext";
```
edloidas: **TL;DR:** single quotes for code, double quotes for JSX attributes.

Sorry for the inconvenience — we don’t have a proper prettifier or styling set up yet. But we use single quotes for strings, except when they’re part of JSX directly. In JSX, we use double quotes like in HTML. For example, `size="lg"` is fine, but in code parts like `const foo = 'bar'`, `className={cn('w-8')}`, or `type Foo = 'bar' | 'baz';`, we use single quotes.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  widgets = [],
+  activeWidgetId = "",
+  handler = () => {},
+}: Props): React.ReactElement => {
+  const project = ProjectContext.get().getProject();
+  const version = "v." + CONFIG.getString("appVersion");
+  const mainWidgets = widgets.slice(0, -1);
+  const lastWidget = widgets.at(-1);
+
+  const SidebarHeader = () => (
```
edloidas: Don't define new components inside the `SidebarComponent`. It's actualy an antipattern, that will cause performance and visual problems:
- [Do Not Declare Components Within Other Components](https://www.perssondennis.com/articles/react-anti-patterns-and-best-practices-dos-and-donts#do-not-declare-components-within-other-components)
- [ESLint: no-nested-component-definitions](https://eslint-react.xyz/docs/rules/no-nested-component-definitions)
- Many other links you can google

edloidas: Remove that one and the rest too

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+    <>
+      <div class="w-[32px] h-[32px]">
+        <ProjectIcon
+          projectName={project.getName()}
+          language={project.getLanguage()}
+          hasIcon={!!project.getIcon()}
+        />
+      </div>
+      <p
```
edloidas: Just semantically, I think it's better to use header like `<h1>`

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+    <div className="flex flex-col gap-2 items-center">
+      {mainWidgets.map((widget) => (
+        <WidgetButton
+          widget={widget}
+          activeWidgetId={activeWidgetId}
+          handler={handler}
+        />
+      ))}
+    </div>
+  );
```
edloidas: Aslo, I don't see the actual reason to move those to separate components. You can just as well do it inside the render function. If you need to visually distinct different parts, you can just add commet separators, like 
```tsx
<div class="flex flex-col justify-between h-full">
    { /* Widgets */ }
    <div className="flex flex-col gap-2 items-center">
        {mainWidgets.map((widget) => (
```

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+import {ProjectIcon} from "../../../../app/ui2/list/ProjectIcon";
+import {ProjectContext} from "../../../../app/project/ProjectContext";
+
+type WidgetHandler = (widgetId: string) => void;
+
+type Props = {
+  name?: string;
+  widgets?: Widget[];
+  activeWidgetId?: string;
+  handler?: WidgetHandler;
```
edloidas: Too generic. Can't really understand what kind of handler it is? Keeping plain old `onClick` would be better

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  widget,
+  activeWidgetId,
+  handler,
+}: Pick<Props, "activeWidgetId" | "handler"> & { widget: Widget }): React.ReactElement => {
+  const [widgetKey, widgetDisplayName, widgetIconUrl, widgetFullIconUrl] = [
+    widget.getWidgetDescriptorKey().toString(),
+    widget.getDisplayName(),
+    widget.getIconUrl(),
+    widget.getFullIconUrl(),
+  ];
```
edloidas: Why destructing? 
```ts
const widgetKey = widget.getWidgetDescriptorKey().toString();
const widgetDisplayName = widget.getDisplayName();
// ...
```
Works the same if not better.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+              ? "bg-surface-primary-selected hover:bg-surface-primary-selected-hover"
+              : ""
+          )}
+          size="lg"
+          aria-label={widget.getDisplayName()}
+          onClick={() => handler(widgetKey)}
+        >
+          <img
+            class={cn(
+              "w-[20px] invert-100 dark:invert-0",
```
edloidas: Must be larger. 24px (`w-6`) in design.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  return (
+    <Tooltip value={widgetDisplayName} side="right">
+      <IconButton
+        className={
+          isWidgetActive
+            ? "bg-surface-primary-selected hover:bg-surface-primary-selected-hover text-white"
+            : "text-black dark:text-white"
+        }
+        size="lg"
```
edloidas: Icon size must be larger too. Can be set via `iconSize={24}`

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+
+  return (
+    <Tooltip value={widgetDisplayName} side="right">
+      <IconButton
+        className={
+          isWidgetActive
+            ? "bg-surface-primary-selected hover:bg-surface-primary-selected-hover text-white"
+            : "text-black dark:text-white"
```
edloidas: We don’t use `dark:*` or default theme colors at all — unless it’s absolutely necessary.
The dark theme is handled automatically through variables.

I realize you don’t have access to the Enonic UI repo, but you can probably see those colors in the source code or inspect them in DevTools.
Here’s a small cheat sheet for the basic text colors:

```css
/* B&W Colors for Text */

  /* Black on light, White on dark theme */
--color-main: var(--color-main);

  /* Slightly lighter/darker than main */
--color-main-hover: var(--color-main-hover);

  /* Slightly lighter/darker than main, close to hover. Use for small subtext */
--color-subtle: var(--color-subtle);

  /* Always white  */
--color-alt: var(--color-alt);

  /* Reversed: White on light, Black on dark */
--color-rev: var(--color-rev);

  /* Always black */
--color-alt-rev: var(--color-alt-rev);
```

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+        }
+        size="lg"
+        icon={icon}
+        aria-label={widget.getDisplayName()}
```
edloidas: `widgetDisplayName`? Since you defined it already

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
edloidas: Also, I think it can be improved. 

You see, when you pass a widget there, even though it may have the same props, it could still be a different `object` — which will cause `WidgetButton` to re-render. 

On the other hand, if I mutate the object somewhere (though that’s not our case right now, since it would also affect the parent component), it won’t trigger a re-render of `WidgetButton`, because the reference and property values remain the same.

edloidas: `widgetIconUrl` is only used inside `if`, while `widgetFullIconUrl` is passed to props. It can be combined into single property.

edloidas: I suggest you to rework the
```tsx
type WidgetButtonProps = {
  // Instead of Display name
  label: string;

  // `widgetIconUrl && widgetFullIconUrl` ?
  iconUrl?: string;

  // Widget doesn't need to check if it;s active itself. Parent should do the check 
  active?: boolean;

  // We pass the right function with the correct ID already, e.g. `() => onClick(widgetId)` from parrent.
  onClick?: () => void;

  // We can also do `...props` and later pass default props in JSX as `{...props} to Button and IconButton for better customization possiblities
} & ComponentPropsWithoutRef<'button'>; 
```

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+}: Pick<Props, "activeWidgetId" | "handler"> & { widget: Widget }): React.ReactElement => {
+  const [widgetKey, widgetDisplayName, widgetIconUrl, widgetFullIconUrl] = [
+    widget.getWidgetDescriptorKey().toString(),
+    widget.getDisplayName(),
+    widget.getIconUrl(),
+    widget.getFullIconUrl(),
+  ];
+  const baseIcon = CircleQuestionMark;
+  const isWidgetActive = widgetKey === activeWidgetId;
+  const icon = SidebarElement.isMainWidget(widget) ? Pen : baseIcon;
```
edloidas:
```diff
- const icon = SidebarElement.isMainWidget(widget) ? Pen : baseIcon;
+ const icon = SidebarElement.isMainWidget(widget) ? Pen : CircleQuestionMark;
```

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+            isWidgetActive
+              ? "bg-surface-primary-selected hover:bg-surface-primary-selected-hover"
+              : ""
+          )}
+          size="lg"
+          aria-label={widget.getDisplayName()}
+          onClick={() => handler(widgetKey)}
+        >
+          <img
+            class={cn(
```
edloidas: Using `class=` allowed, but generally incorrect in React. Use `className`.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  );
+
+  const SidebarWidgets = () => (
+    <div className="flex flex-col gap-2 items-center">
+      {mainWidgets.map((widget) => (
+        <WidgetButton
+          widget={widget}
+          activeWidgetId={activeWidgetId}
+          handler={handler}
+        />
```
edloidas: Must pass `key` to `<WidgetButton>` to avoid re-rendering. Can probably use `String(widget.getWidgetDescriptorKey())` for that.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+      widgetKey.endsWith(SidebarElement.MAIN_APP_ENDING),
+      widgetKey.endsWith(SidebarElement.ARCHIVE_APP_ENDING),
+      widgetKey.endsWith(SidebarElement.SETTINGS_APP_ENDING),
+    ].every((widgetStatus) => widgetStatus === false);
+  }
+
+  onItemSelected(handler: WidgetHandler): void {
+    this.props.setKey("handler", handler);
+
+    return;
```
edloidas: Meaningless returns. Please, remove.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const SidebarFooter = () => (
+    <div>
+      {lastWidget && (
+        <WidgetButton
+          widget={lastWidget}
+          activeWidgetId={activeWidgetId}
+          handler={handler}
+        />
+      )}
+      <p class="text-cs text-surface-primary-selected text-center overflow-hidden no-wrap text-nowrap max-w-[40px]">
```
edloidas: `no-wrap` does not exist in Tailwind. Use `whitespace-nowrap`.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
edloidas: I also see that `handler?: WidgetHandler` is optional. Don't forget to call it like `handler?(widgetId)` for type safety.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+        {version}
```
edloidas: Can also wrap <p> in Tooltip with short timeout or at least add `title`. Sometimes, versions can be much longer, e.g. `v.0.0.0.enonic-ui-SNAPSHOT`.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -0,0 +1,242 @@
...
+type WidgetHandler = (widgetId: string) => void;
+
+type Props = {
+  name?: string;
+  widgets?: Widget[];
+  activeWidgetId?: string;
+  handler?: WidgetHandler;
+};
+
+const WidgetButton = ({
```
edloidas: I'd move it to a separate `WidgetButton` file, as it's not actually a sidebar part.

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
edloidas: Hm, I see you call a static method from a `SidebarElement` class, which might be removed in a few months. Just move that to separate function instead. But probably in Second task. We don't want any widget manipulation logic left in `SdebarElement`.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
reisfmb: 1. 👍 
2. 👍 
3. 👍 
4. There was a reason to wrap ProjectIcon in a div and style the div itself. Without it, screen vertical resizing was affecting the ProjectIcon height. 
5. Couldn't locate that on the figma design... What I adjusted was the padding of the sidebar itself from `p-2` to `px-2.5 py-1.75`

edloidas: 4. I think the reason for this is flex layout. Icon is missing `flex-shrink-0` which won't allow it to shrink below 32px automatically.
image

5. Sorry, not 14 for each side, but in total. You see, your icon is 32px, and is placed right after Sidebar padding, while in design, icon is in a container 46px height, so there is not 10px, but 17px to between the screen top border and the icon top border. So, if we remove the drapper div, you'll have to add `my-2.25`
image

image

image

edloidas (approved):
Nice work 👍

PR #9274, opened 2025-10-25
Enonic UI: Externalize Widgets Data Management #9268

### LINKED ISSUE #9268: Enonic UI: Externalize Widgets Data Management
Move widgets data and management logic from local component state into a global store for better separation of concerns and centralized data management.

---

## Rationale

- **Separation of concerns:** Widgets are serialized data loaded from the server, not UI implementation details. Data management should be decoupled from presentation components.
- **Single responsibility:** Sidebar component should focus on rendering and user interactions, not on data fetching, sorting, or state management logic.
- **Centralized data source:** Multiple components need access to widgets data (`GetWidgetsByInterfaceRequest` is used in several places). A global store eliminates duplicate requests and provides a single source of truth.
- **Simplified component API:** Components consume data reactively from the store without needing imperative methods for data manipulation.

---

## References

- [`WidgetsToolbar.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/widget/WidgetsToolbar.ts) — current implementation managing widget ordering and state
- [`nanostores` documentation](https://github.com/nanostores/nanostores) — store library we're using
- [`@nanostores/react` documentation](https://github.com/nanostores/react) — React bindings for nanostores
- [`GetWidgetsByInterfaceRequest`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/resource/GetWidgetsByInterfaceRequest.ts) — server API for fetching widgets
---

## Things to Consider

- **Store architecture:** Evaluate whether to use a single `$widgets` store or split by widget type (e.g., `$sidebarWidgets`, `$contextWidgets`) based on usage patterns.
- **Data consolidation:** `GetWidgetsByInterfaceRequest` is currently invoked from multiple locations. Consider centralizing all widget fetching logic in the store to avoid redundant requests.
- **Naming convention:** Follow project convention of prefixing store names with `$` (e.g., `$widgets`).
- **Default selection:** First widget should be automatically selected on initial load.
- **Store exports:** Avoid default exports — export named stores and action functions explicitly.
- **Migration path:** `SidebarElement` wrapper methods for widget management will be removed once store is integrated.

---

## Implementation

> [!IMPORTANT]  

1. **Create the store file:**
   - Create `modules/lib/src/main/resources/assets/js/v6/features/store/widgets.store.ts`
   - Follow existing project patterns for store organization

2. **Design store structure:**
   - Use `map` from nanostores for complex state
   - Consider splitting widgets by type if they have different lifecycles/purposes
   - Name main store with `$` prefix (e.g., `$widgets` or `$sidebarWidgets`)
   - Define TypeScript types for widget data structure

3. **Implement store actions:**
   - Create public setter functions for store mutations (following pattern from example)
   - `loadWidgets()` — fetch and populate widgets from server
   - `setActiveWidget(widgetId)` — manage active widget selection
   - Export actions as named functions

4. **Integrate data fetching:**
   - Consolidate `GetWidgetsByInterfaceRequest` calls into store logic
   - Ensure widgets are loaded once and cached
   - Handle loading states and errors appropriately

5. **Connect Sidebar to store:**
   - Use `useStore` hook from `@nanostores/react` in Sidebar component
   - Replace local state and props with store consumption
   - Remove widget management methods from `SidebarElement` wrapper

6. **Implement default selection:**
   - Auto-select first widget on initial load
   - Ensure active state persists across re-renders

7. **Cleanup:**
   - Remove existing related `Widgets*` files.

**UI Behavior:**

- First widget is automatically selected when page loads
- Clicking a widget button updates active state in store and triggers view change
- Active widget button displays visual indication
- Widget selection state persists until user selects another widget

---

## Acceptance Criteria

- [x] Widget data is stored in global nanostores store
- [x] Sidebar consumes widgets from store using `useStore` hook
- [x] First widget is automatically selected on page load
- [x] Clicking widget button updates active state and changes view
- [x] Active widget button shows correct visual indication
- [x] Same widgets are loaded as in old implementation
- [x] No duplicate `GetWidgetsByInterfaceRequest` calls across application
- [x] Widget sorting logic moved from WidgetsToolbar to store
- [x] `SidebarElement` no longer contains widget management methods

---

## Testing Steps

1. Open Content Studio in browse view and locate Sidebar
2. Verify first widget button is automatically selected (active state visible)
3. Verify corresponding widget view is displayed for the selected widget
4. Click different widget buttons and confirm:
   - Active button state updates correctly
   - Widget view changes to match selected button
   - Previous button loses active state
5. Check browser DevTools Network tab:
   - Verify widgets are only loaded once per session
   - Confirm no duplicate requests to `GetWidgetsByInterfaceRequest`
6. Compare widget list and order with old implementation to ensure parity
7. Test with browser refresh — first widget should still be selected

---

## Examples

**Store structure example:**

```ts
import {map} from 'nanostores';
import {GetWidgetsByInterfaceRequest} from '../../app/resources';

export type Widget = {
    id: string;
    name: string;
    iconUrl: string;
    displayName: string;
    order?: number;
};

export type WidgetsStore = {
    items: Widget[];
    activeWidgetId: string | null;
    isLoading: boolean;
    error: string | null;
};

export const $widgets = map<WidgetsStore>({
    items: [],
    activeWidgetId: null,
    isLoading: false,
    error: null,
});

export const loadWidgets = async (): Promise<void> => {
    $widgets.setKey('isLoading', true);
    
    try {
        const request = new GetWidgetsByInterfaceRequest('contentstudio.menuitem');
        const widgets = await request.sendAndParse();
        const sortedWidgets = sortWidgetsByOrder(widgets);
        
        $widgets.setKey('items', sortedWidgets);
        
        if (sortedWidgets.length > 0 && !$widgets.get().activeWidgetId) {
            setActiveWidget(sortedWidgets[0].id);
        }
    } catch (error) {
        $widgets.setKey('error', error.message);
    } finally {
        $widgets.setKey('isLoading', false);
    }
};

export const setActiveWidget = (widgetId: string): void => {
    $widgets.setKey('activeWidgetId', widgetId);
};

const sortWidgetsByOrder = (widgets: Widget[]): Widget[] => {
    return [...widgets].sort((a, b) => (a.order ?? 0) - (b.order ?? 0));
};
```

**React component usage:**

```tsx
import {useStore} from '@nanostores/react';
import {$widgets, setActiveWidget} from './store/widgets.store';

export const Sidebar = () => {
    const {items, activeWidgetId} = useStore($widgets);
    
    return (
        <div className="sidebar">
            {items.map(widget => (
                <IconButton
                    key={widget.id}
                    icon={widget.iconUrl}
                    active={widget.id === activeWidgetId}
                    onClick={() => setActiveWidget(widget.id)}
                />
            ))}
        </div>
    );
};
```

---

## Out of Scope

- Passing widget-specific data/props to other page elements
- Error handling UI for failed widget loads (basic error state in store is sufficient)
- Persisting active widget selection to localStorage/backend

---

## Technical Constraints

- Must use nanostores (not zustand or other state libraries)
- Store must integrate with existing `GetWidgetsByInterfaceRequest` API
- Active widget state should only persist in memory (no persistence to localStorage in this phase)

reisfmb:
@edloidas

1. Opted to manage each type of widget (sidebar / context / liveview) on their on store. The reason is because, joining the necessary logic of all of them on a single store would make the store big and complex. For example, the sort logic of widgets in the sidebar is different than the sort logic of widgets in liveview. Also, some of them combine hardcoded widgets with the one coming from the request.

2. The current draft PR only implements the sidebarWidgets store. I decided to open the PR like so for two reasons: 
- First, for you to check if the approach I took is the one you had in mind;
- Second, because in my head it'd make more sense to implement the context and liveview stores when we are also migrating their components.

3. I believe I can improve the code quality a bit, that's another reason I opened the PR in draft. Feel free to comment on it and I'll adjust as needed.


modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/Sidebar.tsx
```diff
@@ -3,34 +3,24 @@ import {Store} from '@enonic/lib-admin-ui/store/Store';
...
     return widget.getWidgetDescriptorKey().toString().endsWith('studio:main');
 }
 
-const SidebarComponent = ({
-    name = Store.instance().get('application').getName(),
-    widgets = [],
-    activeWidgetId = '',
-    onClick = () => {},
-}: Props): React.ReactElement => {
+const SidebarComponent = (): React.ReactElement => {
```
edloidas: Please rename it to just `Sidebar`. Components are our default items. When we wrap them in `LegacyElement`, we’ll add `*Element` to the end to avoid name collisions and to make it clear that they’re wrappers. The majority of these elements will be removed in the future.

edloidas: And let's just use `ReactElement` importing it from `react` package, since that's how we usually do.

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
+import {computed, map} from 'nanostores';
+import {GetWidgetsByInterfaceRequest} from '../../../app/resource/GetWidgetsByInterfaceRequest';
+import {Widget, WidgetConfig} from '@enonic/lib-admin-ui/content/Widget';
+import {CONFIG} from '@enonic/lib-admin-ui/util/Config';
+import {i18n} from '@enonic/lib-admin-ui/util/Messages';
+import {UrlAction} from '../../../app/UrlAction';
+
+interface WidgetsStore {
+    items: Widget[];
```
edloidas: Maybe rename it to `widgets` since you use it like `{items: widgets, ... }` in code anyway? Or below in code:
```diff
- const widgets = $sidebarWidgets.get().items;
+ const {widgets} = $sidebarWidgets.get();
```

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
...
+    const widgets = $sidebarWidgets.get().items;
+
+    const widgetMatchingUrl = widgets.find((w) => url.endsWith(`/${w.getWidgetDescriptorKey().getName()}`));
+
+    if (widgetMatchingUrl) {
+        setActiveWidget(widgetMatchingUrl);
+        return;
+    }
+
+    setActiveWidget(widgets?.[0]);
```
edloidas: Wouldn't `widgets` be defined already?

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export const $sidebarWidgets = map<WidgetsStore>({
+    items: [],
+    activeWidgetId: null,
+    isLoading: false,
+    error: null,
+});
+
+export async function loadWidgets(): Promise<void> {
```
edloidas: I think that we don't need to call it explicitly. I suggest we don't export it at all and wrap it into IIFE, e.g.:
```ts
async function loadWidgets(): Promise<void> {}

(function () {
  try {
  // ...
})(); 
```

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
...
+import {Widget, WidgetConfig} from '@enonic/lib-admin-ui/content/Widget';
+import {CONFIG} from '@enonic/lib-admin-ui/util/Config';
+import {i18n} from '@enonic/lib-admin-ui/util/Messages';
+import {UrlAction} from '../../../app/UrlAction';
+
+interface WidgetsStore {
+    items: Widget[];
+    activeWidgetId: string | null;
+    isLoading: boolean;
+    error: string | null;
```
edloidas: I’m not sure we need those variables right now in THIS store. We don’t need to track whether widgets are being loaded and don’t show any loading progress in the toolbar. Let’s keep it simple and drop them both.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -144,85 +107,58 @@ export class AppWrapper
...
-            (this.isDefaultWidget(widget) || !widget.getDisplayName()) ? i18n('app.admin.tool.title') : widget.getDisplayName();
+            this.isDefaultWidget(widget) || !widget.getDisplayName()
+                ? i18n('app.admin.tool.title')
+                : widget.getDisplayName();
         document.title = `${prefix} - ${postfix}`;
     }
 
     private isDefaultWidget(widget: Widget): boolean {
-        return widget === this.widgets[0];
+        return widget === Store.$sidebarWidgets.get().items?.[0];
```
edloidas: 1. Move the check into the store itself.
2. The comparison is (was) unsafe - you’re comparing object references, which might change. Compare by IDs instead.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -144,85 +107,58 @@ export class AppWrapper
...
-                        this.activeWidgets.push(widgetKey);
-                    });
+                WidgetHelper.createFromHtmlAndAppend(html, this.widgetsBlock).then((widgetEl: WidgetElement) => {
+                    const widgetKey = widget.getWidgetDescriptorKey().toString();
+                    this.widgetElements.set(widgetKey, widgetEl);
+                    this.activeWidgets.push(widgetKey);
+                });
             })
-            .catch(err => {
+            .catch((err) => {
```
edloidas, line 129: Nah, it’s totally fine to keep a single parameter without `()` in such cases — same as in `.map`, `.filter`, and other utility functions. Unless it’s just an automatic fix, of course.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -144,85 +107,58 @@ export class AppWrapper
...
-    }
-
-    private addMissingWidgets(widgets: Widget[]): void {
-        widgets.filter((widget: Widget) => !this.hasWidget(widget)).forEach((widget: Widget) => {
-            this.widgets.push(widget);
-            this.sidebar.addWidget(widget);
-            this.notifyItemAdded(widget);
-        });
+    private loadWidgets() {
+        Store.loadWidgets().catch(DefaultErrorHandler.handle);
```
edloidas: Can be removed entirely (see the comments in the store for details).
The `catch` block won’t catch anything, since you’re not throwing a new error and already have `try–catch–finally` blocks inside.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -144,85 +107,58 @@ export class AppWrapper
...
 
     private isAppStopStartEvent(event: ApplicationEvent): boolean {
-        return ApplicationEventType.STOPPED === event.getEventType() || ApplicationEventType.UNINSTALLED === event.getEventType()
-               || ApplicationEventType.STARTED === event.getEventType() || ApplicationEventType.INSTALLED === event.getEventType();
+        return (
+            ApplicationEventType.STOPPED === event.getEventType() ||
+            ApplicationEventType.UNINSTALLED === event.getEventType() ||
+            ApplicationEventType.STARTED === event.getEventType() ||
+            ApplicationEventType.INSTALLED === event.getEventType()
+        );
```
edloidas: Thinking about moving this to the store.
We could:
1. Remove widgets using a loose ID search based on the App ID we get from the event.
2. Reload when the apps are loaded.

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
...
+    try {
+        const request = new GetWidgetsByInterfaceRequest('contentstudio.menuitem');
+        const response = await request.sendAndParse();
+        const widgets = [createStudioWidget(), ...response];
+        const sortedWidgets = sortWidgets(widgets);
+
+        $sidebarWidgets.setKey('items', sortedWidgets);
+        setInitialActiveWidget();
+    } catch (error) {
+        $sidebarWidgets.setKey('error', error.message);
```
edloidas: Would be nice to log the error message with `console.error()`.
`DefaultErrorHandler` can also be used, but it should log a human-readable error — that part can be skipped for now.

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -0,0 +1,105 @@
...
+        return [
+            widgetKey.endsWith(MAIN_APP_ENDING),
+            widgetKey.endsWith(ARCHIVE_APP_ENDING),
+            widgetKey.endsWith(SETTINGS_APP_ENDING),
+        ].every((widgetStatus) => widgetStatus === false);
+    };
+    const mainWidget = widgets.find(isMainWidget);
+    const archiveWidget = widgets.find(isArchiveWidget);
+    const settingsWidget = widgets.find(isSettingsWidget);
+    const defaultWidgets = widgets.filter(isDefaultWidget);
```
edloidas: Why not just pass an arrow function to `find()` directly? You’re not reusing or composing it, and the variable names are already descriptive enough.

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
edloidas: The `loadWidgets` function can also be improved if we move the `ApplicationEvent` events here. Since we don’t expect apps to be updated frequently, we can load as soon as we get an update for better UX, and then reload if any new updates arrive while the loading is in progress.

I’d start with this (the parts after `<- ...` are just my comments):
```ts
//
// * Fetching <- Separate the parts of the file
//

let isLoading = false; // <- keeping it as a variable outside the store is good enough
let needsReload = false;

async function loadWidgets() {
    // <- Mark that we need to update the data again even after the current fetching
    if (isLoading) {
        needsReload = true;
        return;
    }

    isLoading = true;

    try {
        // <- current code 
    }

    // <- instead of debounce or do-debounce
    if (needsReload) {
        needsReload = false;
        return loadWidgets(); // <- recursion
    }
}
```

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
```diff
@@ -17,13 +15,10 @@ import {ContentAppBar} from './bar/ContentAppBar';
 import {ResponsiveManager} from '@enonic/lib-admin-ui/ui/responsive/ResponsiveManager';
 import {cn} from '@enonic/ui';
 import {SidebarElement} from '../v6/features/layout/AppShell/Sidebar';
+import * as Store from '../v6/features/store/sidebarWidgets.store';
```
edloidas: Avoid doing batch imports - you only need one or two exports from that module.
Also, Store is our global name for the synced storage shared between app parts.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
reisfmb: Yep, but this was an automatic fix.

modules/lib/src/main/resources/assets/js/app/AppWrapper.ts
reisfmb: 2. When I was implementing I just replaced it with the store value. Good catch!

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
reisfmb: Great way to avoid debouncing 👏

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
reisfmb: At this points the widgets on the store would be defined, but not the activeWidgetId.

edloidas (approved):

edloidas:
Impressive 👍

Issue #9243, opened 2025-10-15
Enonic UI: ConfirmationDialog.Gate


edloidas:
Gate is a validation component that controls the rendering of certain parts of the `ConfirmationDialog` by toggling the `unlocked` state.

### Rationale
- While Gate could be implemented as a standalone component, doing so would require substantial effort to make it robust. In Content Studio, there are only a few places where it’s actually needed. Hence, KISS.

### References
- Refer to the design in the attached screenshots and the Figma project.  
- It’s part of the *Archive Dialog* section in the design documentation.

### Implementation

> [!IMPORTANT]  
> This is not a detailed step-by-step guide. It’s a logically ordered list of required functionality. Ensure all necessary parts are implemented.

1. In the `ConfirmationDialog.tsx` file, create a new component part named `ConfirmationDialogGate` and export it from `ConfirmationDialog` as `Gate`.
2. We won’t use a composite approach, but we’ll still define potential parts within a single component:
   - `Gate.Root` — a basic `<div>` wrapper.  
   - `Gate.Hint` — a text element `<p>` with a predefined format, e.g., `Enter <b>{value}</b> in the field and click Confirm`.  
   - `Gate.Input` — uses the Enonic UI `<Input>` component with specific properties.
3. Default validation should normalize and check the input value.  
4. The input should call `setUnlocked(true)` once validation passes.  
5. Use `value.trim()` during validation to ignore trailing whitespace.  
6. Gate must remain configurable. It should, at a minimum, support the following properties:

```ts
type Props = {
  // Whole validation
  validate: (value: string) => boolean;
  className?: string;
} & Pick<ComponentPropsWithoutRef<'input'>, 'inputMode' | 'autoFocus'>;
```

**UI Logic:**
- Set the input to `readOnly` once it becomes valid.
- Mark the input with an error (both border and message) after approximately 1 second.
- Remove the error immediately when the value changes.
- The error message should be descriptive, explaining the mismatch between the required and entered number.

### Examples

<img width="1480" height="286" alt="Image" src="https://github.com/user-attachments/assets/d8c209c5-003e-4f97-aa17-9e363ad89582" />

PR #9280, opened 2025-10-28
Enonic UI: ConfirmationDialog.Gate #9243

### LINKED ISSUE #9250: Enonic UI: DialogPreset.Confirm
`DialogPreset.Confirm` is a component with a predefined structure and limited customization options. It’s designed to simplify the creation of confirmation dialogs throughout the application.

### Rationale
- We need a preset to easily recreate similar components with a static structure.

### Implementation

> [!IMPORTANT]  
> This is not a detailed step-by-step guide. It’s a list of required functionality, ordered logically. Make sure all necessary parts are implemented.

1. In the `DialogPreset.tsx` file, create a new component called `DialogPresetConfirm` and export it from `DialogPreset` as `Confirm`.
2. Research which properties can and should be shared across different confirmation dialogs, and move them into the preset’s props.
3. Optionally render `<ConfirmationDialog.Gate>` within the preset and allow changing the `intent`.

The preset itself may look like this:

```ts
type DialogPResetConfirmProps = {
  open: boolean;
  onOpenChange: (open: boolean) => void;
  title: string;
  description?: ReactNode;
  confirmLabel?: string;
  onConfirm: () => void | Promise<void>;
};

const DialogPResetConfirm = ({
  open,
  title, 
  description,
  confirmLabel,
  onConfirm,
}: DialogPResetConfirmProps): ReactElement => {
  return (
    <ConfirmationDialog.Root open={open}>
      <ConfirmationDialog.Portal>
        <ConfirmationDialog.Overlay />
        <ConfirmationDialog.Content defaultUnlocked={false}>
          <ConfirmationDialog.DefaultHeader title={title} withClose />
          <ConfirmationDialog.Body>
            {description}
          </ConfirmationDialog.Body>
          <ConfirmationDialog.Footer
            onConfirm={onConfirm}
            confirmLabel={confirmLabel}
          />
        </ConfirmationDialog.Content>
      </ConfirmationDialog.Portal>
    </ConfirmationDialog.Root>
  );
};
```

skjulestad:
Added Dialog Gate and preset for delete confirmation.


modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -11,14 +10,16 @@ import {useI18n} from '../hooks/useI18n';
 type ConfirmationDialogContextValue = {
     submitEnabled: boolean;
     setSubmitEnabled: (submitEnabled: boolean) => void;
+    unlocked: boolean;
+    setUnlocked: (unlocked: boolean) => void;
```
edloidas: We renamed `unlocked` to `submitEnabled`. So we already have it in context, no need to add a duplicate.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -11,14 +10,16 @@ import {useI18n} from '../hooks/useI18n';
...
 };
 
 const ConfirmationDialogContext = createContext<ConfirmationDialogContextValue | null>(null);
 
 export const useConfirmationDialog = (): ConfirmationDialogContextValue => {
     const ctx = useContext(ConfirmationDialogContext);
     if (!ctx) {
-        return {submitEnabled: true, setSubmitEnabled: () => void 0};
+        return {submitEnabled: true, setSubmitEnabled: () => void 0, unlocked: true, setUnlocked: () => void 0};
```
edloidas: Same duplicate

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -30,8 +31,9 @@ type ConfirmationDialogProviderProps = {
 
 const ConfirmationDialogProvider = ({defaultSubmitEnabled = true, children}: ConfirmationDialogProviderProps): ReactElement => {
     const [submitEnabled, setSubmitEnabled] = useState<boolean>(defaultSubmitEnabled);
+    const [unlocked, setUnlocked] = useState<boolean>(true);
```
edloidas: Same duplicate

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -30,8 +31,9 @@ type ConfirmationDialogProviderProps = {
...
 
-    const value = useMemo<ConfirmationDialogContextValue>(() => ({submitEnabled, setSubmitEnabled}), [submitEnabled]);
+    const value = useMemo<ConfirmationDialogContextValue>(() => ({submitEnabled, setSubmitEnabled, unlocked, setUnlocked}), [submitEnabled, unlocked]);
```
edloidas: Same duplicate

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -81,7 +179,7 @@ const ConfirmationDialogFooter = ({
     const {submitEnabled} = useConfirmationDialog();
 
     const cancel = useI18n('action.cancel');
-    const submit = useI18n('action.submit');
+    const submit = useI18n('action.confirm');
```
edloidas: Yeah, we have submit. Rename all related variables from `submit` to `confirm` too, e.g. `submit`, `onSubmit`, etc.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/DialogPreset.tsx
```diff
@@ -53,3 +53,60 @@ export class DialogPresetConfirmElement
         this.setProps({open: false});
     }
 }
+
+type DialogPresetConfirmDeleteProps = {
```
edloidas: It's not Delete-specific. Better name it `DialogPresetGatedConfirm` as we added Gate validation there.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/DialogPreset.tsx
```diff
@@ -53,3 +53,60 @@ export class DialogPresetConfirmElement
...
+  open?: boolean;
+  title: string;
+  description?: string;
+  expected: string | number;
+  validate?: (value: string) => boolean;
+  onSubmit?: () => void;
+  onCancel?: () => void;
+  intent?: 'default' | 'danger';
+};
```
edloidas: Also, I think we can make it shorter by reusing existing props:
```ts
type DialogPresetConfirmDeleteProps = {
  expected: string | number;
  validate?: (value: string) => boolean;
  intent?: 'default' | 'danger';
} & Omit<DialogPresetConfirmProps, 'defaultSubmitEnabled'>;
```

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
                 <ConfirmationDialogProvider defaultSubmitEnabled={defaultSubmitEnabled}>{children}</ConfirmationDialogProvider>
             </Dialog.Content>
         );
     }
 );
 ConfirmationDialogContent.displayName = 'ConfirmationDialog.Content';
 
+//
+// * ConfirmationDialogGate
+//
```
edloidas: Move **Gate** into a separate component. We either don’t need to make it composite, or we don’t need a second level of nesting.
`ConfirmationDialog.Gate.Input` feels not right, so it should be either `ConfirmationDialog.Gate` or `Gate.*`.
I suggest going with the second option.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
+
+const ERROR_DELAY_MS = 500;
+const digitsOnly = (s: string): string => s.replace(/\D+/g, '');
+const normalize = (s: string): string => digitsOnly(s.trim());
+const buildError = (entered: string, expected?: string): string => (
+    expected
+        ? useI18n('dialog.confirmValue.error.mismatch', expected, entered || useI18n('dialog.confirmValue.error.nothing'))
+        : useI18n('dialog.confirmValue.error.invalid')
+);
```
edloidas: Please, rewrite this:
1. Hooks (even though it's not a hook yet) must not be called or defined outside components.
2. Don't put one hook into another.
Start with standalone hooks and more flat structure, something like this:
```ts
const errorInvalid = useI18n('dialog.confirmValue.error.invalid');
const errorMismatch = entered ? useI18n('dialog.confirmValue.error.mismatch', expected, entered) : useI18n('dialog.confirmValue.error.nothing', expected);
const errorMessage = expected ? errorMismatch : errorInvalid;
const error = showError ? errorMessage : undefined;
```

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
     }
 );
 ConfirmationDialogContent.displayName = 'ConfirmationDialog.Content';
 
+//
+// * ConfirmationDialogGate
+//
+
+const ERROR_DELAY_MS = 500;
+const digitsOnly = (s: string): string => s.replace(/\D+/g, '');
```
edloidas: I don't think this works as intended.
If we expect digits only, that must be enforced at the input level.
Here you're just extracting digits from whatever was entered. But `1_d3` is not `13`, and with this normalization it becomes `13`.
I suggest removing this function entirely.

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
+    expected?: string | number;
+    className?: string;
+} & Pick<ComponentPropsWithoutRef<'input'>, 'inputMode' | 'autoFocus'>;
+
+const GateRoot = ({className, children, ...props}: ComponentPropsWithoutRef<'div'>): ReactElement => (
+    <div className={cn('flex flex-col gap-2.5 p-7.5 bg-surface-primary rounded-lg', className)} {...props}>{children}</div>
+);
+GateRoot.displayName = 'ConfirmationDialog.Gate.Root';
+
+const GateHint = ({value, className, ...props}: ConfirmationDialogGateHintProps & ComponentPropsWithoutRef<'p'>): ReactElement => (
```
edloidas: Why not moving `ComponentPropsWithoutRef<'p'>` to `ConfirmationDialogGateHintProps`?

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
+        : useI18n('dialog.confirmValue.error.invalid')
+);
+
+type ConfirmationDialogGateHintProps = {
+    value: string | number;
+    className?: string;
+};
+
+type ConfirmationDialogGateInputProps = {
+    validate?: (value: string) => boolean;
```
edloidas: Would be nice to have custom ` normalize?: (value: string) => string;` function with default to `const defaultNormalize = (value: string) => value.trim();`

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
+        }
+        timerRef.current = window.setTimeout(() => setShowError(true), ERROR_DELAY_MS);
+        return () => {
+            if (timerRef.current) {
+                window.clearTimeout(timerRef.current);
+                timerRef.current = undefined;
+            }
+        };
+    }, [hasValue, valid, value]);
+
```
edloidas: While there's cleanup in the effect, there's no guarantee it runs on unmount if the component is removed while the timer is active. Could cause memory leaks or errors if dialog closes before timer completes. Add top-level unmout for that case:
```ts
useEffect(() => {
      return () => {
          if (timerRef.current) {
              window.clearTimeout(timerRef.current);
          }
      };
  }, []);
```

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
```diff
@@ -50,14 +52,110 @@ const ConfirmationDialogContent = forwardRef<HTMLDivElement, ConfirmationDialogC
...
+} & Pick<ComponentPropsWithoutRef<'input'>, 'inputMode' | 'autoFocus'>;
+
+const GateRoot = ({className, children, ...props}: ComponentPropsWithoutRef<'div'>): ReactElement => (
+    <div className={cn('flex flex-col gap-2.5 p-7.5 bg-surface-primary rounded-lg', className)} {...props}>{children}</div>
+);
+GateRoot.displayName = 'ConfirmationDialog.Gate.Root';
+
+const GateHint = ({value, className, ...props}: ConfirmationDialogGateHintProps & ComponentPropsWithoutRef<'p'>): ReactElement => (
+    <p className={cn('text-xl', className)} {...props}>
+        Enter <b>{value}</b> in the field and click Confirm
```
edloidas: Hardcoded. Must use `useI18n()` hook.

edloidas: Might be a bit tricky, since you need to make text bold. Add a `// TODO: Replace ...` comment, we will fix it later.
But replace `<b>` with `<strong>` as the first is not semantic and won't work with screenreaders.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
skjulestad: My bad, went after the task and forgot we did the renaming. Will fix

modules/lib/src/main/resources/assets/js/app/ui2/dialog/ConfirmationDialog.tsx
skjulestad: Found existing translation for this, no need for todo then!

edloidas (approved):
There are two remaining issues, to be fixed in separate tasks and repositories:
1. We no longer focus the Confirm action after it becomes valid (this must be handled at the Preset level).
2. An empty `<div>` wrapper with the `'mb-2'` class appears when neither a label nor a description is provided. This will be fixed in Enonic UI.

Issue #9279, opened 2025-10-27
Enonic UI: Create AppBar component


reisfmb:
A horizontal bar that appears at the top of the screen in Content Studio, providing quick access to project selection, issues overview, and home screen.

---

## Rationale

- **Fresh implementation:** This is a foundational UI component built from scratch to align with the new Content Studio 6 architecture.
- **Bridge to legacy code:** Using `LegacyElement` wrapper to ensure smooth integration with existing code while establishing modern React patterns.
- **New code structure:** Continues the v6 folder structure (`modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell`), following the patterns established by Sidebar component.

---

## References

- [Existing header implementation] — current implementation to be replaced
- Figma design

---

## Things to Consider

- **Static state for now:** Button labels are static in this iteration — dynamic updates based on external state will be added in follow-up tasks.
- **Button organization:** Three distinct sections: left (project), right group (issues + home). Can be done with flex and `mr-auto` for the project button.
- **Accessibility:** All buttons must be keyboard-accessible and have proper ARIA labels.

---

## Implementation

> [!IMPORTANT]  
> This is not a step-by-step guide — it's a functional checklist ordered logically. Make sure all essential parts are implemented.

1. **Create the component structure:**
   - Create `modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx`
   - Component sits in the same `AppShell` folder as the Sidebar

2. **Build the AppBar component:**
   - Use `<Button>` components from Enonic UI for all interactive elements
   - Left section: Project selector button with static placeholder text (e.g., "Default Project")
   - Right section: Two buttons aligned to the right
     - Issues button with static text (e.g., "Issues")
     - Home/Launcher button with `grid-2x2` icon (will be replaced later)

3. **Implement the legacy wrapper:**
   - Create `AppBarElement` class extending `LegacyElement`
   - Remove tabs functionality

4. **Apply styling:**
   - Match design specifications for height, padding, and background
   - Ensure proper horizontal layout with left and right sections
   - For mobile (small screens `sm:*`), hide the issues button

5. **Localization setup:**
   - Use `useI18n` hook for all button text
   - Add new phrases to `phrases.properties` file if needed

6. **Buttons:**
   - Project button: Opens project selection dialog
     - Currently uses `ProjectViewer` with `onClick` handler that is set in `ContentAppBar`. Drop it.
     - Use `arrow-left-right` icon on the right.
[...]

PR #9302, opened 2025-10-30
Enonic UI: Externalize Projects Data Management #9285

### LINKED ISSUE #9285: Enonic UI: Externalize Projects Data Management
Move projects data and management logic from local component state into a global store for better separation of concerns and centralized data management.

---

## Rationale

- **Separation of concerns:** Projects are serialized data loaded from the server, not UI implementation details. Data management should be decoupled from presentation components.
- **Single responsibility:** AppBar component should focus on rendering and user interactions, not on data fetching, event handling, or state management logic.
- **Centralized data source:** Multiple components need access to projects data and active project state. A global store eliminates duplicate requests and provides a single source of truth.
- **Event-driven updates:** Projects can be created, updated, or deleted during the session. Centralizing event listeners in the store ensures consistent state across the application.
- **Simplified component API:** Components consume data reactively from the store without needing imperative methods for data manipulation.

---

## References

- [`ContentAppBar.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/bar/ContentAppBar.ts#L52) — current implementation managing projects state
- [`ProjectUpdatedEvent.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/settings/event/ProjectUpdatedEvent.ts) — event fired when project is updated
- Events folder — contains `ProjectCreatedEvent` and `ProjectDeletedEvent`
- [`sidebarWidgets.store.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts) — reference implementation for similar store pattern
- [`ProjectContext.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/project/ProjectContext.ts) — singleton class currently used to manage the current project and store the active ID in LocalStorage. Mostly outdated and uses OOP concepts. We need to migrate and simplify its functionality to the store.

---

## Things to Consider

- **Store architecture:** Single `$projects` store containing both projects list and active project state.
- **Event listeners:** Must subscribe to `ProjectCreatedEvent`, `ProjectUpdatedEvent`, and `ProjectDeletedEvent` to keep store synchronized with server state.
- **Event lifecycle:** Register event listeners when store initializes, clean up if needed (though in practice they'll live for the app lifetime).
- **Naming convention:** Follow project convention of prefixing store names with `$` (e.g., `$projects`).
- **Active project management:** Store must track the currently selected project and provide its display name for AppBar button. Can be done as a computed atom store `$activeProjectName`.
- **Default selection:** After load, if there is **_only 1 project available_**, that project must be automatically selected.
- **Store exports:** Avoid default exports — export named stores and action functions explicitly.
- **Migration path:** `ContentAppBar` project management logic will be removed once store is integrated.

---

## Implementation

> [!IMPORTANT]  

1. **Create the store file:**
   - Create `modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts`
   - Follow existing project patterns for store organization (see `sidebarWidgets.store.ts`)

2. **Design store structure:**
   - Use `map` from nanostores for complex state
   - Name main store with `$` prefix: `$projects`
   - Define TypeScript types for project data structure
   - Include fields: `projects`, `activeProjectId`

3. **Implement store actions:**
   - Create public setter functions for store mutations (those are just examples):
   - `loadProjects()` — fetch and populate projects from server
   - `setActiveProject(projectId)` — manage active project selection
   - `updateActiveProject()` — refresh projects list from server

4. **Integrate data fetching:**
   - Use existing API from `ContentAppBar` to load projects
   - Ensure projects are loaded once
   - Updated projects in response to server events, do not reload them

5. **Implement event listeners:**
   - Subscribe to `ProjectCreatedEvent` — reload projects when new project is created
   - Subscribe to `ProjectUpdatedEvent` — reload projects when project is modified
   - Subscribe to `ProjectDeletedEvent` — reload projects when project is removed
   - Register all listeners in store initialization
   - Probably no need for debounce

6. **Connect AppBar to store:**
   - Use `useStore` hook from `@nanostores/preact` in AppBar component
   - Use active project display name for project button label

7. **Implement default selection:**
   - Auto-select first project on initial load if no active project exists
   - Switch active project if the current one was removed

8. **Project selection dialog:**
  - Hook the simplest logic to update project list and selected dialog in the legacy dialog for now, once `ContentAppBar` logic removed.

**UI Behavior:**

- First project is automatically selected when page loads (if no prior selection and projects count is 1)
- Project button in AppBar displays active project's display name
- Clicking project button opens project selection dialog
- When projects are created/updated/deleted, store automatically reloads and updates UI (you can simulate this, not need to delete them)
- Active project persists across store updates (unless deleted)

---

## Acceptance Criteria

- [x] Project data is stored in global nanostores store
- [x] AppBar consumes active project name from store using `useStore` hook
- [x] First project is automatically selected on page load if no active project exists
- [x] Project button displays correct active project name
- [x] Store subscribes to all three project events (Created, Updated, Deleted)
- [x] Projects list automatically reloads when any project event fires
- [x] Active project selection persists across project reloads
- [x] If active project is deleted, store automatically selects another project
- [ ] No duplicate project loading requests across application
- [ ] Project loading logic removed from `ContentAppBar`

---

## Out of Scope

- Persisting active project selection to localStorage
- Project selection dialog implementation (already exists, just needs to use store)
- Error handling UI for failed project loads
- Project permissions/access control logic (done on backend)

reisfmb:
### 1.1 - No duplicate project loading requests across application

There are many places requesting projects directly (see `new ProjectListRequest`), and manipulating the active one via `ProjectContext`.

From my understanding it was out of the scope of this task to replace all `ProjectListRequest` and `ProjectContext` usage with the new projects store, but rather create the store and start using it in minimal context, and that's what I've done on `ProjectSelectionDialog`.

Note that I still left the `ProjectContext `manipulation there, with proper code comments to remember us to get rid of it once we fully migrate to use only the projects store.

### 1.2 - Project loading logic removed from ContentAppBar

I didn't touch `ContentAppBar` because we're not instantiating it anymore after the `AppBarElement` replacement of `ContentAppBar` (task #9279)

The only place still calling it is the `SettingsAppPanel` constructor, and I already added a comment there for future removal, but that depends on updating lib-admin-ui as well.

---

## 2 - Some extra comments and questions (reason for PR being draft)
a. Currently we only automatically set the active project if there's no project set as active and there's only one project loaded in the store. If none is active and there is more than one project, then we don't automayically set the active project! What should be the logic in this case?

b. I could've gone further and improved the `ProjectSelectionDialog` and `main.ts`, removing `ProjectListRequest` usage from both, and instead, getting projects from the store... I have two reasons for not doing that yet:
1) because I was unsure if that would be out of the scope
2) because there would be fricction with the `readonly` projects from the projects store (since most functions expect projects as `Project[]` not `Readonly<Project>[]`)

c. What should happen in case no project is selected?!

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/dialog/ProjectSelectionDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/dialog/ProjectSelectionDialog.ts
@@ -12,6 +12,7 @@ import {ProjectContext} from '../project/ProjectContext';
 import {Project} from '../settings/data/project/Project';
 import {ProjectHelper} from '../settings/data/project/ProjectHelper';
 import {ProjectListRequest} from '../settings/resource/ProjectListRequest';
+import {setActiveProject} from '../../v6/features/store/projects.store';
 
 export class ProjectSelectionDialog
     extends ModalDialog {
@@ -47,6 +48,8 @@ export class ProjectSelectionDialog
             itemView.onClicked((event: MouseEvent) => {
                 if (!event.ctrlKey && !event.shiftKey) {
                     if (itemView.isSelectable()) {
+                        setActiveProject(itemView.getProject());
+                        // TODO: Enonic UI - Remove once ProjectContext is replaced by projects store
                         ProjectContext.get().setProject(itemView.getProject());
                         this.close();
                     }
@@ -75,10 +78,13 @@ export class ProjectSelectionDialog
     close() {
         super.close();
 
+        // TODO: Enonic UI - Change to $isInitialized from projects store once ProjectContext is replaced by projects store
         if (!ProjectContext.get().isInitialized()) {
             const project: Project = this.getDefaultProject();
 
             if (project) {
+                setActiveProject(project);
+                // TODO: Enonic UI - Remove once ProjectContext is replaced by projects store
                 ProjectContext.get().setProject(project);
             } else {
                 Body.get().addClass('no-projects');
@@ -106,11 +112,11 @@ export class ProjectSelectionDialog
         this.mask();
 
         return new ProjectListRequest(true).sendAndParse().then((projects: Project[]) => {
-            this.setProjects(projects);
-            this.showItems();
+                this.setProjects(projects);
+                this.showItems();
         }).catch(DefaultErrorHandler.handle).finally(() => {
-            this.unmask();
-        });
+                this.unmask();
+            });
     }
 
     private showItems() {
--- a/modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
@@ -6,11 +6,11 @@ import {ProjectSelectionDialog} from '../../../../app/dialog/ProjectSelectionDia
 import {Store} from '@enonic/lib-admin-ui/store/Store';
 import {ShowIssuesDialogEvent} from '../../../../app/browse/ShowIssuesDialogEvent';
 import {useI18n} from '../../../../app/ui2/hooks/useI18n';
-import {$activeProjectName} from '../../store/projects.store';
 import {useStore} from '@nanostores/preact';
+import {$activeProjectName} from '../../store/projects.store';
 
 const AppBar = (): ReactElement => {
-    useStore($activeProjectName);
+    const activeProjectName = useStore($activeProjectName);
 
     return (
         <header className="bg-surface-neutral h-15 px-5 py-2 flex items-center gap-2.5 border-b border-bdr-soft">
@@ -22,7 +22,7 @@ const AppBar = (): ReactElement => {
                     ProjectSelectionDialog.get().open();
                 }}
                 aria-label={useI18n('wcag.appbar.project.label')}
-                label={$activeProjectName.get()}
+                label={activeProjectName}
             />
 
             <Button
--- a/modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
@@ -6,8 +6,12 @@ import {ProjectSelectionDialog} from '../../../../app/dialog/ProjectSelectionDia
 import {Store} from '@enonic/lib-admin-ui/store/Store';
 import {ShowIssuesDialogEvent} from '../../../../app/browse/ShowIssuesDialogEvent';
 import {useI18n} from '../../../../app/ui2/hooks/useI18n';
+import {$activeProjectName} from '../../store/projects.store';
+import {useStore} from '@nanostores/preact';
 
 const AppBar = (): ReactElement => {
+    useStore($activeProjectName);
+
     return (
         <header className="bg-surface-neutral h-15 px-5 py-2 flex items-center gap-2.5 border-b border-bdr-soft">
             <Button
@@ -18,9 +22,8 @@ const AppBar = (): ReactElement => {
                     ProjectSelectionDialog.get().open();
                 }}
                 aria-label={useI18n('wcag.appbar.project.label')}
-            >
-                Default Project
-            </Button>
+                label={$activeProjectName.get()}
+            />
 
             <Button
                 className="max-sm:hidden"
@@ -30,9 +33,8 @@ const AppBar = (): ReactElement => {
                     new ShowIssuesDialogEvent().fire();
                 }}
                 aria-label={useI18n('wcag.appbar.issues.label')}
-            >
-                Issues
-            </Button>
+                label="Issues"
+            />
 
             <IconButton
                 size="sm"
--- a/modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts
@@ -16,13 +16,13 @@ export const $projects = map<ProjectsStore>({
 });
 
 export function setActiveProject(project: Readonly<Project> | undefined): void {
-    const isProjectInStore = $projects.get().projects.some((p) => getProjectKey(p) === getProjectKey(project));
-    if (!isProjectInStore) throw new Error('Project not found in store');
-    $projects.setKey('activeProjectId', getProjectKey(project));
+    const existsInStore = $projects.get().projects.some((p) => getProjectId(p) === getProjectId(project));
+    if (!existsInStore) return;
+    $projects.setKey('activeProjectId', getProjectId(project));
 }
 
 export const $activeProject = computed($projects, (store) => {
-    return store.projects.find((p) => getProjectKey(p) === store.activeProjectId);
+    return store.projects.find((p) => getProjectId(p) === store.activeProjectId);
 });
 
 export const $activeProjectName = computed($activeProject, (activeProject) => {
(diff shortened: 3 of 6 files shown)


edloidas:
**1.1**
Okay, if it’s too much to remove, keep it as is. Just make sure we don’t make excessive requests when using the stores themselves.

**1.2**
So you’re suggesting removing `ContentAppBar` altogether rather than removing its functionality piece by piece? I’m not sure that’s a better approach, but if it’s too much work otherwise and your way is easier, then okay.

**2.a**
We show the dialog so that the user can select a project. It will be implemented within the dialog, e.g.

```jsx
<ProjectDialog open={activeProjectId == null && projects.length > 1} />
```

**2.b**
Not in scope. Ignore it. The project selection dialog is a separate task.

**3.c**
If there are no projects, we don’t select anything — just disable the project selection button and change its label.
If there’s exactly one project and no selection, we automatically select it after loading.
If there are two or more projects, we show the dialog so the user can make a selection manually.

reisfmb:
1.1
Yep, its quite a lot. On the store we only load once and then only at Project create/update events. On delete we manually delete the deleted project from the store (no need to reload).

1.2
Exactly, because after `AppBar` implementation we basically don't use `ContentAppBar` anymore.

That raises an UX concern: what should happen to the `AppBar` when we are outside of the content widget? We're essentially breaking the current behavior. Also the "tabs" are gone to (as we internally discussed).

2.(a,b,c) 
👍 

---

Tomorrow morning I'll fix conflicts and change from draft to ready!

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
```diff
@@ -19,7 +23,7 @@ const AppBar = (): ReactElement => {
                 }}
                 aria-label={useI18n('wcag.appbar.project.label')}
             >
-                Default Project
+                {$activeProjectName.get()}
```
edloidas: Better use `label` property of our Button component. More consistent.

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -17,6 +17,8 @@ export const $sidebarWidgets = map<WidgetsStore>({
 });
 
 export function setActiveWidget(widget: Readonly<Widget> | undefined): void {
+    const existsInStore = $sidebarWidgets.get().widgets.some((p) => getWidgetKey(p) === getWidgetKey(widget));
```
edloidas, line 24: Just a sm<i>aaa</i>ll change here: start with `is` for consistency with our code style. We drop `is`, `has`, etc only in prop names.

modules/lib/src/main/resources/assets/js/app/dialog/ProjectSelectionDialog.ts
```diff
@@ -47,6 +48,8 @@ export class ProjectSelectionDialog
             itemView.onClicked((event: MouseEvent) => {
                 if (!event.ctrlKey && !event.shiftKey) {
                     if (itemView.isSelectable()) {
+                        setActiveProject(itemView.getProject());
```
edloidas, line 51: Please, add annotation to the `ProjectContext` file itself.
```ts
/**
 * @deprecated The class due to be replaced by the projects.store
 */
export class ProjectContext {
```

edloidas (commented):
And please,

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
```diff
@@ -6,8 +6,12 @@ import {ProjectSelectionDialog} from '../../../../app/dialog/ProjectSelectionDia
 import {Store} from '@enonic/lib-admin-ui/store/Store';
 import {ShowIssuesDialogEvent} from '../../../../app/browse/ShowIssuesDialogEvent';
 import {useI18n} from '../../../../app/ui2/hooks/useI18n';
+import {$activeProjectName} from '../../store/projects.store';
+import {useStore} from '@nanostores/preact';
 
 const AppBar = (): ReactElement => {
+    useStore($activeProjectName);
```
edloidas: Not technically correct. You should do
```ts
const activeProjectName = useStore($activeProjectName);
// ...
label={activeProjectName}
```

modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts
```diff
@@ -0,0 +1,104 @@
...
+export const $projects = map<ProjectsStore>({
+    projects: [],
+    activeProjectId: undefined,
+});
+
+export function setActiveProject(project: Readonly<Project> | undefined): void {
+    const isProjectInStore = $projects.get().projects.some((p) => getProjectKey(p) === getProjectKey(project));
+    if (!isProjectInStore) throw new Error('Project not found in store');
+    $projects.setKey('activeProjectId', getProjectKey(project));
+}
```
edloidas, line 22: Put under the stores

modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts
```diff
@@ -0,0 +1,104 @@
...
+export const $isInitialized = computed($projects, (store) => {
+    return store.activeProjectId === undefined || store.projects.length > 0;
+});
+
+//
+// * Utilities
+//
+export function getProjectKey(project: Readonly<Project> | undefined): string | undefined {
+    return project?.getName();
+}
```
edloidas, line 43: Maybe it's `getProjectId` if it's used for setting ID?

modules/lib/src/main/resources/assets/js/v6/features/store/sidebarWidgets.store.ts
```diff
@@ -17,6 +17,8 @@ export const $sidebarWidgets = map<WidgetsStore>({
...
+    if (!existsInStore) throw new Error('Widget not found in store');
```
edloidas: It's not really correct to throw an error here, because you are not handling it in the code later. And it's not something to be handled here. Just return.

modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const $projects = map<ProjectsStore>({
+    projects: [],
+    activeProjectId: undefined,
+});
+
+export function setActiveProject(project: Readonly<Project> | undefined): void {
+    const isProjectInStore = $projects.get().projects.some((p) => getProjectKey(p) === getProjectKey(project));
+    if (!isProjectInStore) throw new Error('Project not found in store');
```
edloidas: Just return. No need to throw error. You expect to (and actually do) pass a valid project, that you took from `$projects`. There should not be any error there.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/layout/AppShell/AppBar.tsx
reisfmb: Yep, sure... Idk why I did it that way. Probably was testing something and forget to change to the proper way before opening the PR. Thx.

edloidas (approved):

Issue #4185, opened 2022-02-01
Unpublish Item dialog - dependent items are not displayed when content with children are scheduled


sgauruseu:
1. Select a folder with children and schedule the tree
2. Click on `Unpublish` button in the toolbar
BUG - `Show dependent items` is not displayed in the modal dialog and  `Unpublish` button does not contain the number of items
![image](https://user-images.githubusercontent.com/3728712/151953041-7e205dba-d776-467f-869b-9c8c03c12534.png)

3. Click on `Unpublish` button in the modal dialog.  `11 items are unpublished` message appears  
![image](https://user-images.githubusercontent.com/3728712/151953142-7cb42116-32f8-4e0b-a808-debbcd43a167.png)

PR #9307, opened 2025-10-31
Enonic UI: Add Filter Panel component #4185

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/4187

PR #9316, opened 2025-11-03
Enonic UI: Update confirmation dialog focus #9315

### LINKED ISSUE #9315: Enonic UI: Update confirmation dialog focus
Update focus behavior, so that:
1. We focus on input once it's rendered
2. We switch focus to Confirm action, once the input become valid / readonly.

skjulestad:
Replaced autoFocus with focus based in useRef on render and validation.


edloidas (approved):
Most certainly may lead to race conditions. Focus must be handled in one place to stay in sync, so that only one element receives focus at a time.

PR #9309, opened 2025-10-31
Enonic UI: Externalize Issues Stats Data Management #9289

reisfmb:
Didn't implement the `issues` store, just the `issuesStats` one, since it was the only one needed for the `AppBar`.

PR #9325, opened 2025-11-03
Enonic UI: Add LocalStorage sync for stores #9292

### LINKED ISSUE #9292: Enonic UI: Add LocalStorage sync for stores
Create utility functions to synchronize nanostores state with `LocalStorage`, enabling persistent state across page reloads.

---

## Rationale

- **State persistence:** Critical application state (like active project) should survive page reloads.
- **Generic solution:** Multiple stores will need LocalStorage sync — create reusable utilities instead of duplicating logic.
- **Selective sync:** Not all store properties need persistence — utility must support property filtering, or custom encode/decode functions.
- **Migration path:** Replace `ProjectContext.ts` singleton pattern with modern functional approach.

---

## References

- [`ProjectContext.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/app/project/ProjectContext.ts) — existing LocalStorage implementation to replace
- [`projects.store.ts`](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/v6/features/store/projects.store.ts) — first store to integrate with LocalStorage sync

---

## Things to Consider

- **Utility location:** Create in `modules/lib/src/main/resources/assets/js/v6/features/utils/` or similar location for shared utilities.
- **Property filtering:** Support selective sync — e.g., sync only `activeProjectId` from projects store, not the entire `projects` array.
- **Type safety:** Utility functions should preserve TypeScript types.
- **Storage key format:** Use consistent naming pattern. May need "brand" prefix for that, e.g. `enonic-cs-${storename}`.
- **Error handling:** LocalStorage can fail (quota exceeded, private mode) — handle gracefully without breaking app, as if there were no data.
- **Initial load:** Restore state from LocalStorage on store initialization. Good idea would be to attach store sync right after all the stores created, before we list getters and setters.
- **Bidirectional sync:** Changes in store → LocalStorage, LocalStorage → store on init.

---

## Implementation

> [!IMPORTANT]  

1. **Create utility file:**
   - Create `modules/lib/src/main/resources/assets/js/v6/features/utils/stores.ts`

2. **Implement core utility:**
   - `syncStore(store, storageKey, props)` — main function
   - Subscribe to store changes and persist selected properties to LocalStorage
   - On init, restore values from LocalStorage if they exist
   - Handle LocalStorage errors gracefully (try-catch)

3. **Type definitions:**
   - Type-safe property selection

4. **Integrate with projects store:**
   - Import utility in `projects.store.ts`
   - Call `syncStore($projects, 'projects', {encode, decode})`
   - Sync only `activeProjectId`, ignore `projects` array
   - Initialize sync after store creation

5. **Migration from ProjectContext:**
   - Remove LocalStorage logic from `ProjectContext.ts` once projects store integration is complete (or the file itself)

6. **Testing:**
   - Select a project in UI
   - Reload page
   - Verify active project persists (button shows correct project name)
   - Check LocalStorage in DevTools for stored value

---

## Acceptance Criteria

- [x] Generic LocalStorage sync utility created and exported
- [x] Utility supports selective property sync (not entire store)
- [x] Utility handles LocalStorage errors gracefully (try-catch)
- [x] Projects store integrates with utility to persist `activeProjectId`
- [x] Active project persists across page reloads
- [x] LocalStorage key follows consistent naming pattern
- [x] TypeScript types of the stores preserved through utility functions
- [x] No console errors if LocalStorage is unavailable

---

## Out of Scope

- Syncing of non-serializable via `JSON.stringify` objects (only primitive values for now)
- Cross-tab synchronization (listening to `storage` events)
- Encryption or obfuscation of stored values
- Migration of existing LocalStorage values from old implementation

---

## Technical Constraints

- Must work with nanostores `map` stores
- Must not break if LocalStorage is unavailable (private browsing, quota)
- Must preserve TypeScript types
- Storage keys should be namespaced to avoid conflicts (e.g., `contentstudio.*`)

---

## Examples

Very basic example of store sync function with throttle, custom mapper, encode, decode, and storage selection. Must be used as a base, but improved and polished:

```ts
import { BaseDeepMap, DeepMapStore, MapStore, ReadableAtom } from 'nanostores';
import { normalize } from './text';

type SyncStore<T extends BaseDeepMap> = MapStore<T> | DeepMapStore<T> | ReadableAtom<T>;

interface SyncOptions<T> {
  storage?: Storage;
  encode?: (data: T) => string;
  decode?: (raw: string) => Partial<T>;
  mapValue?: (stored: Partial<T>, current: T) => T;
  mergeStrategy?: 'replace' | 'merge';
  throttleMs?: number;
}

function deepMerge<T>(target: T, source: Partial<T>): T {
  if (source === null || typeof source !== 'object') {
    return (source ?? target) as T;
  }

  const result = structuredClone(target);
  
  for (const key in source) {
    const sourceValue = source[key];
    const targetValue = result[key as keyof T];
    
    if (sourceValue !== undefined) {
      if (typeof sourceValue === 'object' && sourceValue !== null && 
          typeof targetValue === 'object' && targetValue !== null &&
          !Array.isArray(sourceValue)) {
        result[key as keyof T] = deepMerge(targetValue, sourceValue);
      } else {
        result[key as keyof T] = sourceValue as T[keyof T];
      }
    }
  }
  
  return result;
}

function createThrottle<T extends (...args: any[]) => void>(
  fn: T,
  delay: number
): T & { cancel: () => void } {
  let timeoutId: ReturnType<typeof setTimeout> | null = null;
  let lastArgs: Parameters<T> | null = null;

  const throttled = function (this: any, ...args: Parameters<T>) {
    lastArgs = args;

    if (timeoutId === null) {
      timeoutId = setTimeout(() => {
        if (lastArgs) {
          fn.apply(this, lastArgs);
        }
        timeoutId = null;
        lastArgs = null;
      }, delay);
    }
  } as T & { cancel: () => void };

  throttled.cancel = () => {
    if (timeoutId !== null) {
      clearTimeout(timeoutId);
      timeoutId = null;
      lastArgs = null;
    }
  };

  return throttled;
}

export function syncStore<T extends BaseDeepMap>(
  store: SyncStore<T>,
  storeName: string,
  options: SyncOptions<T> = {}
): () => void {
  const {
    storage = localStorage,
    encode = (data) => JSON.stringify(data || {}),
    decode = (raw) => JSON.parse(raw),
    mapValue,
    mergeStrategy = 'merge',
    throttleMs = 50
  } = options;

  const storeKey = `enonic-cs-${normalize(storeName)}`;

  const rawData = storage.getItem(storeKey);
  const storedData = rawData ? decode(rawData) : {};
  const currentData = store.get();

  let mergedData: T;
  if (mergeStrategy === 'replace') {
    mergedData = { ...currentData, ...storedData } as T;
  } else {
    mergedData = deepMerge(currentData, storedData);
  }

  const finalData = mapValue ? mapValue(storedData, currentData) : mergedData;
  store.set(finalData);

  const writeToStorage = (data: T) => {
    const encoded = encode(data);
    storage.setItem(storeKey, encoded);
  };

  const throttledWrite = createThrottle(writeToStorage, throttleMs);

  const unsubscribe = store.subscribe(throttledWrite);

  return () => {
    throttledWrite.cancel();
    unsubscribe();
  };
}
```

reisfmb:


reisfmb:
@edloidas 

Suggestion of merging order for epic branch `issue-9278`:

1. #9302 
2. #9309 
3. #9325

edloidas (approved):
Had to spend some time improving the current implementation. Can be considered done, but needs to wait for previous PRs to be merged first.

reisfmb:
Here are my thoughts on the improvements you did on synching:

## 1 - Error handling
Amazing 👏 

I was not aware of that many edge cases for when local storage could fail or was not available... Could not have done better myself!

## 2 - Tab synching
Didn't think of that... nice addition as well!

## 3 - Partial store synching
After some thought, in my opinion, the best approach here would be the following:

We de not export `syncStore`, but rather export wrapper functions that will manually set encode / decode functions based on the nanostore type (Atom, Map, DeepMap).

### Atom
For Atom, I think the encoder / decoder should be the default ones:

```ts
encode = (data: V): string => JSON.stringify(data),
decode = (raw: string): V => JSON.parse(raw) as V,
```

### Map
For Map, you've already done like what I had in mind.

### DeepMap
For DeepMap, my suggestion would be to have the decode as in the `syncMapStore`, but the encode like so:
```ts
 const encode = isPartialSync
        ? (data: M): string => {
              let partial: Partial<M> = {};
              
              for (const key of keys) {
                    const pathValue = getPath(data, key);
                    partial = setPath(partial, key, pathValue)
              }
              
              return JSON.stringify(partial);
          }
        : (data: M): string => JSON.stringify(data);
```

Where `getPath` and `setPath` are nanostore functions.

In this specific case, the `options.keys` would be of type: `AllPaths<M>[]`, which provides proper typings for all paths in the DeepMap store.

---

I think its important to now allow custom encoder / decoder, because if we do it like so, the data on the storage might not have the same structure as the data of the store (or a path of it)... I mean, nothing blocks me to do something like:

```ts
syncStore($myMapStore, 'my-map-store', {
    encode: (data) => JSON.stringify({"a-non-existing-key-on-$myMapStore": 123}),
});
```

In my point of view, allowing that is problematic because, it not only stores in the storage a data that differs from the store, as it can also change the data in the store (`store.set` on initialValue and tabSync).

## 4 - Store subscription

The current subscription to the store is done by `store.listen`. That is triggered on any store change... Now, if our store is Map or DeepMap, we only want to trigger our callback (`throttledWrite`) when a specific key / path changes. Nanostores have this `listenKeys` that we can use! The problem is that implementing that change is not straightforward, and would require quite some changes in code. 

## 5- Other less relevant points

- If we only sync a key of a Map store or a path of a DeepMap store, I guess it'd make sense to consider that key / path on the storage key as well, since the data in the storage is not the whole store itself, but rather a section of it.
- Remove console logs such as `console.log('removeFromStorage <${storageKey}>');`?!

reisfmb:
Now, my thoughts on the changes you did on `projects.store.ts`:

If my point 5 [here](https://github.com/enonic/app-contentstudio/pull/9325#issuecomment-3500600648) is not implemented, I'd rename the localStorage key:

```ts
syncMapStore($projects, 'projects.activeProjectId', {
    keys: ['activeProjectId'],
    loadInitial: true,
    syncTabs: true,
});
```

I also suggest to move this:
```ts
const projectIdFromUrl = getProjectIdFromUrl();
if (projectIdFromUrl) {
    $projects.setKey('activeProjectId', projectIdFromUrl);
}
```

To the scope of`updateActiveProject`:
```ts
function updateActiveProject(): void {
    if ($activeProject.get()) {
        return;
    }

    const {projects} = $projects.get();

    if (projects.length === 1) {
        setActiveProject(projects[0]);
        return;
    }

    const projectFromUrl = projects.find((p) => getProjectId(p) === getProjectIdFromUrl());
    if (projectFromUrl) {
        setActiveProject(projectFromUrl);
    }
}
```

Consider the case where we have > 2 projects. If we don't do that change above, upon deleting the selected one, none will be selected as the active one.

edloidas:
**Regarding comments:**

3. Partial store synching

> We de not export syncStore, but rather export wrapper functions that will manually set encode / decode functions based on the nanostore type (Atom, Map, DeepMap).

Not really sure how that would work.

4. Store subscription

> The current subscription to the store is done by store.listen. That is triggered on any store change

Yeah, I thought about this too. It's a balance between simplicity and full coverage of edge cases — and there are *a lot* of them. I haven't found a simple way to organize the code without adding significantly more logic to trigger updates only on specific key changes. A throttle should be more than enough to avoid performance issues (and our stores are simple enough anyway). I'd rather keep things simple. If we go deeper here, it should probably be a separate library — something we could hack on in spare time. The devs have their own [implementation](https://github.com/nanostores/persistent) of localStorage sync, but I don't find it that flexible.

5. Other less relevant points

> make sense to consider that key / path on the storage key as well

Too complex. DeepMaps are powerful, but very tricky to cover properly. I don't think we should try to use them with partial sync at all right now. They're also difficult to cover with types. So the approach is: we stick to atoms and maps primarily. DeepMaps are treated like regular maps for now.

> Remove console

Yep, just leftovers from debugging.

6.

> Consider the case where we have > 2 projects. If we don't do that change above, upon deleting the selected one, none will be selected as the active one.

Good point — we should add that as you suggested. But this shouldn't replace the `projectIdFromUrl` logic above, because the example won't run if we reload the page with a different URL when an active project is already set.

**Summarizing:**

* We drop DeepMap support for now and don't worry about key-level subscriptions. Too long to implement properly, low immediate value. Could be a separate side-project library later.
* Remove logs.
* Add your `projectIdFromUrl` change.

edloidas:
Great work, Bruno

PR #9368, opened 2025-11-11
Enonic UI: Create emulators dropdown selector #9364

### LINKED ISSUE #9364: Enonic UI: Create emulators dropdown selector

reisfmb:


modules/lib/src/main/resources/assets/js/v6/features/store/liveviewWidgets.store.ts
```diff
@@ -0,0 +1,82 @@
...
+    const {widgets} = $liveviewWidgets.get();
+
+    setActiveWidget(widgets[0]);
+}
+
+function sortWidgets(widgets: Widget[]): Widget[] {
+    return widgets.sort((a, b) => {
+        const orderA = a.getConfig().getProperty('order');
+        const orderB = b.getConfig().getProperty('order');
+        return (parseInt(orderA) ?? 999) - (parseInt(orderB) ?? 999);
```
edloidas: Can use `Number.MAX_SAFE_INTEGER ` instead of `999`

modules/lib/src/main/resources/assets/js/app/view/context/widget/emulator/EmulatorDevice.ts
```diff
@@ -2,17 +2,15 @@ import {i18n} from '@enonic/lib-admin-ui/util/Messages';
...
 import {ObjectHelper} from '@enonic/lib-admin-ui/ObjectHelper';
 import {Equitable} from '@enonic/lib-admin-ui/Equitable';
+import {LucideIcon, Monitor, Smartphone, Tablet, TvMinimal} from 'lucide-react';
 
 export enum DeviceType {
     MONITOR,
     TABLET,
-    MOBILE
+    MOBILE,
 }
```
edloidas: Is this enum actually used elsewhere? If not, assume removing it.

edloidas (commented):
* Remove `EmulatorDropdown`.
* Remove `EmulatorContext`.
* Adjust `EmulatedEvent` to use a native custom event. For example:

  ```ts
  const event = new CustomEvent<EmulatorDevice>('emulated', {
    detail: {
      device: EmulatedDevice,
    },
    bubbles: true,
    cancelable: true,
  });
  ```
* I’d suggest adding a small utility file, `events.ts`, to create, dispatch, listen to, and subscribe to custom events.
  `listen`/`subscribe` should return an unsubscribe handler so we can clean it up on unmount or in `.onRemove` for legacy elements.
* Then, instead of using `EmulatorContext`, we can simply listen for the `EmulatedEvent`.

reisfmb:
@edloidas Nice suggestion!

Please check last commit.

I suggest, for simplicity, to remove the deprecated files from this epic as the last commit on the epic branch after merging all the sub tasks PR's.

edloidas (approved):

PR #9399, opened 2025-11-19
Enonic UI: TreeToolbar #9390

### LINKED ISSUE #9390: Enonic UI: TreeToolbar
Migrate the current `ListBoxToolbar` used in `ContentBrowsePanel` to a new component.

* Do **not** modify the existing implementation in `lib-admin-ui`.
* New name: **`TreeListToolbar`**.
* Suggested location:
  `modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/`
* Enonic UI checkbox already supports *partial selection* — reuse that API/state.
* Introduce **`content.store.ts`** to handle selection:
  * Preferred: move selection logic fully into the store.
  * If not feasible initially: mirror the current `ContentsTreeGridRootList` behaviour.

### UI Examples

Toolbar
Image

Selections reference
Image

---

## Acceptance Criteria

* `TreeListToolbar` replaces `ListBoxToolbar` usage in `ContentBrowsePanel`.
* Visual behaviour matches existing UI and the selection examples.
* Checkbox supports full / partial selections.
* Selection state is handled via `content.store.ts`:
  * moved fully to the store **or**
  * mirrored from legacy selection logic without regressions.
* No changes are made in `lib-admin-ui`.

---

## Notes

* Existing selection logic should continue to work exactly the same.

reisfmb:


reisfmb:
@ashklianko 

Please check if the store approach makes sense to you, since you'll continue coding on it while migrating the `ContentBrowsePanel`

@edloidas 

Questions (reason for PR being in draft)

1. The current implementation of the indeterminate checkbox state differs from the design. In our version, the checkbox displays a filled background when in the indeterminate state, whereas the design shows no filled background. Should we care about that?

image
image

2. In the current implementation, the selection only occurs when we click on the actual box. If we click over the 'selected' label, it toggles the visualization of the selected items in the tree grid. I did not implement that due to how the click works our new `Checkbox` component. Should we care about that?

ashklianko:
Copied the message from Slack:

```
type ContentStore = {
    isLoading: boolean;
    contents: Readonly<ContentSummaryAndCompareStatus>[];
    activeContent: Readonly<ContentSummaryAndCompareStatus | null>;
    selectedContentIds: Readonly<string>[];
};
```

one thing that bothers me is that we can't have simultaneously selected and active/highlighted item, but this signatures implies that it can.
A highlighted item is basically a selected item and it behaves as one except the part with the selection mode. All the selection listeners react to a highlighted item as to a selected one. I would instead use this :
```

type ContentStore = {
    isLoading: boolean;
    contents: Readonly<ContentSummaryAndCompareStatus>[];
    selectionMode: 'multiple' | 'highlight';
    selectedContentIds: Readonly<string>[];
}; 
```

edloidas:
@ashklianko How would selection mode benefit us here? It feels more like over-engineering in an attempt to replicate our old patterns. If we stick with selection mode (which I don’t think is the right choice), then it should be `selectionMode: 'multiple' | 'single'`. Meanwhile, having active and selected elements aligns better with the new approach and the new UI state management.

The current content and mode can always be derived, for example:
`$current = $selected.at(0) ?? $active`
And if you need mode:
`$mode = $selected.length > 1 ? 'multiple' : 'single'`;

ashklianko:
@edloidas

A few points to clarify why I believe the current store shape is misleading and creates unnecessary ambiguity:

- **It gives the false impression that we can have both `selected` and `active` items simultaneously.**  
  In our actual behavior, the “highlighted” item *is* the selected item — it only differs by how its checkbox is rendered. Functionally, listeners treat it exactly the same.

- **We already use `active` in the UI libs to mean the visually focused element.**  
  Re-introducing `activeContent` as a separate concept in the store risks mixing up “focused” vs “selected” states. These two are not the same in our UI model.

- **Do we really want to keep a highlighted item *outside* of the actual selection array?**  
  A highlighted item still behaves as a selected item: selection listeners, dependent panels, etc., expect that a “highlight” is a proper selection event.  
  If we store it separately, we either:  
  - duplicate the ID (in `activeContent` and `selectedContentIds`)  
  - or constantly manage moving it between the two fields.  
  Both approaches introduce complexity for no functional gain.

- **Highlighting affects only the checkbox visualization, not selection semantics.**  
  It's simply a rendering difference — not a separate conceptual state.

- **Tree-list already has `single` and `multiple` selection modes.**  
  Using `activeContent` as a separate selection state makes the terminology inconsistent with the component that is supposed to drive the behavior.  
  Even if we modify the UI logic later, aligning naming and semantics with the existing selection mechanisms avoids confusion.

edloidas:
@reisfmb Ignore the checkbox visuals for now — we’ll update them when we have time. We’ll adjust those and a few other icons, then distribute them as part of Enonic UI.

Regarding the label click: the label must keep the new behaviour and simply stay connected to the checkbox. **However:**
- Add an `aria-label` property to the checkbox (same value as the current label).
- If none are selected: show both the checkbox and the label.
- If some are selected:
  - Don’t pass the label text, keep only the aria property for screen readers.
  - Show a Toggle next to the checkbox instead.

The Toggle will use `aria-label` `"Show selected only"` / `"Show all"`, while the visible label text will be just `"Selected (x)"`.

The Toggle button will change background on click — try switching the `(x)` background to `surface-secondary`.

Toggle design will likely be updated soon.

PR #9424, opened 2025-11-20
Enonic UI: replace ContentUnpublishDialog #9423

### LINKED ISSUE #9423: Enonic UI: replace ContentUnpublishDialog
Replace showUnPublishConfirmationDialog dialog.

skjulestad:
Replaced ContentUnpublishConfirmation dialog.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/publish/ContentUnpublishDialog.ts
+++ b/modules/lib/src/main/resources/assets/js/app/publish/ContentUnpublishDialog.ts
@@ -8,7 +8,6 @@ import {ContentUnpublishPromptEvent} from '../browse/ContentUnpublishPromptEvent
 import {CompareStatus} from '../content/CompareStatus';
 import {ContentId} from '../content/ContentId';
 import {ContentSummaryAndCompareStatus} from '../content/ContentSummaryAndCompareStatus';
-import {ContentIds} from '../content/ContentIds';
 import {DependantItemsWithProgressDialogConfig} from '../dialog/DependantItemsWithProgressDialog';
 import {ResolveUnpublishRequest} from '../resource/ResolveUnpublishRequest';
 import {UnpublishContentRequest} from '../resource/UnpublishContentRequest';
@@ -62,7 +61,7 @@ export class ContentUnpublishDialog
         if (this.isSiteOrMultipleItemsToUnPublish()) {
             this.showUnPublishConfirmationDialog();
         } else {
-            this.doUnPublish();
+            this.doUnPublish(this.getContentToUnpublishIds());
         }
     }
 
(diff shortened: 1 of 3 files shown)


modules/lib/src/main/resources/assets/js/app/publish/ContentUnpublishDialog.ts
```diff
@@ -70,19 +71,26 @@ export class ContentUnpublishDialog
...
-        if (!this.unPublishConfirmationDialog) {
-            this.initUnPublishConfirmationDialog();
-        }
-
-        this.unPublishConfirmationDialog.setValueToCheck('' + this.countTotal()).open();
-    }
+        const totalToUnpublish = this.countTotal();
+        const contentIds = ContentIds.create()
+            .fromContentIds(this.getContentToUnpublishIds())
+            .build();
```
edloidas: But why? You convert `ContentId[]` to `ContentIds` for what? 
> BTW, you can just do `ContentIds.from(this.getContentToUnpublishIds())` for shorter version.

modules/lib/src/main/resources/assets/js/app/publish/ContentUnpublishDialog.ts
```diff
@@ -70,19 +71,26 @@ export class ContentUnpublishDialog
...
+
+
+        this.unPublishConfirmationDialog = new DialogPresetConfirmDeletePreset({
+            open: true,
+            title: i18n('dialog.unpublish.confirm.title'),
+            description: i18n('dialog.unpublish.confirm.subtitle'),
+            expected: totalToUnpublish,
+            onConfirm: () => {
+                const selectedIds = contentIds.map(id => id);
+                this.doUnPublish(selectedIds, totalToUnpublish);
```
edloidas: Seems just to convert it back to `ContentId[]` from `ContentIds`?

modules/lib/src/main/resources/assets/js/app/publish/ContentUnpublishDialog.ts
```diff
@@ -126,10 +134,13 @@ export class ContentUnpublishDialog
         });
     }
 
-    private doUnPublish(): void {
+    private doUnPublish(selectedIds: ContentId[] = this.getContentToUnpublishIds(), total?: number): void {
```
edloidas: So that you end up with the same `ContentId[]` that came from `this.getContentToUnpublishIds()`, either default, or from confirmation, but with extra steps.

edloidas (changes requested):
Just keep `contentIds` inside the `doUnPublish`, r remove the defaults and pass required arguments on every call.

edloidas (approved):

PR #9436, opened 2025-11-21
Enonic UI: Styling issues in Filter Panel/TreeGrid/Split Panel #9435

### LINKED ISSUE #9435: Enonic UI: Styling issues in Filter Panel/TreeGrid/Split Panel
- [x] Remove search icon from the top most text input filter in the filter panel
- [x] Add search icon for filterable buckets search inputs
- [x] Add a gap according to design in filterable buckets, between title-<missing gap>-selected checkboxes-<missing gap>-search input
- [x] Fix SplitPanel splitters appearance to be aligned with general design
- [x] (→ Enonic UI) Hovered and active tree row has the same color, so it's impossible to say if it's selected or just hovered expept the status color.
- [x] Icon colors of active items in emulator must always be "white" (use alt color). Same for the icon in the trigger button, when it's active. Now they are "black" in dark theme. 
- [x] Select All checkbox must not change its name after it becomes checked. It's a checkbox, where the label indicates state, and the checkbox itself shows if the state is on or off. It's also better to avoid "action" in label. Simple "All" or "All items" should be enough. Selected count would also go to toggle button to show selected only.

ashklianko:


ashklianko:
Need to fix one more issue

PR #9497, opened 2025-11-30
Enonic UI: DeleteDialog #9491

### LINKED ISSUE #9491: Enonic UI: DeleteDialog
Create a new DeleteDialog component to replace the legacy `ContentDeleteDialog`. The new dialog should handle both Archive and Delete actions using a unified interface with Archive as the primary action and Delete available through a dropdown menu.

---

## Rationale

- Migrate delete/archive functionality to modern React components without LegacyElement wrappers
- Unify Archive and Delete actions into a single dialog with consistent UX
- Align with the established v6 dialog patterns (PublishDialog as reference)

---

## References

- **Reference Implementation:** `modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/publish/PublishDialog.tsx`
- **Button Group Pattern:** `modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx`
- **Confirmation Pattern:** `modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/DialogPreset.tsx`
- **Legacy Implementation:** `modules/lib/src/main/resources/assets/js/app/remove/ContentDeleteDialog.ts`

---

## Things to Consider

- Archive is the default/primary action, Delete should appear on the right as Link button
- LinkButton should be a `/shared` variant of regular button with styled text (regular weight and underline style).
- Archive button should use solid style button with black background
- Delete button action text must use error color styling (indicating destructive action)
- Button group should visually match the Toolbar Publish action group (SplitActionButton pattern)
- Follow existing v6 dialogs visual design (no specific Figma design for this dialog)
- Confirmation dialogs (DialogPresetConfirmDelete) must be called from DeleteDialog when deleting multiple items or sites
- All dialog components must be native React - no LegacyElement wrappers allowed
- Socket events (`$contentArchived`, `$contentDeleted`) should update dialog state

---

## Implementation

> [!IMPORTANT]
> This is not a step-by-step guide — it's a functional checklist ordered logically.

1. Create `deleteDialog.store.ts` with nanostores for state management (similar to `publishDialog.store.ts`)
2. Create `DeleteDialog.tsx` in `modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/`
3. Implement dialog structure following PublishDialog pattern:
   - `Dialog.Root` > `Dialog.Portal` > `Dialog.Overlay` > `Dialog.Content`
   - `Dialog.DefaultHeader` for title and close button
   - `SelectionStatusBar` for status/error display
   - `Dialog.Body` with `ContentItemCheckable` components
   - `Dialog.Footer` with action buttons
4. Create two buttons: Delete (link style) + Archive (primary):
   - Use SplitActionButton pattern
   - Apply solid style to Archive button
   - Apply error color to Delete menu action text
5. Implement confirmation flow using `ConfirmationDialog` and `DialogPresetConfirmDelete`
6. Register DeleteDialog in `BrowseAppShell.tsx`
7. Update toolbar Archive action to open the new DeleteDialog
8. Preserve existing archive/delete request logic (ArchiveContentRequest, DeleteContentRequest)

**UI Behavior:**

- Archive button is the primary action (main clickable area) with error color background
- Dropdown chevron has matching error color background
- Delete action in dropdown menu uses error color for text
- Both actions should show dependencies (inbound references) before confirmation
- Optional message input for archive action (same as current behavior)
- Numeric gate confirmation (type count) for multi-item or site deletion

---

## Acceptance Criteria

- [x] DeleteDialog is a pure React component without LegacyElement wrappers
- [x] Dialog follows existing v6 dialogs visual design
- [x] Archive is the primary action, Delete is in dropdown menu
- [x] Archive button and dropdown use error color background
- [x] Delete menu action text uses error color
- [x] Confirmation dialogs (DialogPresetConfirmDelete) are called when appropriate
- [x] Current delete/archive logic is preserved (requests, progress, notifications)
- [x] Dialog responds to socket events for real-time updates
- [x] Legacy ContentDeleteDialog can be deprecated after migration

## References

Image

skjulestad:
Migrated new dialog for archive / delete with new confirmation dialog as well.
image


modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+                                noPermissions: {count: 0, onExclude: () => {}},
+                                inbound: {
+                                    count: inboundCount,
+                                    onIgnore: () => ignoreDeleteInboundDependencies(),
+                                },
+                            }}
+                        />
+
+                        <Dialog.Body className="flex flex-col gap-y-10 py-5">
+                            <div className="flex flex-col gap-y-2.5">
```
edloidas: No need for this div wrapper.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+                onConfirm={() => {
+                    setConfirmOpen(false);
+                    void executeDeleteDialogAction(confirmAction);
+                    setConfirmAction('delete');
+                }}
+                onCancel={() => {
+                    setConfirmOpen(false);
+                    setConfirmAction('delete');
+                }}
+            />
```
edloidas: You add DeleteDialog to the root of the app and always render it. But confirmation should not be always present in the DOM. Render it only when you require. Might not even need a separate Portal and overlay for it.

modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -358,19 +359,11 @@ async function startApplication() {
...
-
-        contentDeleteDialog
-            .setContentToDelete(event.getModels())
-            .setYesCallback(event.getYesCallback())
-            .setNoCallback(event.getNoCallback())
-            .open();
+        openDeleteDialog(event.getModels(), {
+            onYes: event.getYesCallback(),
+            onNo: event.getNoCallback(),
+        });
```
edloidas, line 366: 1. We should not use callbacks anymore. If we need to do something on dialog visibility change, we'd rather subscribe to store.
2. I don't actually see that the yes/no callbacks are used or set anywhere. You can just drop their support.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+const DELETE_DIALOG_NAME = 'DeleteDialog';
+
+export const DeleteDialog = (): ReactElement => {
+    const {open, loading, failed, items, dependants, inboundIgnored} = useStore($deleteDialog);
+    const ready = useStore($deleteDialogReady);
+    const total = useStore($deleteTotalItems);
+    const hasSite = useStore($deleteHasSite);
+    const inboundIds = useStore($deleteInboundIds);
+    const inboundSet = useMemo(() => new Set(inboundIds), [inboundIds]);
+    const isInbound = (id: string) => inboundSet.has(id);
```
edloidas: Can pass content directly and do `item.getContentId().toString()` inside function.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+                                            onClick={() => {
+                                                new EditContentEvent([item]).fire();
+                                            }}
+                                        />
+                                    ))}
+                                </ul>
+                            </div>
+
+                            <div className={cn('flex flex-col gap-y-7.5', dependants.length === 0 && 'hidden')}>
+                                <Separator className="" label={dependantsLabel} />
```
edloidas: Add `pr-1` as it's too close to scroll

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/deleteDialog.store.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    const remaining = pendingIds.filter(id => !ids.has(id));
+    if (remaining.length === 0) {
+        const total = state.pendingTotal || pendingIds.length;
+        if (pendingAction === 'archive') {
+            const message = total > 1
+                ? i18n('dialog.archive.success.multiple', total)
+                : i18n('dialog.archive.success.single', state.pendingPrimaryName ?? '');
+            NotifyManager.get().showSuccess(message);
```
edloidas, line 329: You can call `showSuccess` as we already have those utilities, and planning to move from `NotifyManager`.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/deleteDialog.store.ts
```diff
@@ -0,0 +1,364 @@
...
+    inboundTargets: ContentId[];
+    inboundIgnored: boolean;
+    archiveMessage: string;
+    pendingAction?: DeleteAction;
+    pendingIds: string[];
+    pendingTotal: number;
+    pendingPrimaryName?: string;
+    yesCallback?: () => void;
+    noCallback?: () => void;
+};
```
edloidas, line 38: Store is too complex. You are mixing base properties with some properties to depend your actions and responses on. Functions that you use should be pure and do not depend on external state, only on the arguments.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/deleteDialog.store.ts
```diff
@@ -0,0 +1,364 @@
...
+//
+
+$deleteDialog.subscribe(({open}) => {
+    if (open && !wasOpen) {
+        void reloadDeleteDialogData();
+    } else if (!open && wasOpen) {
+        instanceId += 1;
+    }
+    wasOpen = open;
+});
```
edloidas, line 354: Why do we need to increment instanceId here?
Also, you can get the previous store state from second argument of subscribe handler and check if the dialog was opened.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/deleteDialog.store.ts
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    return {removedMain, removedDependant};
+};
+
+//
+// * Completion handling
+//
+
+const handleCompletionEvent = (changeItems: ContentServerChangeItem[], kind: DeleteAction): void => {
```
edloidas, line 300: I think that notifying about successful deletion or archive must not depend on server events.
Basically, you press Archive -> send request -> show message on success or error based on data you used to send request.
This handler should probably be used to reload the dialog.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
edloidas: So you have single Root, single "open" and focus. But you switch content depending on the view.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+                                invalid: {count: 0, onExclude: () => {}},
+                                noPermissions: {count: 0, onExclude: () => {}},
+                                inbound: {
+                                    count: inboundCount,
+                                    onIgnore: () => ignoreDeleteInboundDependencies(),
+                                },
+                            }}
+                        />
+
+                        <Dialog.Body className="flex flex-col gap-y-10 py-5">
```
edloidas: Just like in publish dialog, body is incorrectly become scrolled, while having y-axis padding. This leads to the problem, when padding is also scrolled and space between the body and other elements become smaller. Add another scrollable wrapper to keep those paddings intact.

Image

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
edloidas: I'd recommend decoupling and moving Content into separate files. Confirmations must not completely depend on the global state of the dialog or what state was set on button click. We can have two separate confirmations for Delete and Publish that will just receive the list of IDs and perform separate requests. And those requests must not depend on if the dialog is shown/loading/failed/empty or not, they are not those kinds of requests. You are already on confirmation step, all it takes is to confirm action

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/items/ContentItem.tsx
```diff
@@ -43,8 +48,8 @@ export const ContentItem = ({
     return (
         <ListItem selected={selected}>
             <ListItem.Content>
-                <Button onClick={onClick} className="block flex-1 w-[calc(100%+10px)] h-8 -mx-1.25 -my-1 px-1.25 py-1">
-                    <ListItem.DefaultContent label={label} icon={Icon} />
+                <Button onClick={onClick} className={cn('block flex-1 w-full h-8 -mx-1.25 -my-1 px-1.25 py-1', mainItem && "h-13")}>
```
edloidas: The calculated width was actually used to visually enlarge the button, while keeping the content in place: we increase its size by 10px, then center the content with padding, and add negative margin to keep the content in place inside its container.
For larger content, the horizontal border should be larger. Try to apply those in the browser:
```css
margin-inline: -10px;
padding-inline: 10px;
width: calc(100% + 20px);
```

edloidas: Also, would be nice to have the link active area of the same height, as the button itself, matching the row height. Right now it's not consistent.

Image

edloidas:
We still need to update how we render the dialogs.
Currently, you show new dialog as a separate dialog. This is not quite correct. It makes overlay to re-render every time and complicates the data management:
- Return from `DeleteDialog` a single `<Dialog.Root>` node from top level.
- Add new local type called `type View = 'main' | 'confirmation' | 'progress'`
- Reset that `view` state onClose, no need to keep it in global store.
- Depending on the current view you should render either Delete content, or Confirmation content.
- You'll have to rework delete confirmation preset to return Content only, not the whole dialog. Footer of the confirmation content must also be adjusted, so that the buttons won't trigger close.

I updated the OnlineStatus. Keep it as is for now, we will refactor status elements in a separate branch.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/items/ContentItem.tsx
skjulestad: I will set it back to 100% + 10px as 20px results in overlap:
image

Height was set for sub items but is now adjusted based on if it is mainItem or not:

image
image

Also gave some more space for scrollbar:
image

PR #9514, opened 2025-12-03
Enonic UI: Filter Panel Dependencies Section #9510

### LINKED ISSUE #9510: Enonic UI: Filter Panel Dependencies Section
Filter panel must parse the URL and use parameters to set corresponding filter options.
Root item should also be set, by will be done in different task. Use a stub for that functionality now. 

### Problem

There, on click handlers, we trigger `ShowDependenciesEvent`, which opens a new tab with a URL that depends on the parameters passed to the event.

That URL is not only ignored by the filter panel (which makes the behavior of the dependency's widget incomplete) but also resets after the page is loaded (haven't investigated that yet).

---

### References

https://github.com/user-attachments/assets/bb783d35-5e21-415f-a90e-56568c96632b

reisfmb:
@ashklianko 

Moved the `BrowseFilter` to `./filter`, folder where I also placed the `BrowseDependencies`.

Not really sure I liked that... Feel free to update names / placements in case you think of something more suitable!


modules/lib/src/main/resources/assets/js/app/browse/filter/DependenciesSection.ts
ashklianko, line 1: Can remove this file totally, it is not used anywhere

modules/lib/src/main/resources/assets/js/v6/features/views/browse/layout/filter/BrowseDependencies.tsx
```diff
@@ -0,0 +1,91 @@
...
+    inbound,
+    onCancelClick = () => {},
+}: BrowseDependenciesProps): ReactElement | null => {
+    const inboundLabel = useI18n('panel.filter.dependencies.inbound');
+    const outboundLabel = useI18n('panel.filter.dependencies.outbound');
+    const label = inbound ? inboundLabel : outboundLabel;
+
+    if (!item) return null;
+
+    return (
```
ashklianko, line 29: Screenshot 2025-12-03 at 15 50 29

Text overflow handling doesn't work accordingly

ashklianko (changes requested):

ashklianko (approved):

edloidas (commented):
Two comments:
1. I'm not sure if it looks good. I'd like to wait for the final design. Delete button seems to be a little off. At the same time aligning it by the icon is not ideal too.

Image

2. We have the Content view (icon + display name + path) used a lot across the app (tree list, details widget, lots of comboboxes). I think it might be a good opportunity to create a shared/content component for that.

reisfmb:
@edloidas 

1. Since we do not have a design for it, I just coded a minimal version based on how it currently looks in CS. @alansemenov approved it internally via Slack. Can't we merge it for now and improve later once the design comes?  _(Will the design eventually come? Is in the scope of the designer to work on that?)_

2. Totally agree, but IMO better if done on a separate task, where we not only implement it, but also update relevant code to use it.

PR #9526, opened 2025-12-05
Enonic UI: make clicks outside WidgetsDropdown close it #9524

### LINKED ISSUE #9524: Enonic UI: make clicks outside WidgetsDropdown close it
Image

Clicking outside should close the expanded dropdown

ashklianko:


edloidas:
While this fixes the issue, I have 2 problems with this issue and PR:

1. We have a "click outside" hook in Enonic UI. No need to create a new one.
2. This component should have been a `Selector` (we don't have an implementation for it yet, but we need one in Enonic UI), which we could say is somewhat similar to a `Combobox` by its appearance, but by behavior closer to a `Menu` component with `RadioGroup`, that already supports click outside, keyboard navigation, and selection. Something like [this one](https://enonic.github.io/npm-enonic-ui/?path=/story/components-menu--close-on-select).

So, I can't merge it, as it needs adjustments (1), but making them is pointless, because the component itself is incorrect. Please, consider implementing the Selector pattern first and then applying it.

PR #9527, opened 2025-12-05
Enonic UI: Create shared ContentLabel #9519

### LINKED ISSUE #9519: Enonic UI: Create shared ContentLabel
Create a new `ContentLabel` component (yes, it's a good name, I was skeptical at first too) under the `shared/content/` folder (or a better one, if possible).

1. Start with creating `shared/ItemLabel` component.
  - It must accept `icon`, `primary`, `secondary` (optional).
  - Icon must adapt to the text content, which means if only the title is present, the icon must be smaller.
  - You can use grid layout to create proper position of the elements.
  - Icon should be any `ReactNode`, don't try to control it's size directly, try using container. End consumers will need to provide correct adjustable icons.
  - `primary` and `secondary` should also accent any  `ReactNode` but be placed inside `<span>` and `<small>` accordingly. Adjust text size and color. Don't forget about `group-data-[tone=inverse]:text-*` support to correctly display it on selected components.
  - The default height of the full Label is 41px in height, but I'd suggest going with 40px (22px primary, 18px secondary).
  - Compact version is 30px height. See Publish Dialog design for those items.
2. Create `ContentLabel` component. 
  - `content: ContentSummaryAndCompareStatus` — just to get the data from
  - `compact?: boolean` — hides secondary, renders `path` in `primary` instead of `displayName`.
  - `hideStatus?: boolean` — disable Content status icon in the top right corner of the main icon. 
  - Example: `<ContentLabel content={content} hideStatus compact>`
3. Create `ProjectLabel` (see content of `ProjectItem.tsx`)
  - `project: Project` — just to get the data from.
    - In `ProjectItem` we accept just the primitives, which is a bit better, but we'll adjust it later and make all those Labels accept smaller simpler objects in the future.
    - Also fix the language in the title (currently it has subtle color and is semibold, but must be main and normal).
  - Example: `<ProjectLabel project={project}>`.

Make sure the text is correctly truncated on small sizes.

**References:**

Image

Image

Image

ashklianko:


reisfmb:
Some notes about the [task description](https://github.com/enonic/app-contentstudio/issues/9519#issue-3694886940):

> The default height of the full Label is 41px in height, but I'd suggest going with 40px (22px primary, 18px secondary).

Height is 40px in case of `Itemlabel` usage with both primary and secondary, and 30px in case of usage without secondary.

> Icon must adapt to the text content, which means if only the title is present, the icon must be smaller.

I'm not sure my [approach](https://github.com/enonic/app-contentstudio/commit/51060dca86732c229bf81da3be516f995433b9c5#diff-73b262c1a8b7d182a7aca9ca328cce26f1faec750537172028d3017cee8682aeR19-R27) was the best... I probably lack css knowledge for this. Feel free to update it if needed.

> Create `ContentLabel` component.

Implemented with a prop not described on the task: `fullPath?: boolean`. Reason is because with it we can control weather we want to display the `content.getPath().toString()` or the `content.getPath().getName()`, the latter being the one used in the tree grid.

edloidas (approved):

PR #9534, opened 2025-12-08
Enonic UI: UnpublishDialog #9532

### LINKED ISSUE #9532: Enonic UI: UnpublishDialog
Create Unpublish dialog, that will look similar to the Delete dialog, with the difference that content items must not have "Show reference" links.

skjulestad:
Migratet UnpublishDialog, moved description back to header. Changed order of footer buttons to fit figma design.


modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -374,17 +375,11 @@ async function startApplication() {
...
-            contentUnpublishDialog = new ContentUnpublishDialog();
-        }
-
-        contentUnpublishDialog
-            .setContentToUnpublish(event.getModels())
-            .open();
+        openUnpublishDialog(event.getModels(), {
+            onYes: event.getYesCallback(),
+            onNo: event.getNoCallback(),
+        });
```
edloidas: We no longer use `ContentUnpublishPromptEvent` to control the dialog state.
Remove it, it's handlers, and replace it's usage in Actions with direct `openUnpublishDialog()` call

modules/lib/src/main/resources/assets/js/app/browse/ContentUnpublishPromptEvent.ts
```diff
@@ -5,10 +5,32 @@ import {ContentSummaryAndCompareStatus} from '../content/ContentSummaryAndCompar
 
 export class ContentUnpublishPromptEvent extends BaseContentModelEvent {
```
edloidas: Remove completelly

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialog.tsx
```diff
@@ -112,15 +112,22 @@ export const DeleteDialog = (): ReactElement => {
 
         return (
             <>
-                <Dialog.DefaultHeader title={confirmTitle} withClose />
+                <Dialog.DefaultHeader title={confirmTitle}>
+                    <p className="text-subtle">{confirmDescription}</p>
```
edloidas: Shouldn't it be just in `description` in DefaultHeader? Also, text is not subtle by design.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                    <Separator className="pr-1" label={dependantsLabel}/>
+                                    <ul className="flex flex-col gap-y-1.5">
+                                        {dependants.map(item => (
+                                            <ContentListItemWithReference
+                                                key={`dep-${item.getId()}`}
+                                                variant='compact'
+                                                content={item}
+                                                branch={Branch.DRAFT}
+                                                hasInbound={isInbound(item)}
+                                            />
```
edloidas: Saem here

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+                            resetView();
+                        }}
+                    />
+                </Dialog.Footer>
+            </>
+        );
+    };
+
+    return (
+        <>
```
edloidas: Why do you need to wrap it in Fragment? You return just a single component.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+                                errors={{
+                                    inProgress: {count: 0, onExclude: () => { }},
+                                    invalid: {count: 0, onExclude: () => { }},
+                                    noPermissions: {count: 0, onExclude: () => { }},
+                                    inbound: {
+                                        count: inboundCount,
+                                        onIgnore: () => ignoreUnpublishInboundDependencies(),
+                                    },
+                                }}
+                            />
```
edloidas: I see we use Status such way twice already. Maybe it's better to adjust SelectionStatusBar or create a new component, so that we don't need to initialize it with lots of "nullish" data.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+                            <Dialog.Body className="flex flex-col gap-y-10">
+                                <ul className="flex flex-col gap-y-2.5">
+                                    {items.map(item => (
+                                        <ContentListItemWithReference
```
edloidas: I specifically mentioned that Unpublish dialog shows regular content without "Show Reference" links.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+            </>
+        );
+    };
+
+    return (
+        <>
+            <Dialog.Root open={open} onOpenChange={handleOpenChange}>
+                <Dialog.Portal>
+                    <Dialog.Overlay/>
+                    {view === 'confirmation' ? (
```
edloidas: You will have progress step in the future. Using ternary here is not a good idea, as you'll have to check for more than two views. Just check for it directly and render the corresponding body `view === 'confirmation' && <`

edloidas: And it's better to place confirmation after main dialog.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+        });
+
+        await request.sendAndParse();
+        return true;
+    } catch (error) {
+        showError(error?.message ?? String(error));
+        $unpublishDialogPending.set(initialPendingState);
+        $unpublishDialog.set({
+            ...viewState,
+            open: true,
```
edloidas: Don't open dialog on fail. We don't have multiple roots, we have multiple content components. It's either already open, or closed by the user.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+type UnpublishDialogStore = {
+    open: boolean;
+    loading: boolean;
+    failed: boolean;
+    items: ContentSummaryAndCompareStatus[];
+    dependants: ContentSummaryAndCompareStatus[];
+    inboundTargets: ContentId[];
+    inboundIgnored: boolean;
+    yesCallback?: () => void;
+    noCallback?: () => void;
```
edloidas: Remove and recheck the logic in the store.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+import {showSuccess} from '@enonic/lib-admin-ui/notify/MessageBus';
+import {ContentSummaryAndCompareStatus} from '../../../../app/content/ContentSummaryAndCompareStatus';
+import {ContentId} from '../../../../app/content/ContentId';
+import {ContentSummaryAndCompareStatusFetcher} from '../../../../app/resource/ContentSummaryAndCompareStatusFetcher';
+import {ResolveUnpublishRequest} from '../../../../app/resource/ResolveUnpublishRequest';
+import {UnpublishContentRequest} from '../../../../app/resource/UnpublishContentRequest';
+import {$contentUnpublished} from '../socket.store';
+import {createDebounce} from '../../utils/timing/createDebounce';
+
+export type UnpublishAction = 'unpublish';
```
edloidas: You have one action only. Remove it. We can simplify it and make it closer to publish.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+// * Derived state
+//
+
+export const $unpublishTotalItems = computed($unpublishDialog, ({items, dependants}) => items.length + dependants.length);
+
+export const $unpublishHasSite = computed($unpublishDialog, ({items, dependants}) => {
+    return [...items, ...dependants].some(item => item.getContentSummary().isSite());
+});
+
+export const $unpublishHasBlockingInbound = computed($unpublishDialog, ({inboundTargets, inboundIgnored}) => {
```
edloidas: Must have different names following best practices in naming and start with `$is*`. Check the Delete dialog store.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+    return inboundTargets.length > 0 && !inboundIgnored;
+});
+
+export const $unpublishInboundIds = computed($unpublishDialog, ({inboundTargets}) => inboundTargets.map(id => id.toString()));
+
+export const $unpublishDialogReady = computed([$unpublishDialog, $unpublishHasBlockingInbound], (state, hasInbound) => {
```
edloidas: This name too

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+});
+
+// ! Guards against stale async results (increment on each dialog lifecycle)
+let instanceId = 0;
+
+//
+// * Helpers
+//
+
+const sortDependants = (dependants: ContentSummaryAndCompareStatus[], inboundTargets: ContentId[]): ContentSummaryAndCompareStatus[] => {
```
edloidas: Looks similar to Delete dialog. Probably can be extracted to utils.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+//
+// * Public API
+//
+
+export const openUnpublishDialog = (items: ContentSummaryAndCompareStatus[], callbacks?: {onYes?: () => void; onNo?: () => void;}): void => {
+    if (items.length === 0) {
+        return;
+    }
+
+    instanceId += 1;
```
edloidas: Instance must be incremented not here, but when the new request starts.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+
+    $unpublishDialog.set({
+        ...initialViewState,
+        open: true,
+        loading: true,
```
edloidas: Loading must be handled via subscription `$unpublishDialog.subscribe(({open}, oldState) => ...);`
See delete and unpublish dialogs.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+
+export const openUnpublishDialog = (items: ContentSummaryAndCompareStatus[], callbacks?: {onYes?: () => void; onNo?: () => void;}): void => {
+    if (items.length === 0) {
+        return;
+    }
+
+    instanceId += 1;
+
+    $unpublishDialog.set({
+        ...initialViewState,
```
edloidas: See how it's done in Delete dialog store. You should use `structuredClonde` here, since `initialViewState` has mutable properties and applying it with spread operator you still keep the original arrays from the initial state, that can be accidentally mutated.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+//
+
+$unpublishDialog.subscribe(({open}, prev) => {
+    const wasOpen = prev?.open ?? false;
+    if (open && !wasOpen) {
+        void reloadUnpublishDialogData();
+    } else if (!open && wasOpen) {
+        instanceId += 1;
+    }
+});
```
edloidas: Update it to work the same as in other stores.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+
+$contentUnpublished.subscribe((event) => {
```
edloidas, line 414: Still missing `$contentUpdated`, `$contentCreated`, `$contentArchived`, `$contentDeleted` handlers to update and reload the lists on server content changes.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+});
+
+export const $unpublishHasBlockingInbound = computed($unpublishDialog, ({inboundTargets, inboundIgnored}) => {
+    return inboundTargets.length > 0 && !inboundIgnored;
+});
+
+export const $unpublishInboundIds = computed($unpublishDialog, ({inboundTargets}) => inboundTargets.map(id => id.toString()));
+
+export const $unpublishDialogReady = computed([$unpublishDialog, $unpublishHasBlockingInbound], (state, hasInbound) => {
+    const {submitting} = $unpublishDialogPending.get();
```
edloidas: Non-reactive. Computed stores must not get data from other stores via `get()`. Add store to the stores array.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+type UnpublishDialogPendingStore = {
+    submitting: boolean;
+    pendingAction?: UnpublishAction;
+    pendingIds: string[];
+    pendingTotal: number;
+    pendingPrimaryName?: string;
+};
+
+// Initial state snapshot for reset
+const initialViewState: UnpublishDialogStore = {
```
edloidas: Probably misleading name, as it's not view state. Just `initialState` or even `INITIAL_STATE` would be good.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/unpublishDialog.store.ts
```diff
@@ -0,0 +1,385 @@
...
+    $unpublishDialog.set(initialViewState);
+    $unpublishDialogPending.set(initialPendingState);
+};
+
+export const ignoreUnpublishInboundDependencies = (): void => {
+    $unpublishDialog.setKey('inboundIgnored', true);
+};
+
+export const executeUnpublishDialogAction = async (): Promise<boolean> => {
+    const viewState = $unpublishDialog.get();
```
edloidas: Just get the data from store directly via destructing `{loading, failed, ...}`. No need to restore the state of the dialog on error. It'll either stay unchanged, or user will select another data and open unpublish dialog again just to see that after failed publishing the old content will be restored.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
skjulestad: This got refactored to reflect DeleteDialog.tsx, I do not understand what change is wanted here.

skjulestad: Patched in next commit

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/unpublish/UnpublishDialog.tsx
skjulestad: This will be refactored in next commit as well, but after removing showReferences I am starting to wonder if we even need SelectionStatusBar for UnpublishDialog, at least we have it and can remove it if we want then

skjulestad:
There is now an open pr for implementing progress bar based on this branch: https://github.com/enonic/app-contentstudio/pull/9546

edloidas (approved):
Good work. There are still some problems that need to be addressed, but they are common to several dialogs and rather big, so will be done in a separate task.

PR #9554, opened 2025-12-11
Enonic UI: Settings page toolbar #9553

### LINKED ISSUE #9553: Enonic UI: Settings page toolbar
Migrate settings page menu bar.

Current:
Image

Reference ([browse view toolbar](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/BrowseToolbar.tsx)):
Image

reisfmb:


reisfmb:
@edloidas 

Followed approach taken with `BrowseToolbar`: we prepend the new toolbar and hide the legacy one.

Its hacky, but does the job for now.

---

From my understanding the reason we can't easily migrate it is due to the fact that `SettingsBrowsePanel`( as well as `ContentBrowsePanel`) extends `BrowsePanel`, which expects an implementation of a method named `createToolbar`, which is finally handled internally via _lib-admin-ui_, leaving no control for us directly in cs.

---

## Tech debt

We need to be aware of that, and when properly migrating `BrowsePanel`, get rid of those legacy toolbars. Worth creating an issue to remember this?!

edloidas:
Looks great. We'll remove the old code later.

PR #9585, opened 2025-12-19
Enonic UI: DuplicateDialog #9568

### LINKED ISSUE #9568: Enonic UI: DuplicateDialog
Migrate ContentDuplicateDialog to new UI.

## Rationale

- Migrate the duplicate dialog to use `@enonic/ui` components and Preact patterns
- Align visual design with the updated DeleteDialog for consistency
- Simplify the dialog by removing unnecessary complexity (no confirmation step needed for duplicate)

---

## References

- Related issue: #9491
- Pattern example: `DeleteDialog.tsx` — shared dialogs implementation
- Old implementation: `ContentDuplicateDialog.ts`

---

## Things to Consider

- Unlike DeleteDialog, DuplicateDialog does not require a confirmation step — just main view and progress
- No "Show References" links needed in the dependant items list
- No status indicators (Online/Offline) for items
- Only one action button (Duplicate) — no secondary actions like Archive/Delete

---

## Implementation

> [!IMPORTANT]
> This is not a step-by-step guide — it's a functional checklist ordered logically.

1. Create `DuplicateDialog.tsx` component structure following the DeleteDialog pattern
2. Implement the store (`duplicateDialog.store.ts`) for dialog state management
3. Create `DuplicateDialogMainContent.tsx` with:
   - Main item display (title, description, icon)
   - Dependant items list section ("OTHER ITEMS THAT WILL BE DUPLICATED")
   - Duplicate action button with item count
4. Integrate with existing duplicate action flow and task polling

**UI Behavior:**

- Dialog opens with selected content items displayed
- Dependant items (children) are loaded and shown in expandable list
- Duplicate button shows total count of items to be duplicated
- Progress view appears during duplication operation
- Dialog closes on completion or can be cancelled

---

## Acceptance Criteria

- [ ] DuplicateDialog renders using `@enonic/ui` Dialog component
- [ ] Main view shows selected items with icon, title, and description
- [ ] Dependant items section displays children when applicable
- [ ] Duplicate button shows correct item count
- [ ] Progress is displayed during the duplicate operation
- [ ] Dialog integrates with existing duplicate task handling
- [ ] Cancel closes the dialog properly

---

## Out of Scope

- Confirmation step (not needed for duplicate unlike delete)
- "Show References" functionality for dependant items
- Status indicators (Online/Offline) for items
- Multiple action buttons (Archive, Delete)

---

_Old dialog example_

Image

---

_Drafted with AI assistance_

skjulestad:
Implement DuplicateDialog, with store, content and progressbar.


modules/lib/src/main/resources/assets/js/app/browse/action/DuplicateContentAction.ts
```diff
@@ -16,12 +15,9 @@ export class DuplicateContentAction
...
     protected handleExecuted() {
-        const contents: ContentSummaryAndCompareStatus[] = getSelectedItems();
-
-        new ContentDuplicatePromptEvent(contents)
-            .setYesCallback(() => {
-                resetSelection();
-            }).fire();
+        const contents = getSelectedItems();
+        openDuplicateDialog(contents);
+        resetSelection();
```
edloidas: This behavior is incorrect. Previously you would reset selection on "YesCallback", which means that selection will be reset after duplication, but in your case you reset it right after the dialog is opened. This must part of the dialog, when you click "Duplicate" button inside the dialog.

modules/lib/src/main/resources/assets/js/app/wizard/action/DuplicateContentAction.ts
```diff
@@ -12,7 +12,8 @@ export class DuplicateContentAction
         super(i18n('action.duplicate'));
         this.onExecuted(() => {
             const content = ContentSummaryAndCompareStatus.fromContentSummary(wizardPanel.getPersistedItem());
-            new ContentDuplicatePromptEvent([content]).setOpenActionAfterDuplicate(true).fire();
+            setDuplicateOpenTabAfterDuplicate(true);
```
edloidas: Previously, `setOpenActionAfterDuplicate` was set in the event to pass it down to the consumer, so that it'll call corresponding handlers in the dialog later. We don't need those.

Content is opened in new tab only if we duplicate it from the wizard.
I suggest the following: 
- Go to `app.store.ts`
- Add `page: 'browse' | 'wizard'` property and set it during initialization.
  - For now you can just use `ContentAppHelper.isContentWizardUrl()` but it would be better if you get rid of this legacy and use our own utils for that.
- Go to `App.tsx` and replace `isWizard` with computed `cosnt isWizard = useStore($isWizard);`.
- Then in the `duplicate.store.ts` you will use `$isWizard.get()` to check if we need to open content in new tab after it is duplicated and reported back.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/delete/DeleteDialogProgressContent.tsx
edloidas, line 31: You've forgotten to pass name as `data-component` prop.

edloidas: It's probably the same for other *Content components too.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/duplicateDialog.store.ts
```diff
@@ -0,0 +1,431 @@
...
+    duplicateCompletionHandled = true;
+
+    const state = $duplicateDialog.get();
+    const total = state.pendingTotal || state.items.length || state.pendingIds.length;
+    const primaryName = state.pendingPrimaryName
+        || state.items[0]?.getDisplayName()
+        || state.items[0]?.getPath()?.toString()
+        || '';
+    const itemDuplicated = i18n('dialog.duplicate.success.multiple', total);
+    const itemsDuplicated = i18n('dialog.duplicate.success.single', primaryName);
```
edloidas: Use names like `singleMessage` / `multipleMessage`. `itemsDuplicated` is just confusing when used for message names.

modules/lib/src/main/resources/assets/js/v6/features/shared/items/ContentListItem.tsx
```diff
@@ -9,6 +9,7 @@ import {StatusBadge} from '../status/StatusBadge';
 export type ContentItemProps = {
     content: ContentSummaryAndCompareStatus;
     variant?: ContentLabelVariant;
+    status?: boolean;
```
edloidas: `status` is a rather incorrect name for a boolean variable. We should go with something like `withStatus` or `showStatus`. 
But for consistency, I'd like you to keep the status here too. Don't omit it, so remove this property completely.

modules/lib/src/main/resources/assets/js/v6/features/shared/items/ContentListItemWithChildren.tsx
```diff
@@ -9,6 +9,9 @@ export type ContentListItemWithChildrenProps = {
     defaultIncludeChildren?: boolean;
     onIncludeChildrenChange?: (checked: boolean) => void;
     showIncludeChildren?: boolean;
+    status?: boolean;
+    selectionReadOnly?: boolean;
+    includeChildrenReadOnly?: boolean;
```
edloidas: Nah, this adds complexity. Just create component for that:
- Rename current `ContentListItemWithChildren` → `ContentListItemSelectableWithChildren`
- Create new `ContentListItemWithChildren` that wraps `ContentListItem`.
- Keep the status there, as I mentioned above

It'll look good, and we have design for this too:

Image

edloidas (changes requested):

edloidas (approved):

PR #9586, opened 2025-12-23
Enonic UI: NewContentDialog #9570

### LINKED ISSUE #9570: Enonic UI: NewContentDialog

reisfmb:


modules/lib/src/main/resources/assets/js/app/browse/action/ShowNewContentDialogAction.ts
```diff
@@ -16,7 +16,8 @@ export class ShowNewContentDialogAction extends ContentTreeGridAction {
 
     protected handleExecuted() {
         const contents: ContentSummaryAndCompareStatus[] = getSelectedItems();
-        new ShowNewContentDialogEvent(contents.length > 0 ? contents[0] : null).fire();
+        //new ShowNewContentDialogEvent(contents.length > 0 ? contents[0] : null).fire();
+        openNewContentDialog(contents?.[0]);
```
edloidas: This is not a big deal at all, but `contents` is probably defined, so no need for additional `?.` checks. This just adds noise and will give us linting error, once we migrate to strict TS checks without `any`. So it's more of a future wish.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+    const titleLabel = useI18n('dialog.new.title');
+    const allTabLabel = useI18n('dialog.new.tab.all');
+    const suggestedTabLabel = useI18n('dialog.new.tab.suggested');
+    const mediaTabLabel = useI18n('dialog.new.tab.media');
+    const hintLabel = useI18n('dialog.new.hint.filter');
+
+    //
+    // * Labels
+    //
```
edloidas: Again, more of a stylistic recommendation: use such comments only on root level, not inside the component. Too much visual noise for insignificant differences of code blocks.

modules/lib/src/main/resources/assets/js/v6/features/shared/ProgressBar.tsx
```diff
@@ -0,0 +1,14 @@
...
+        <div className={cn('w-full h-2.5 bg-surface-selected rounded-full overflow-hidden', className)}>
+            <div
+                className="h-full bg-success transition-all duration-500 ease-in-out"
+                style={{width: `${progress}%`}}
+            />
+        </div>
+    );
+};
+
+ProgressBar.displayName = 'ProgressBar';
```
edloidas: We already have `ProgressBar` under shared/primitives. Use that one

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
+import {ContentSummaryAndCompareStatus} from 'src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus';
```
edloidas: This file must be part of the store. It's not a component, it's just a bunch of utility functions.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/newContentDialog.store.ts
```diff
@@ -0,0 +1,155 @@
...
+    suggestedContentTypes: [],
+    filteredBaseContentTypes: [],
+    filteredSuggestedContentTypes: [],
+};
+
+export const $newContentDialog = map<NewContentDialogStore>(structuredClone(initialState));
+
+//
+// * Derived State
+//
```
edloidas: Remove commented code and black sections like this

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/newContentDialog.store.ts
```diff
@@ -0,0 +1,155 @@
...
+
+//
+// * Listeners
+//
+listenKeys($newContentDialog, ['open', 'parentContent'], ({open, parentContent}) => {
+    const activeProject = $activeProject.get() as Project; // TODO: fix type
```
edloidas: Fix type :)

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    tasks.forEach((task) => {
+        task.match(handlers.onEachSuccess, handlers.onEachError);
+    });
+}
+
+//
+// * Internal
+//
+const handlers = {
```
edloidas: Looks like those functions can be outside the `handlers` object, just as standalone utility handlers.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+    onEachSuccess: (success: UploadMediaSuccess) => {
+        removeContentTreeUploadItem(success.mediaIdentifier);
+
+        reload();
+
+        NotifyManager.get().showSuccess(`${success.mediaIdentifier} uploaded successfully!`);
```
edloidas: Use `showSuccess` and other notifiers directly, it looks cleaner. We have it exported from `'@enonic/lib-admin-ui/notify/MessageBus'`

edloidas: Messages must be localized, use i18n() call for now.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+            id,
+            file,
+            parentContent,
+            onProgress: (id, progress) => updateContentTreeUploadItemProgress(id, progress),
+        });
+    });
+
+    tasks.forEach((task) => {
+        task.match(handlers.onEachSuccess, handlers.onEachError);
+    });
```
edloidas: Either explicitly ignore via `void`, or better do:
```ts
await Promise.all(tasks.map(task => task.match(handlers.onEachSuccess, handlers.onEachError)));
```
Then you'll have to mark the function async and add proper return type. You'll ignore it explicitly in the dialog where it is called: `void uploadMediaFiles();`

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+            name,
+            imageSource: src,
+            parentContent,
+            onProgress: (id, progress) => updateContentTreeUploadItemProgress(id, progress),
+        });
+    });
+
+    tasks.forEach((task) => {
+        task.match(handlers.onEachSuccess, handlers.onEachError);
+    });
```
edloidas: Same `await` changes here

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export function uploadDragImages({dataTransfer, parentContent}: UploadOptions) {
+    const htmlData = dataTransfer.getData('text/html');
+    const imgSources = extractImageSources(htmlData);
+
+    if (imgSources.length === 0) return;
+
+    const tasks = imgSources.map((src) => {
+        if (!src.startsWith('data:')) return uploadRemoteImage({imageSource: src, parentContent});
```
edloidas: 1. If it's not just simple `return;` or `return value;`, avoid placing it on one line, better wrap in `{}`.

2. But the problem here is that such does not create an item with progress bar in the gird.

3. Also, it usually throws an error like this, but it might not be fixable on frontend: 
```
image-20251229185635.jpg failed to upload: Content at path [/image-20251229185635.jpg] in repository [com.enonic.cms.features] in branch [draft] already exists
```

4. Upload does not support dragging images from the page that were defined via `data:` and throws:
```
data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7 failed to upload: Failed to load image from URL
```
It is intentional for security reasons, but we need to show a readable message here that we don't support it.

modules/lib/src/main/resources/assets/js/v6/features/utils/image/generateUniqueName.ts
```diff
@@ -0,0 +1,18 @@
...
+    const dateParts = [
+        date.getFullYear(),
+        date.getMonth() + 1,
+        date.getDate(),
+        date.getHours(),
+        date.getMinutes(),
+        date.getSeconds(),
+    ];
+
+    const padNumber = (n: number): string => String(n).padStart(2, '0');
```
edloidas: Move outside

modules/lib/src/main/resources/assets/js/v6/features/utils/image/generateUniqueName.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    return `image-${dateParts.map(padNumber).join('')}.${type}`;
```
edloidas: Add random to avoid possible collisions if the images will be processed at the same time.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/newContentDialog.store.ts
```diff
@@ -0,0 +1,155 @@
...
+    return contentTypes
+        .filter((contentType) => !contentType.getContentTypeName().isDescendantOfMedia())
+        .sort((a, b) => a.getDisplayName().localeCompare(b.getDisplayName()));
+}
+
+function getSuggestedContentTypes(
+    contentTypes: ContentTypeSummary[],
+    aggregations: AggregateContentTypesResult
+): ContentTypeSummary[] {
+    const DEFAULT_MAX_ITEMS = 100; //2;
```
edloidas: 2?

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
```diff
@@ -0,0 +1,45 @@
+import {Tab} from '@enonic/ui';
+import {ReactElement} from 'react';
+import {$newContentDialog, closeNewContentDialog} from '../../../store/dialogs/newContentDialog.store';
+import {useStore} from '@nanostores/preact';
+import {Image} from 'lucide-react';
+import {TargetedEvent} from 'preact';
```
edloidas: Should be imported from 'react' (it's alias now)

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogSearch.tsx
```diff
@@ -0,0 +1,40 @@
+import {cn, SearchField} from '@enonic/ui';
+import {ReactElement} from 'react';
+import {$newContentDialog, setInputValue} from '../../../store/dialogs/newContentDialog.store';
+import {useStore} from '@nanostores/preact';
+import {MutableRef} from 'preact/hooks';
```
edloidas: Should be imported from 'react' (it's alias now)

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
```diff
@@ -0,0 +1,45 @@
...
+        });
+
+        closeNewContentDialog();
+    };
+
+    return (
+        <Tab.Content value={tabName} className="py-7.5 min-h-50 h-full">
+            <input id="file-upload" type="file" multiple onChange={handleChange} className="hidden" />
+            <label
+                for="file-upload"
```
edloidas: Must be `htmlFor` instead of `for` when in JSX.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {$newContentDialog, closeNewContentDialog} from '../../../store/dialogs/newContentDialog.store';
+import {useStore} from '@nanostores/preact';
+import {Image} from 'lucide-react';
+import {TargetedEvent} from 'preact';
+import {uploadMediaFiles} from './NewContentDialogUpload';
+import {useI18n} from '../../../hooks/useI18n';
+
+const NEW_CONTENT_DIALOG_MEDIA_TAB_NAME = 'NewContentDialogMediaTab';
+
+export const NewContentDialogMediaTab = ({tabName}: {tabName: string}): ReactElement => {
```
edloidas: Move to props type

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogContentTypesTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {ItemLabel} from '../../ItemLabel';
+import {ContentIcon} from '../../icons/ContentIcon';
+import {NewContentEvent} from '../../../../../app/create/NewContentEvent';
+import {ContentSummaryAndCompareStatus} from '../../../../../app/content/ContentSummaryAndCompareStatus';
+import {ContentTypeSummary} from '@enonic/lib-admin-ui/schema/content/ContentTypeSummary';
+import {useI18n} from '../../../hooks/useI18n';
+
+const NEW_CONTENT_DIALOG_CONTENT_TYPES_TAB_NAME = 'NewContentDialogContentTypesTab';
+
+export const NewContentDialogContentTypesTab = ({
```
edloidas, line 18: Move to props type

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialog.tsx
```diff
@@ -0,0 +1,163 @@
...
+    };
+
+    const handleInputChange = (newValue: string) => {
+        if (newValue.length > 0) return;
+
+        setIsInputVisible(false);
+
+        setTimeout(() => {
+            dialogContentRef.current?.focus();
+        }, 100);
```
edloidas: Looks like a hack. DO you really need setTimeout here? Maybe `requestAnimationFrame` is better? Or even direct reactive response on some event? A hook on content change?

modules/lib/src/main/resources/assets/js/v6/features/store/contentTreeUpload.store.ts
```diff
@@ -0,0 +1,24 @@
+import {map} from 'nanostores';
+import {ContentUploadData} from '../views/browse/grid/ContentUploadData';
+
+type ContentTreeUploadStore = Record<string, ContentUploadData>;
+
+export const $contentTreeUpload = map<ContentTreeUploadStore>({});
+
+export function addContentTreeUploadItem(id: string, name: string, parentId = ''): void {
```
edloidas: Marking `parentId` optional might be a better choice if you want to have "unset" option, that will fall back to root by default.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogUpload.ts
reisfmb: - 1 and 2: improved
- 3: What would you suggest? Currently it's showing the error that comes from the backend...
- 4: take a look at point 5 [here](https://github.com/enonic/app-contentstudio/pull/9586#issuecomment-3697812098)

reisfmb:
Some things to be aware / discuss...

## 1 - Scroll indicator
image

Some scroll indication would improve UX (something like [this](https://i0.wp.com/css-tricks.com/wp-content/uploads/2021/09/Screen-Shot-2021-09-27-at-2.25.14-PM.png?w=588&ssl=1)) in cases we an overflow due to many content types available... This is not the only place where this might be needed.

## 2 - Hide Hint
image

After the user starts typing, we've decided to hide the "Start typing to filter the list" hint.

## 3 - Dropzone
image

The user can drag and drop files anywhere on the dialog, and it'll automatically switch to the "media" tab and handle dropped files. I'll work on a proper dropzone visual feedback in a [new task](https://github.com/enonic/app-contentstudio/issues/9595).

## 4 - Notifications
image

Success / error notifications are handled individually for each upload.

## 5 - "data:" files are not allowed
image

Here we need to improve the error message. A suggestion would be to update the notification ui, adding ellipsis overflow!?

## 6 - Concurrency
image

I've updated the upload media code to handle upload concurrently rather than sequentially.

alansemenov:
@reisfmb please register a new issue for the bug you described in nr 5

Otherwise it's looking great!

reisfmb:
@alansemenov 

Registered [here](https://github.com/orgs/enonic/projects/74/views/1?pane=issue&itemId=147775846&issue=enonic%7Capp-contentstudio%7C9606)

edloidas (approved):

Issue #9561, opened 2025-12-14
Enonic UI: Versions comparison dialog


ashklianko:
https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=44381-84157&m=dev

<img width="1066" height="856" alt="Image" src="https://github.com/user-attachments/assets/ea70fb39-3298-49bf-b8bb-e5b4a14ff3c2" />

## Overview
Provide a read-only versions comparison dialog that lets editors view the diff between two selected content versions. The dialog opens from the versions widget after selecting two versions and shows a clear comparison with optional inclusion of unchanged fields.

## Goals
- Enable quick, read-only comparison of two selected versions.
- Keep UI consistent with new v6 dialogs and versions list styling.
- Support both browse and wizard views.

## Non-goals
- Editing or changing selection within the dialog.
- Revert actions inside the dialog.
- Version aliasing or dropdown-based selection.

## Users
- Content editors comparing changes over time.
- Reviewers validating what changed before publish/unpublish decisions.

## UX Summary
- Dialog header: “Compare versions” with content path subtitle.
- Two selected version cards labeled “Older” and “Newer.”
- Diff panel showing jsondiffpatch output.
- Checkbox: “Show the entire content” (toggles unchanged fields).
- Read-only; no selection changes inside the dialog.

## User Flow
1. User selects two versions in the versions widget.
2. “Show changes” action opens the dialog.
3. Dialog renders older/newer cards and diff.
4. User optionally toggles “Show the entire content.”
5. User closes the dialog.

## Functional Requirements
- Open dialog only when exactly two versions are selected.
- Determine “Older” vs “Newer” by timestamp ordering.
- Render the two selected versions in a compact card layout matching versions list visuals.
- Fetch content JSON for both versions and render diff using jsondiffpatch HTML formatter.
- Cache fetched version JSON during a dialog session to avoid re-fetching.
- Provide a “Show the entire content” toggle that controls unchanged fields visibility.
- Gracefully handle loading and error states.
- Make dialog available in both browse and wizard shells.

## Data & State
- Required: content ID, content path, two version IDs, show-all toggle, open/close state.
- Sources: versions list selection + context content.
- Diff data: fetched via existing content version API.

## Error & Empty States
- Loading indicator while diff is computed.
- Error message if version fetch fails.
- “Versions are identical” message when no diff exists.

## Accessibility
- Keyboard focus should land inside dialog content.
[...]

PR #9618, opened 2026-01-08
Enonic UI: Versions comparison dialog #9561

### LINKED ISSUE #9626: Version History: misc issues
- updateMetadata status is shown with 'Unknown' text and Edit icon
- remove offline icon and related logic
- publish version should be comparable (but not revertable!)
- unknown items - check if all cases are handled
- expired and scheduled - show when
- unpublish version - show the one one who unpublished
- Make a patch version

ashklianko:


modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/SelectedVersionCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {getOperationLabel} from '../../../store/context/versionStore';
+
+type SelectedVersionCardProps = {
+    label: string;
+    version: ContentVersion;
+};
+
+export const SelectedVersionCard = ({label, version}: SelectedVersionCardProps): ReactElement => {
+    const modifierDisplayName = version.getModifierDisplayName() || version.getPublishInfo()?.getPublisherDisplayName();
+    const modifierLabel = modifierDisplayName ? useI18n('field.version.by', modifierDisplayName) : null;
```
edloidas: Must not be called conditionally.

modules/lib/src/main/resources/assets/js/v6/features/shared/status/VersionItemPublishStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const VersionItemPublishStatus = ({version, className}: VersionItemPublishStatusProps): ReactElement | null => {
+    if (!version) {
+        return null;
+    }
+
+    const publishStatus = getVersionPublishStatus(version);
+    const onlineLabel = useI18n('status.online');
+    const expiredLabel = useI18n('status.expired');
+    const scheduledLabel = useI18n('status.scheduled');
```
edloidas: Hooks after early return (also conditional calculation). They must always be on top, as it will break how they handled and how component is cached in React.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const [error, setError] = useState<Error | null>(null);
+
+    const leftVersion = useMemo(
+        () => (leftVersionId ? ContentVersionHelper.getVersionById(versions, leftVersionId) : null),
+        [leftVersionId, versions]
+    );
+    const rightVersion = useMemo(
+        () => (rightVersionId ? ContentVersionHelper.getVersionById(versions, rightVersionId) : null),
+        [rightVersionId, versions]
+    );
```
edloidas, line 60: Ideally, we move from custom classes to simple composable and testable utility functions. ContentVersionHelper too must be split as it mixes utilities, fetchers, relies on old Q lib (but it's not part of the current changes).
Here we can have use common function that will receive shape/interface with getId()/id fields and find it in array of such object. It can later be used in lot's of other places too.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+    }, [open, showAllContent]);
+
+    useEffect(() => {
+        if (!open) {
+            return;
+        }
+
+        if (!leftVersionId || !rightVersionId || !leftVersion || !rightVersion) {
+            closeCompareVersionsDialog();
+        }
```
edloidas: Potential race condition, as versions might not be loaded yet. Can we check if they are initialized first?

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+            cacheRef.current.clear();
+            setDiffHtml('');
+            setError(null);
+            setIsEmpty(false);
+            setIsLoading(false);
+            showUnchanged(false, null, 0);
+            return;
+        }
+
+        showUnchanged(showAllContent, null, 0);
```
edloidas: Potentially problematic, as it's called later with the same arguments for a wider list of dependencies in different effect.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+            cancelled = true;
+        };
+    }, [open, contentId, olderVersionId, newerVersionId, diffPatcher, versionsIdenticalLabel]);
+
+    useEffect(() => {
+        if (!open) {
+            return;
+        }
+
+        showUnchanged(showAllContent, null, 0);
```
edloidas: Second time we call it in an effect in similar way but now on `diffHtml` change too. This is probably the correct one.

modules/lib/src/main/resources/assets/js/v6/features/shared/status/VersionItemPublishStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+    version: ContentVersion | null;
+    className?: string;
+};
+
+export const VersionItemPublishStatus = ({version, className}: VersionItemPublishStatusProps): ReactElement | null => {
+    if (!version) {
+        return null;
+    }
+
+    const publishStatus = getVersionPublishStatus(version);
```
edloidas, line 24: Non-reactive store read. This must not be done inside the components (unless it's a handler for one-time data retrieval for some function call). Must use computed atom instead.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+    }, [open, showAllContent, diffHtml]);
+
+    if (open && !orderedVersions) {
+        return (
+            <Dialog.Root open={open} onOpenChange={handleOpenChange}>
+                <Dialog.Portal>
+                    <Dialog.Overlay />
+                </Dialog.Portal>
+            </Dialog.Root>
```
edloidas: Is it intended? To just show some kind of dark overlay on the page without the dialog?

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+        if (!open) {
+            return;
+        }
+
+        if (!leftVersionId || !rightVersionId || !leftVersion || !rightVersion) {
+            closeCompareVersionsDialog();
+        }
+    }, [open, leftVersionId, rightVersionId, leftVersion, rightVersion]);
+
+    useEffect(() => {
```
edloidas, line 109: Data loading in component. It's a bad practice and, again, violates the single responsibility principle. We have api/ and stores with their utilities for this.
I would recommend move it to `useVersionsDiff` to other version hooks and also make use of `neverthrow` there, as we do in some of the new code.

edloidas: We have `neverthrow` used in `uploadMedia.ts` btw.

Suggest moving `fetchContentVersion(contentId, versionId)` to `api/versions.ts`

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+                const leftJson = await fetchVersion(olderVersionId);
+                const rightJson = olderVersionId === newerVersionId
+                    ? leftJson
+                    : await fetchVersion(newerVersionId);
+
+                if (cancelled) {
+                    return;
+                }
+
+                const delta = diffPatcher.diff(leftJson, rightJson);
```
edloidas: After moving requests outside, I'd move `diffResult` to memo and depend on result of versions hook, that updates the versions. Something like:
```ts
const [leftJson, rightJson] = useVersionsJson(contentId, olderVersionId, newerVersionId);

const diffResult = useMemo(() => {
      if (!leftJson || !rightJson) return null;
      // ...
  }, [leftJson, rightJson, diffPatcher]);
```

edloidas: Now you can memorize heavy operation and split async and sync code.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+    useEffect(() => {
+        if (!open || !contentId || !olderVersionId || !newerVersionId) {
+            return;
+        }
+
+        let cancelled = false;
+
+        const fetchVersion = async (versionId: string): Promise<ContentJson> => {
+            const cached = cacheRef.current.get(versionId);
```
edloidas: Is `versionId` a UUID or another type of unique ID?

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const loadingLabel = useI18n('widget.versions.loading');
+    const versionsIdenticalLabel = useI18n('dialog.compareVersions.versionsIdentical');
+
+    const diffPatcher = useMemo(() => new DiffPatcher(), []);
+    const cacheRef = useRef<Map<string, ContentJson>>(new Map());
+
+    const [diffHtml, setDiffHtml] = useState<string>('');
+    const [isLoading, setIsLoading] = useState(false);
+    const [isEmpty, setIsEmpty] = useState(false);
+    const [error, setError] = useState<Error | null>(null);
```
edloidas: Might not be a huge performance problem, but I see that we usually update multiple entries one after another, especially on reset. Maybe we can place those ones in map store instead? And, when needed, reset them in one function call. Each `set*` update might trigger it's own re-rendering (though react usually batches such changes). Still, I think we can optimize it.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+                            )}
+
+                            {!isLoading && error && (
+                                <div className="text-sm text-red-600">{error.message}</div>
+                            )}
+
+                            {!isLoading && !error && (
+                                <div
+                                    className={cn('jsondiffpatch-delta text-sm', isEmpty && 'empty')}
+                                    dangerouslySetInnerHTML={{__html: diffHtml}}
```
edloidas, line 162: Obviously dangerous, may need a comment to replace with a better React compatible and save diff alternative. Current top-tier lib is this one:
https://diffs.com/

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/compare-versions/CompareVersionsDialog.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+    const handleOpenChange = (nextOpen: boolean) => {
+        if (!nextOpen) {
+            closeCompareVersionsDialog();
+        }
+    };
+
+    const handleShowAllChange = (checked: boolean) => {
+        setCompareVersionsShowAll(checked);
+    };
```
edloidas: Let's useCallback for those ones.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/compareVersionsDialog.store.ts
```diff
@@ -0,0 +1,51 @@
...
+const initialState: CompareVersionsDialogState = {
+    open: false,
+    contentId: null,
+    contentPath: null,
+    leftVersionId: null,
+    rightVersionId: null,
+    showAllContent: false,
+};
+
+export const $compareVersionsDialog = map<CompareVersionsDialogState>(structuredClone(initialState));
```
edloidas, line 22: We store `contentId` as a class there, which we should avoid (we don't actually need that functionality there). And it become problematic using `structuredClone` on classes, they'll just become plain objects: 
```ts
class Foo {}
const foo = new Foo();
console.log( foo instanceof Foo ); // true
console.log( structuredClone( foo ) instanceof Foo ); // false
```

edloidas (changes requested):

edloidas:
Tremendous work you did there 👍

PR #9706, opened 2026-01-23
Enonic UI: New issue dialog #9613

### LINKED ISSUE #9613: Enonic UI: New issue dialog
Implement dialog for new issue to replace legacy:

Image

New:

Image

Functionality:

New issue needs to receive and set title, description and assignee.

Add items is depending on new component but button can be implemented visually.

Footer is to have cancel and create new issue confirm button.

### LINKED ISSUE #9704: Enonic UI: Assignee selector
Implement assignee selector for new issue and issue details dialog.

Image

Image

Needs to render available assignees based on publish access to project and items.
Keyboard friendly navigation and list added assignees with the option to remove on command.

### LINKED ISSUE #9705: Enonic UI: Items selector
Implement item selector for new issue and issue details dialog.

Image

Image

Needs to handle items with popup and display selected items with / without children.
Should not display add children if all children belonging to item are online / non conflicting.

skjulestad:
Implementation of new issue dialog along with assignee and item selectors.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -272,8 +272,8 @@ importers:
   .xp/dev/lib-contentstudio:
     dependencies:
       '@enonic/ui':
-        specifier: ~0.33.0
-        version: 0.33.0(@radix-ui/react-slot@1.2.3(@types/react@19.2.2)(react@19.2.0))(focus-trap-react@11.0.4(@types/react-dom@19.2.1(@types/react@19.2.2))(@types/react@19.2.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(lucide-react@0.545.0(react@19.2.0))(preact@10.27.2)(react-dom@19.2.0(react@19.2.0))(react-virtuoso@4.18.1(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(react@19.2.0)
+        specifier: ~0.35.1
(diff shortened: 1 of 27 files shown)


edloidas (changes requested):
Generally, everything looks good.

---

#### What I’d change (but we don’t have time for that now — just keep it in mind):

- When creating elements like `ContentSelector`, we should have **separate implementations** for:
  - `ContentCombobox`
  - `ContentItems` (the element with excludable items)

  These should then be composed inside `ContentSelector`.

  This approach would allow us to:
  - Control which items are rendered in `ContentItems`
  - Reuse components in other places
  - Avoid tight coupling between responsibilities

  With the current implementation, we’re bound by existing constraints. Any display change would likely introduce boolean flags, which tends to cause problems long-term.

- `ItemSelector` is not yet complete, but it is functional for now. We’ll update it later.

---

#### Required Changes

Please make the following updates:

1. **Move the label outside `AssigneeSelector`**  
   The label must live in the dialog content, just like with other selectors and inputs.

2. **Remove the `text-subtle` color**  
   By design, labels should be black.

3. **Keep the Assignee row height consistent (48px)**  
   It should match the content rows. We display similar content with icons and two lines — they must be identical in size and layout.

4. **Update Enonic UI to `0.35.1`**  
   This version includes fixes for Portal usage inside dialogs, so you won’t need hacks or a custom portal anymore.  
   Just use `Combobox.Portal` and you’re done.

   Reference:  
   https://enonic.github.io/npm-enonic-ui/?path=/story/components-combobox--dialog-with-portal  
   (This should be trivial now.)

5. **Decouple your selectors and reorganize folders**  
   Place everything into subfolders. For example:

   ```text
   selectors/
     assignee/
       AssigneeSelector
       AssigneeOptionsContent
       AssigneeOptionRow
       ...
    ```

PR #9714, opened 2026-01-24
Enonic UI: Match design for filter panel dependencies #9679

### LINKED ISSUE #9679: Enonic UI: Match design for filter panel dependencies
Section on top of the Filter panel when showing dependencies for a specific content. [Updated Figma](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=48322-92746&t=pcfJ52rI2ezYqDCw-4)

Image

reisfmb:
Preferrer `button` + `icon` instead of of `IconButton` in this case, because `IconButton` hover styling was not resulting in a good UI for this specific component.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/filter/ContentBrowseFilterPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/filter/ContentBrowseFilterPanel.ts
@@ -69,7 +69,7 @@ export class ContentBrowseFilterPanel<T extends ContentSummaryAndCompareStatus =
             exportOptions: this.getExportOptions(),
         });
 
-        this.appendChild(this.dependenciesSection);
+        super.appendChild(this.dependenciesSection);
         this.appendChild(this.filterComponent);
 
         this.handleEvents();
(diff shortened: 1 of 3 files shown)


edloidas (changes requested):
Please, adjust styles according to out discussion in Slack:

---

IMO, it must stay a button with default styles. No need to change it.
But to make it nicer, I'd suggest you do the following:
Use iconSize lg (same as 24, and same with content icon), not 30px
Set button size to 28 px (size-7) to make it less big.
Adjust the right padding to position the button correctly, so there'll be similar gaps between button and container on each side.
I'd also add rounded-sm to match the input and other elements. Sharp edges doesn't follow our design and looks better when there are active elements with round corners inside.

I meant that there is 10px padding on the left and right, but content icon has not additional padding on the left. But if we are talking about the button, the icon is smaller than the button itself, so to place icon correctly on the same distance from the right border, we must either add negative right margin or decrease the container padding on the right.

Add rounded-sm to container and add gap-2 between label and container to match the rest of the elements that we have. Without that gap it looks bad with letters pqy and inconsistent with other labels.
And a quick note: you don't need to combine cursor styles with :hover as it's mostly pointless and add problems when you override styles.

reisfmb:
> I meant that there is 10px padding on the left and right, but content icon has not additional padding on the left. But if we are talking about the button, the icon is smaller than the button itself, so to place icon correctly on the same distance from the right border, we must either add negative right margin or decrease the container padding on the right.

Had to adjust the left padding as well, otherwise, the right element would be too close to the right end of the container, so the hover effect of the close button would be "touching" the right end of the container.

Looks good to me. Nice suggestions!
image

PR #9715, opened 2026-01-25
Enonic UI: Changes to the Delete dialog #9655

### LINKED ISSUE #9655: Enonic UI: Changes to the Delete dialog
We want to rename the "Archive" action to "Delete" while keeping the default archiving behaviour as is. There will be no option to permanently delete content in Content Studio, only from CS+.

Here's what needs to be done:
* Replace "Archive" action with "Delete" (use the existing `action.delete` i18n resource) in the toolbar and the context menu
* Change title of the Archive/Delete dialog from "Archive item(s)" to "Delete item(s)" (use the existing `dialog.delete` i18n resource). Keep the subtitle as is.
* Remove "Delete (x)" link/button on the left of the main button in the dialog
* Replace caption on the main button with "Delete (x)" (use the i18n resource of the removed link/button)
* Title of the confirm dialog should always say "Confirm delete"
* Texts in the progress dialog should always say "Archiving...", as they do now when archiving
* Remove all the now redundant logic that shows either "Delete" or "Archive" depending on which button was clicked in the dialog

Check the Figma design [here](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=48160-103841&t=s9MPSJF1poR73AEB-4)

skjulestad:
Replace archive labels with delete, keep archive action and refactor code.


modules/lib/src/main/resources/i18n/dialogs.properties
```diff
@@ -25,10 +25,13 @@ dialog.archive.action=Archive
 dialog.archive.success.single=Item \"{0}\" has been archived.
 dialog.archive.success.multiple={0} items have been archived.
 dialog.archive.subname=Archive selected items and their children
+dialog.delete.subname=Delete selected items and their children
```
alansemenov: This one is not used and should be removed

modules/lib/src/main/resources/i18n/dialogs.properties
```diff
@@ -25,10 +25,13 @@ dialog.archive.action=Archive
...
 dialog.archive.dependants=Other items that will be archived
+dialog.delete.dependants=Other items that will be deleted
```
alansemenov: This one is not used and should be removed

edloidas (changes requested):
Incorrect texts in subname and dependencies (still mentioning Archive):

Image

Same for confirmation, progress dialog and notification.

alansemenov:
> Incorrect texts in subname and dependencies (still mentioning Archive):
> 

this is actually correct and according to the task:

> Change title of the Archive/Delete dialog from "Archive item(s)" to "Delete item(s)" (use the existing dialog.delete i18n resource). **Keep the subtitle as is.**

alansemenov (approved):

edloidas (approved):

PR #9721, opened 2026-01-26
Enonic UI: Match design for widget selector #9680

### LINKED ISSUE #9680: Enonic UI: Match design for widget selector
Widget selector. [Updated Figma](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=46788-185746&t=pcfJ52rI2ezYqDCw-4)

Image

Image

reisfmb:


modules/lib/src/main/resources/i18n/phrases.properties
```diff
@@ -183,7 +183,7 @@ field.preview.toolbar.versionHistory.label=Version history
 field.preview.failed=Failed to render content preview.
 field.preview.failed.description=Please check logs for errors
 field.preview.missing.description=Some applications required for rendering are not available
-field.contextPanel.selector.placeholder=Select a widget...
+field.contextPanel.selector.placeholder=Type to search...
```
edloidas: I don't really like this placeholder text. Since it's a selector, you can't type anything into the trigger, but it looks like it's a Combobox — a component with Input and Dropdown, but it is not.
I would prefer to keep it as is.

modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/WidgetsSelector.tsx
edloidas: This one is useless, remove.

modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/WidgetsSelector.tsx
```diff
@@ -34,12 +35,22 @@ const WidgetsSelector = ({widgetViews = [], externalSelectedWidgetView = undefin
...
+                            <span className="text-subtle">
+                                {selectorPlaceholder}
+                            </span>
+                        ) : (
+                            <WidgetsSelectorItem
+                                widgetView={widgetViews.find(
+                                    (widgetView) =>
+                                        getWidgetKeyForSelector(widgetView) ===
+                                        value,
+                                )}
```
edloidas: `useMemo` for `widgetView`, otherwise it'll do find on each rerender.

edloidas (changes requested):

modules/lib/src/main/resources/i18n/phrases.properties
reisfmb: This is based on a recent update in figma. Please validate with Alan the proper way to proceed, since your suggestion differs from what was defined.

alansemenov: "Type to search..." is correct, this is the standard placeholder which we use in inputs with typeahead (filter).

But we shouldn't add a new one, we already have one in lib-admin-ui's i18n resource bundle called "field.option.placeholder".

reisfmb: Thanks Alan.

But should we do this change or keep as Mikita suggested, that is, no placeholder with typing suggestion on the widget selector when it is expanded.

edloidas: Just to clarify: this is a selector (like a browser's `<select>`). Type-ahead (quick typing to jump to options) != filtering (typing into an input to filter the dropdown content).

While we do use comboboxes with a "Type to search..." placeholder, this isn't one of them, and I'm afraid that could be confusing. It would still work, sure — but there's no input.

reisfmb:
As discussed via meeting, we'll move to use combobox + view

edloidas (approved):

PR #9752, opened 2026-02-01
Enonic UI: New project dialog - Step 1 #9732

### LINKED ISSUE #9732: Enonic UI: New project dialog - Step 1

reisfmb:


modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+        return Math.min(contentHeight, MAX_HEIGHT);
+    }, [nodesCount]);
+
+    return (
+        <Dialog.StepContent step={STEP}>
+            <h3 className="mb-2 font-semibold">{label}</h3>
+
+            {/* Parent projects */}
+            <Combobox.Root value={searchValue} onChange={setSearchValue} contentType="tree" selectionMode={selectionMode}>
```
edloidas: This could be a separate `ProjectSelector` component (combobox + selected view).
Also, it looks like you forgot to test the single selection without the flag. There are two modes (must be supported at the component level):

- Multi-selection — similar to what we have now, but it also needs to be a [staged](https://enonic.github.io/npm-enonic-ui/?path=/story/components-combobox--staged) type.
- Single selection — we must not show checkboxes in the dropdown. Selection will be instant (not staged), but once you select something, the whole dropdown will be hidden, and you'll only see the selected option. Remove the additional side padding for the item in that view.

Items in the dropdown must be rendered with our new `GridList` component for better a11y.

Image

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/NewProjectDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+        if (open) return;
+
+        closeNewProjectDialog();
+    };
+
+    return (
+        <Dialog.Root defaultStep={'step1'} data-component={NEW_PROJECT_DIALOG_NAME} open={open} onOpenChange={handleOpenChange}>
+            <Dialog.Portal>
+                <Dialog.Overlay />
+                <Dialog.Content className="h-178 w-200 max-w-auto">
```
edloidas: Height should adjust to the dialogs content, unlike the `NewContentDialog`. See how it's done in other dialogs (publish, delete, etc)

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/newProjectDialog.store.ts
```diff
@@ -0,0 +1,56 @@
...
+type NewProjectDialogStore = {
+    config: {
+        open: boolean;
+        isMultiInheritance: boolean;
+        selectedProjects: Project[];
+    };
+    data: {
+        parentProjects?: Readonly<Project>[];
+    };
+};
```
edloidas, line 17: Let's use separate stores for config and data. There are problems with updating and correctly triggering updates for nested properties in `DeepMap` in `nanostores`. So try flat maps first.
Either keep all props on the same level, or split into several stores.

modules/lib/src/main/resources/assets/js/v6/features/store/dialogs/newProjectDialog.store.ts
```diff
@@ -0,0 +1,56 @@
...
+import {ProjectConfigContext} from '../../../../app/settings/data/project/ProjectConfigContext';
+
+//
+// * Store State
+//
+
+type NewProjectDialogStore = {
+    config: {
+        open: boolean;
+        isMultiInheritance: boolean;
```
edloidas: Move this one to `config.store.ts`. We should store app level configuration there.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
...
+                                    onSelectionChange={setSelection}
+                                    onExpand={handleExpand}
+                                    onCollapse={handleCollapse}
+                                    virtuosoRef={virtuosoRef}
+                                    aria-label={label}
+                                    className="h-full"
+                                >
+                                    {({items, getItemProps, containerProps}) =>
+                                        filteredItems.length > 0 ? (
+                                            <Virtuoso<FlatNode<Readonly<Project>>>
```
edloidas: As I mentioned, use `GridList` here. And probably no need to use Virtuoso here, as it might be overkill.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+export const NewProjectDialogStep1Content = (): ReactElement => {
+    // Hooks
+    const {
+        config: {selectedProjects, isMultiInheritance},
+    } = useStore($newProjectDialog);
+    const {projects} = useStore($projects);
+    const [searchValue, setSearchValue] = useState<string | undefined>();
+    const [selection, setSelection] = useState<ReadonlySet<string>>(new Set(selectedProjects.map((p) => p.getName())));
+    const [expanded, setExpanded] = useState<Set<string>>(new Set());
```
edloidas: I think that using Set here might be overkill. You don't need it for performance, but you map it to an array later to iterate through it. It's more secure, but we don't expect to have duplicates here anyway. And if there are duplicates, it must be fixed on the backend.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
...
+    const virtuosoRef = useRef<VirtuosoHandle>(null);
+
+    useEffect(() => {
+        const parentProjects = Array.from(selection).map((id) => projects.find((p) => p.getName() === id));
+        setNewProjectDialogParentProjects(parentProjects);
+    }, [selection]);
```
edloidas: `projects` must be in dependencies array too.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
+import {cn, Dialog, FlatNode, IconButton, VirtualizedTreeList} from '@enonic/ui';
+import {useStore} from '@nanostores/preact';
+import {forwardRef, ReactElement, useEffect, useMemo, useRef, useState} from 'react';
+import {$projects} from '../../../../store/projects.store';
+import {$newProjectDialog, setNewProjectDialogParentProjects} from '../../../../store/dialogs/newProjectDialog.store';
+import {Combobox} from '@enonic/ui';
+import {Virtuoso, VirtuosoHandle} from 'react-virtuoso';
+import {Project} from 'src/main/resources/assets/js/app/settings/data/project/Project';
+import {HTMLAttributes} from 'preact';
```
edloidas: Use `react` import. We should stick to `react` everywhere (it is aliased now, and we'll replace Preact with React in the future)

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
+import {cn, Dialog, FlatNode, IconButton, VirtualizedTreeList} from '@enonic/ui';
+import {useStore} from '@nanostores/preact';
+import {forwardRef, ReactElement, useEffect, useMemo, useRef, useState} from 'react';
+import {$projects} from '../../../../store/projects.store';
+import {$newProjectDialog, setNewProjectDialogParentProjects} from '../../../../store/dialogs/newProjectDialog.store';
+import {Combobox} from '@enonic/ui';
+import {Virtuoso, VirtuosoHandle} from 'react-virtuoso';
+import {Project} from 'src/main/resources/assets/js/app/settings/data/project/Project';
```
edloidas: Incorrect (yet working) import. Use relative one

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
```diff
@@ -0,0 +1,206 @@
...
+import {HTMLAttributes} from 'preact';
+import {ProjectLabel} from '../../../project/ProjectLabel';
+import {X} from 'lucide-react';
+import {useI18n} from '../../../../hooks/useI18n';
+import {projectsToTreeListItems} from '../../../../utils/url/projects';
+
+const STEP = 'step1';
```
edloidas: Let's keep Step number in comments only. number gives us almost nothing about the context. And it will be a problem if we switch the steps at some point. It's better to name step according to what it's doing. What about `NewProjectDialogParentStep` instead of `NewProjectDialogStep1`? Or something similar.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/NewProjectDialog.tsx
```diff
@@ -0,0 +1,43 @@
...
+                    <NewProjectDialogSteps.Step1.Header />
+
+                    <Dialog.Body className="p-1.5">
+                        <NewProjectDialogSteps.Step1.Content />
+                    </Dialog.Body>
+
+                    <Dialog.Footer className="flex flex-col">
+                        <Dialog.StepIndicator previousLabel={previousLabel} nextLabel={nextLabel} dots />
```
edloidas, line 33: We should hide the "previous" button for the first step. We probably need to update the component in Enonic UI, I believe.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-project/steps/NewProjectDialogStep1.tsx
edloidas: Regarding checkboxes in single mode — we'll fix it in Enonic UI, they'll be automatically hidden when selection is in single mode.

reisfmb:
@edloidas 

Implemented most of ur suggestions. Now, some discussion points.

---

**_1 - About the project selector tree grid:_**

 I kept it as it was, i.e, using `VirtualizedTreeList` + `Virtuoso`.

Tried to simplify using `TreeList`, but faced some problems with some states in _staged_ mode.

Also, its still not clear to me how to properly use `GridList` to create the tree view for the combobox items... All I had in mind looked hacky, so maybe I'm not getting how you envisioned it. Could you please clarify?

---

**_2 - Using `GridList` in another place_**

An update I did by my own was to use `GridList` to show the list of selected projects. Wdyt? 

---

3 - 

> Remove the additional side padding for the item in that view.

The design actually contains that side padding, so I kept it.

edloidas:
You can look at the usage of `GridList` [here](https://github.com/enonic/app-contentstudio/blob/epic-enonic-ui/modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/publish/PublishDialogMainContent.tsx#L171).
It would work very similarly to the `TreeList`, but instead of passing custom rows into it, we need to pass `GridList.Row` and place elements in `GridList.Cell` (preferably focusable ones into their own cells). We already have components for content specifically.

As for the paddings: I meant we don't need them when there is only one item selected. Maybe? I just see inconsistency in the design, since we show items without that padding in lists and when displayed separately. Close icons are also of different sizes. Also, Step 4 with assignees doesn't have that additional padding on the left, but that's probably a mistake. What I see is that we add that extra padding under the filter to keep the items visually connected to the combobox, while when they are displayed separately in any other place we don't add that extra padding.

Regarding the "Previous" button on the first step — I still think we must hide it. There are situations when we should keep prev/next, but in dialogs it just makes no sense. I think we should update how the stepper works in dialogs in the Enonic UI and make it the default behavior (and customizable, I guess, at least via styles).

edloidas:
Ah, we had a discussion about selecting and hiding the filter when in "single" mode. It was decided to keep the combobox visible in all dialogs and steps. So, just ignore my suggestion about padding and hiding comboboxes in the steps.

reisfmb:
As discussed via Slack, we had a misunderstanding on the usage of `GridList`.

It's now clear and being used in the proper place, that is, in the list of selected projects below the project selector. 

The hide of _Previous_ button when in first step will be done in npm-enonic-ui repo.

PR #9763, opened 2026-02-02
Enonic UI: Move dialog #9749

### LINKED ISSUE #9749: Enonic UI: Move dialog
Implement move dialog to content studio.

Image

Move dialog needs to contain:
Title for single, plural content with children.
Name/description.
Description.
Item selector.
Move and cancel button.

skjulestad:
Added move dialog with expanded functionality as well as some hooks and a new selector.

**Needs two more hooks and some refactoring before complete**


skjulestad:
Will be refactored later, some hooks may need to be removed.

edloidas:
Had to restore some old functionality that was not on the design (but still required), as well as adjusting type safety. Will probably share a skill to compare it with the old code later. Anyway, great work, Mathias.

Issue #9798, opened 2026-02-04
Enonic UI: Warn before reverting a version before patch


ashklianko:
Show a Confirmation dialog before reverting an item to a version before patch operation

PR #9801, opened 2026-02-05
Enonic UI: Adjust Version History UI #9797

### LINKED ISSUE #9797: Enonic UI: Adjust Version History UI
After latest discussion next actions required: 

- No import event is coming, remove the icon
- Remove border around active version history item
- Fix 'Mark as ready' operation that is now displayed as 'Unknown'
- Show 'Rename' separately from 'Move', use field 'name' for that
- Pick another icon for Move/Rename operations
- Allow "Compare" for all versions in both modes
- Remove "Compare" button,
- On hover, when active or in compare mode (when at least one item selected), version card should display compare checkbox in place of the icon
- Don't expand version card when clicking on the compare checkbox, expand on click in any other area of version entry
- Keyboard: Pressing 'Space' on an active item toggles selection,  pressing 'Enter' expands the item
- Filter out sort operations where 'manualOrderValue' changed

ashklianko:


ashklianko:
#9798 is also here
Tested locally by generating extra versions with patch operations.

ashklianko:
Requires https://github.com/enonic/xp/pull/11741

ashklianko:
Ready to be checked and merged!

PR #9851, opened 2026-02-15
Enonic UI: Project dialog visual problems #9815

### LINKED ISSUE #9815: Enonic UI: Project dialog visual problems
Several visual and keyboard navigation problems in the Project dialog. Some may not be specific to the Project dialog, but since it's the only dialog with steps, we should test and fix them there first.

1. Focus the Next button and press it. The next step is shown, but focus moves to the Dots in the stepper instead of the content's first focusable item
2. "Copy from parent" button uses default styles and outline from legacy CSS. Remove the outline and use the correct focus ring. The button itself needs a design rework. I spoke to Alan and can confirm, that we need to use the same styles: no underline, background change on hover, semibold. Just don't add 3px gap for focus ring and make sure it is horizontally aligned with the label (add just a little inner padding and compensate it via negative margin)
3. Radio group is not tabbable until something is selected. This is an Enonic UI issue (https://github.com/enonic/npm-enonic-ui/issues/327), but must be verified here after it's fixed
4. Roles selector dropdown does not react to arrow navigation (should work like ContentCombobox or Language selector: arrow down marks the first/last active item). Clicking an item in the dropdown prevents typing in the input afterward
5. Applications selector dropdown has the same arrow navigation problem as Roles selector

**Rationale:** These accessibility and styling issues degrade keyboard navigation and visual consistency in the Project dialog.

_Drafted with AI assistance_

reisfmb:


reisfmb:
@edloidas 

As mentioned [here](https://github.com/enonic/app-contentstudio/issues/9815#issuecomment-3898540091), this solves the points you and Alan mentioned.

Now, some observations:

1 - About your point 2 [here](https://github.com/enonic/app-contentstudio/issues/9815#issue-3913653770).
It was not clear to me how the button was intended to be, so this PR just uses our enonic ui `Button` with size sm and a negative margin to make sure the text of the button aligns with the text of the label on the left. If that's not how it is meant to be, please adjust.

2 - About your point 4 [here](https://github.com/enonic/app-contentstudio/issues/9815#issue-3913653770): _"Clicking an item in the dropdown prevents typing in the input afterward"_.
Yes, but this seems to be a bug we have in the combobox when in multiselect / listbox, not related to those specific selectors... Please check [here](https://enonic.github.io/npm-enonic-ui/?path=/story/components-combobox--multi). After clicking on any item in the list, you can't go back to type in the search input.

edloidas (approved):

edloidas:
1. Updated, along with some other styles.
2. I see. Will create an issue for that.

PR #9852, opened 2026-02-15
Enonic UI: Confirm on click outside in New Project dialog #9739

### LINKED ISSUE #9739: Enonic UI: Confirm on click outside in New Project dialog

reisfmb:


edloidas (changes requested):
There are several problems with behavior and UI:
- We already have `ConfirmationDialog` with correct styles in CS. Use it instead (you can just use `ConfirmationDialog.Content`). This would also fix the button color (since action is non-destructive, it must be solid black). And there must be no X Close button in it.
- False positive, if we open dialog when other project was selected (first step become automatically populated with it). Should still be able to close, as no changes were made by the user.
- Clicking outside the confirmation view should still work, but return to the main view, like "cancel", that's what we do now in other dialogs with confirmation. Same for the "Esc" key in confirmation.

edloidas:
Never mind. Looked at the `ConfirmationDialog` and saw that we had to refactor it a bit. Fixed my comments and some old code.

edloidas (approved):

PR #9900, opened 2026-02-23
Context view in Content wizard #9899

### LINKED ISSUE #9899: Context view in Content wizard

reisfmb:


modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/versions/VersionsWidget.tsx
```diff
@@ -5,23 +5,19 @@ import {ContentSummaryAndCompareStatus} from '../../../../../../app/content/Cont
...
 export const VersionsWidget = (): ReactElement => {
-    const isContextOpen = useStore($isContextOpen);
     const activeWidget = useStore($activeWidgetId);
     const isActiveWidget = activeWidget === VERSIONS_WIDGET_KEY;
     const content = useStore($contextContent);
 
-    return (isContextOpen && isActiveWidget && content && <VersionsList />)
-}
-
+    return isActiveWidget && content && <VersionsList />;
```
ashklianko: Why this check was removed? We only render widgets when the context panel is open

modules/lib/src/main/resources/assets/js/v6/features/store/context/contextContent.store.ts
```diff
@@ -1,11 +1,19 @@
...
 import {EditContentEvent} from '../../../../app/event/EditContentEvent';
-import {$currentItem} from '../contentTreeSelection.store';
+import {$currentItem as $treeContent} from '../contentTreeSelection.store';
+import type {ContentSummaryAndCompareStatus} from '../../../../app/content/ContentSummaryAndCompareStatus';
 
-// Re-export for backwards compatibility
-export {$currentItem as $contextContent} from '../contentTreeSelection.store';
+const $wizardContent = atom<ContentSummaryAndCompareStatus | null>(null);
+
+export const $contextContent = computed([$wizardContent, $treeContent], (wizardContent, treeContent) => wizardContent ?? treeContent);
```
ashklianko: In general for this set of code changes:

- $wizardContent is initialised once (via the only call to 'setContextContent'), not kept in sync with webscocket events, though it works in content browse panel. We don't see any changes in widgets after saving the content.
- Let's indroduce 'mode' like state variable for the 'browse' mode and 'wizard' mode, thus we don't try to compute the active content from non-relevant store (tree is not relevant in content wizard)

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/dependencies/index.tsx
ashklianko, line 35: This effect is going to be run not only when content is updated, but even while context is closed/inactive. As a result: unnecessary resolveDependencies calls. Include visibility guards in effect dependencies/checks (isContextOpen, isActiveWidget).

modules/lib/src/main/resources/assets/js/v6/features/store/context/contextContent.store.ts
```diff
@@ -1,14 +1,85 @@
...
+    const content = $contextContent.get();
+
     if (!content) {
         return;
     }
 
     new EditContentEvent([content]).fire();
 };
+
+export const setContextContent = (content: ContentSummaryAndCompareStatus): void => {
```
ashklianko: Noisy and redundant check that pollutes console in browse mode. Just keep  `$wizardContent.set(content); ` here and in `$contentUpdated.subscribe(...` use a condition:

```
if (getMode() === 'wizard') {
    setContextContent(updatedContextContent);
}
```

modules/lib/src/main/resources/assets/js/v6/features/store/context/contextContent.store.ts
```diff
@@ -1,14 +1,85 @@
...
+
+    return () => {
+        contentUnsubscription?.();
+        modeUnsubscription();
+    };
+});
+
+// Update the context content when the content is updated.
+$contentUpdated.subscribe((event) => {
+    const contextContent = $contextContent.get();
```
ashklianko: The function is called immediately on subscribe with empty event, before any socket event is received.
Add a guard: `if (!event?.data?.length) return;`

ashklianko (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/dependencies/index.tsx
reisfmb: Good catch!

ashklianko:
- Added handling of Publsih/Unpublish/Move events
- Disabled PageEditorWidgetItemView in content wizard (it is not working yet ands shows nasty errors in console)

ashklianko (approved):

PR #9987, opened 2026-03-03
Reuse Project Wizard for editing an existing project #9915

### LINKED ISSUE #9915: Reuse Project Wizard for editing an existing project
We will drop the project form and use the Project Wizard dialog for modifying an existing project. 

All of the wizard steps will be shown and function as for a new project with small exceptions:

- [x] 1) Parent project will be readonly - no selector, only label + selected option(s). "NB!..." text should not be shown either.
- [x] 2) If the project doesn't have a parent, hide the entire "Parent Project" input block - only language will be shown on this step
- [x] 3) Identifier on the 2nd step should be shown as static text, not an input
- [x] 4) Changing display name shouldn't change Id
- [x] 5) Change title of the first step to "Default language for project content"
- [x] 6) Hide the help texts (subtexts shown under the step title) on all steps.
- [x] 7) Change caption of the button on the last step from "Create Project" to "Update Project"
- [x] 8) Updating an existing project should refresh the project in the project tree and the right-hand panel that displays project's meta-data
- [x] 9) Dialog title should say "Edit Project: `<project display name>`" instead of "Create Project"

reisfmb:


reisfmb:
PR is in draft mode due to internal Slack discussions on the architecture of this implementation.

Once that is defined and worked on, the PR will be updated and changed again to ready for review.

reisfmb:
PR is now ready for review.

For now, we agreed on reusing the existing code of new project dialog for the edit project dialog, that is, its only one dialog with a single store, and the store provides a mode: 'create' | 'edit' that is used on the steps to customize logic / ui.

edloidas (approved):

PR #9993, opened 2026-03-03
Content Wizard: update workflow according to current state #9986

### LINKED ISSUE #9986: Content Wizard: update workflow according to current state
Workflow icon should follow the same logic as in legacy code.

Also fix Mark as ready here?

ashklianko:


ashklianko:
## What was done

Migrated the Content Wizard's workflow state management from the legacy imperative `WorkflowStateManager` class to the v6 reactive nanostores architecture. The 144-line `WorkflowStateManager.ts` was deleted entirely, and its responsibilities were distributed across reactive stores and subscribers.

## Why

The old `WorkflowStateManager` was tightly coupled to `ContentWizardPanel` — it read directly from the wizard header, form validity, and content state on every `update()` call, which was manually invoked from 5 different call sites. This made the workflow state management brittle, hard to follow, and incompatible with the ongoing migration to v6 nanostores where state should flow reactively through computed atoms rather than being imperatively recalculated.

## Architecture

The new architecture splits workflow state into three concerns:

1. **`$wizardContentState`** (batched computed) — the UI-facing content state (`INVALID`, `READY`, `IN_PROGRESS`), derived from `$wizardDraftDisplayName`, `$wizardDraftName`, `$wizardDraftWorkflowState`, and `$wizardDataValidation`.  Data validation is coming later!!!

2. **`$wizardHasContentChanges` subscriber** — resets `$wizardDraftWorkflowState` from READY to IN_PROGRESS when content is edited. 

3. **`$wizardIsMarkedAsReady`** (computed) — derived directly from `$wizardDraftWorkflowState`, replacing the `markedAsReady` boolean field on `ContentWizardPanel`.

## Practical meaning

- Workflow status is handled 
- Mark as ready works again

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/content/contentState.ts
```diff
@@ -0,0 +1,5 @@
+export enum ContentState {
+    INVALID = 'invalid',
+    READY = 'ready',
+    IN_PROGRESS = 'in-progress',
+}
```
edloidas: It's type, not utils. SHould be in different place and with Capitalized file name.
Also, since it's a new code, I ask you to use string literals instead of enums. Enums is TS are just a hack and don't have native support, they bloat code and considered "bad" option in modern TS.

modules/lib/src/main/resources/assets/js/v6/features/shared/status/DiffStatusBadge.tsx
```diff
@@ -1,15 +1,15 @@
...
 import {useI18n} from '../../hooks/useI18n';
+import {ContentState} from '../../utils/cms/content/contentState';
 import {createCompareStatusKey, createPublishStatusKey, createWorkflowStatusKey} from '../../utils/cms/content/status';
 import {StatusIcon} from '../icons/StatusIcon';
 
 type Props = {
     publishStatus: PublishStatus;
     compareStatus: CompareStatus;
-    workflowStatus?: WorkflowStateStatus;
+    workflowStatus?: ContentState;
```
edloidas: Should be `contentState`.

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/content/contentState.ts
edloidas: I also don't like that we have Status for other things and State here. It's also confusing that we have Status and State across our app. Think we can discuss the naming convention in app?

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -363,15 +443,25 @@ function applyPersistedSectionsSnapshot(snapshot: WizardPersistedSectionsSnapsho
 // * Actions
 //
 
+export function setWizardFormValidation(isValid: boolean): void {
```
edloidas, line 446: Was it added for future use?

modules/lib/src/main/resources/assets/js/app/wizard/action/BasePublishAction.ts
```diff
@@ -57,7 +59,7 @@ export abstract class BasePublishAction
     }
 
     private checkOnExecuted(callback: () => void) {
-        if (this.config.wizard.checkContentCanBePublished()) {
+        if ($wizardContentState.get() !== ContentState.INVALID) {
```
edloidas: Is this check sufficient?

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
ashklianko: Yes

modules/lib/src/main/resources/assets/js/app/wizard/action/BasePublishAction.ts
ashklianko: It mimics the legacy logic where it checked the content form validity state

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/content/contentState.ts
ashklianko: Needed to make a point that this is not pure workflowStatus related thing, we kind of mix here workflow status and validity state (or status?), so some new name was required. ContentStatus?

modules/lib/src/main/resources/assets/js/app/wizard/action/BasePublishAction.ts
edloidas: It was checking `isContentFormValid && contentType && allMetadataFormsValid &&
  allMetadataFormsHaveValidUserInput`? Will state be invalid in similar cases?

ashklianko: We are going to replace it here when new validation is added

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/content/contentState.ts
ashklianko: We have WorkflowState both on the backend and frontend, I'll leave it as ContentState for now

edloidas (approved):

PR #9996, opened 2026-03-04
Version History: adjustments (part 2) #9992

### LINKED ISSUE #9992: Version History: adjustments (part 2)
Fix the widget behaviour to follow the definition [here](https://enonic.atlassian.net/wiki/spaces/~343801284/pages/3836182531/Version+history+widget):
* Only version types with green mark in the "Default list" column should be shown in the default mode
* Only version types with green mark in the "Can diff" column should have a checkbox for comparison. Note that these are the same as the ones displayed in the default mode.
* Only version types with green mark in the "Can restore" column should have a button for restoring the version
* Only version types with green mark in the "Online History" column are able to have the Online/Scheduled/Expired badge. If the actual published version is not of one of these types, we need to find the first earlier version of the correct type and show the badge for it.
* Special handling for "Created" version. If the first version is "created" - it'll be shown as is. If the first version is "imported", then we'll show it as "created". If the first version is anything else, we'll show a "fake" entry using value from "createdTime" field, and it should not be diffable or restorable. 
* "Imported" version can show up anywhere in the list, and we should handle it and show as "Imported".
* Special handling for "Patched" - should be shown in default mode and be diffable only if data fields were patched. Otherwise - shown only in "all actions" mode, non-diffable.

ashklianko:


ashklianko:
See [the comment ](https://github.com/enonic/app-contentstudio/issues/9992#issuecomment-3997499477)

alansemenov (approved):

PR #10035, opened 2026-03-10
Add a tooltip to issue status selector issue #9808

### LINKED ISSUE #9808: Add a tooltip to issue status selector in IssueDialogDetailsContent
1. Open Issues List dialog, click on a issue, Issue Details dialog loads - OK 

Image

**Expected** -  A tooltip should display information about who created or closed the issue

Image

sgauruseu:


edloidas (approved):
Generally good. But I had to do fixes:
- Tooltip misses delay (we should not show it instantly) and proper position.
- Tooltip text is always passed, so there is no point in not wrapping trigger with Tooltip component. If text is empty, tooltip won't show.
- Not part of this task, but user context must be received from the new global store, not from the legacy context class.

PR #10039, opened 2026-03-11
Move action in wizard fails with “Nothing to move” #10038

### LINKED ISSUE #10038: Move action in wizard fails with “Nothing to move”
1. Open  a folder.
2. Click on `Move` button in the wizard toolbar
3. Select a folder in the `Move Content` dialog
4. Click on `Move` button in the dialog

**BUG** -  the action fails with “Nothing to move”

Image

sgauruseu:
Open

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/action/MoveContentAction.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/action/MoveContentAction.ts
@@ -17,7 +17,7 @@ export class MoveContentAction
     }
 
     protected handleExecuted() {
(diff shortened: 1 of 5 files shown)


edloidas (changes requested):
The suggested fix is not fully correct, but in the right direction.

See how we call `openDeleteDialog` and do the similar: we pass the items right into the function from the action and use them after.

Move dialog should also just accept items and calculate the `count` from array length. It must not call `getCurrentItems()` from inside.

edloidas (approved):

PR #10041, opened 2026-03-11
Actions Toolbar: Fold overflowing actions into adjacent menu #9951

### LINKED ISSUE #9951: Actions Toolbar: Fold overflowing actions into adjacent menu
When the actions toolbar contains more buttons than can fit in the available width, overflowed actions should be moved into a foldout menu rendered next to the visible actions. Currently, dense toolbars can wrap, clip, or shift nearby UI depending on viewport size and action label length. The toolbar should keep as many actions visible as possible and place only non-fitting actions in the overflow menu. Overflow behavior should update dynamically on resize and keep action order stable between visible buttons and menu items. Actions moved to the menu must preserve existing behavior, including enabled/disabled state, label semantics, and permission-based visibility. Keyboard navigation and focus handling should remain consistent for both visible buttons and overflowed menu items.

This improves responsive behavior and prevents layout breakage while keeping all actions reachable. It also gives predictable overflow behavior that scales better with localization and future action additions.

Must be used in both Wizard and Browse panels.

To be implemented directly in CS (not in ui lib)!

Attaching a video of how it worked on legacy CS:

https://github.com/user-attachments/assets/87653772-1afd-4358-9ea6-4af742c8880b

skjulestad:
Change to dynamic collapsing toolbar for ContentWizardToolbar.tsx Small stylistic changes as well to elements for better layout for all screen sizing. Added small recipient for mobile chevron layout without focus ring overflowing.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
@@ -36,7 +36,8 @@ const getActionState = (action: Action): ActionState => ({
  * - The buttons are visually joined together as a single unit
  */
 export const SplitActionButton = ({actions, className, triggerClassName, disabled = false}: Props): ReactElement | null => {
-    const [, setRenderVersion] = useState(0);
+    const moreLabel = useI18n('tooltip.moreActions');
+    const [renderVersion, setRenderVersion] = useState(0);
 
     useEffect(() => {
         const subscriptions = actions.map(({action}) => {
@@ -50,7 +51,10 @@ export const SplitActionButton = ({actions, className, triggerClassName, disable
         };
     }, [actions]);
 
-    const actionStates = actions.map(({action}) => getActionState(action));
+    const actionStates = useMemo(
+        () => actions.map(({action}) => getActionState(action)),
+        [actions, renderVersion]
+    );
 
     // Determine primary action index (first enabled action)
     const primaryIndex = useMemo(() => {
@@ -65,8 +69,6 @@ export const SplitActionButton = ({actions, className, triggerClassName, disable
     );
 
     const hasMenuActions = menuStates.length > 0;
-
-    const moreLabel = useI18n('tooltip.moreActions');
 
     if (!primaryState || actions.length === 0) {
         return null;
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
@@ -1,8 +1,7 @@
 import {type Action} from '@enonic/lib-admin-ui/ui/Action';
 import {Button, cn, IconButton, Menu, Toolbar, Tooltip} from '@enonic/ui';
 import {ChevronDown} from 'lucide-react';
-import {type ReactElement, useMemo} from 'react';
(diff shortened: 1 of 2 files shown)


modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
```diff
@@ -41,7 +64,7 @@ export const SplitActionButton = ({actions, className, disabled = false}: Props)
         [actionStates, primaryIndex]
     );
 
-    const hasMenuActions = menuStates.some((state) => state.enabled);
+    const hasMenuActions = menuStates.length > 0;
 
     const moreLabel = useI18n('tooltip.moreActions');
```
ashklianko: Should be on top

modules/lib/src/main/resources/assets/js/v6/features/views/browse/toolbar/SplitActionButton.tsx
```diff
@@ -12,22 +11,46 @@ export type SplitActionButtonAction = {
...
+            action.onPropertyChanged(listener);
+            return {action, listener};
+        });
+
+        return () => {
+            subscriptions.forEach(({action, listener}) => action.unPropertyChanged(listener));
+        };
+    }, [actions]);
+
+    const actionStates = actions.map(({action}) => getActionState(action));
```
ashklianko: actionStates on is a new array every render (actions.map(...)), it is not memoized.  Both primaryIndex (line 56) and menuStates (line 62) depend on it, so their useMemo wrappers never cache

ashklianko (changes requested):
Small changes only, works good!

ashklianko (changes requested):
Hei Mathias, I checked this pull more thoroughly and now see the picture clearer and the direction we need to move to!
The current pull is moving in a good direction. It proves the toolbar can collapse actions dynamically based on available space, which is exactly the behavior we want. The next step is to turn that into a reusable feature instead of keeping the logic inside the wizard toolbar, because we’ll want the same action-row behavior in `BrowseToolbar` as well.

A good rework target is to extract the folding logic into a shared action-row component or hook that only handles one job: given an ordered list of actions, show as many as fit and move the rest into an adjacent overflow split button. `ContentWizardToolbar` should then just provide the actions and render the surrounding layout, while the shared feature owns measurement, overflow calculation, and rendering of visible vs folded actions. If hidden measurement nodes are needed, keep them inside that shared feature so the toolbar itself stays simple.

One possible solution is to introduce a reusable `OverflowActionRow` or `OverflowToolbarActions` component. The input can be a simple ordered list like `{ id, action }`, and internally the component can:
- measure the available width for the row
- render as many actions as fit using `ToolbarActionButton`
- put the remaining actions into `SplitActionButton`
- react to action state changes and recalculate when needed

That would give us a structure where:
- `SplitActionButton` remains the shared primitive for "primary action + dropdown"
- the new overflow component becomes the shared primitive for "action row that folds when space is tight"
- `ContentWizardToolbar` and later `BrowseToolbar` become thin composition layers that just declare which actions belong in the row

What I'd like you to aim for is:
- extract a reusable overflow action-row feature
- update `ContentWizardToolbar` to use it
- keep the API generic enough that `BrowseToolbar` can adopt it next
- fix the `SplitActionButton` disabled-menu behavior
- add a few focused tests around width-based folding and disabled overflow actions

Good work! You're on the right track!

ashklianko:
I'll merge it as is, and create a new task for extracting it into a reusable feature.

Issue #10018, opened 2026-03-09
Input type: MediaSelector


edloidas:
A content picker restricted to media types with inline file upload. Extends ContentSelector behavior but defaults the content type filter to all media types (`ContentTypeName.getMediaTypes()`) and removes the site path restriction (default `allowPath` is empty).

Value type: `REFERENCE` (string content IDs).

#### Config options

Inherited from ContentSelector with different defaults:

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `allowContentType` | `string[]` | all `media:*` types | If config specifies types outside media, they're filtered out; falls back to all media types if none match |
| `allowPath` | `string[]` | `''` (empty) | No site scoping by default |
| `treeMode` | `boolean` | `false` | Tree view toggle |
| `hideToggleIcon` | `boolean` | `false` | Hide tree/flat toggle |

#### Upload

Upload is the key differentiator from ContentSelector. Legacy uses `MediaUploaderEl` integrated directly into the dropdown:

- **Endpoint**: `content/content/createMedia` (POST)
- **Upload target**: current content ID (`context.content.getContentId()`) or root path if no context
- **MIME type filtering**: resolved server-side via `GetMimeTypesByContentTypeNamesRequest` based on allowed content types
- **Multi-file**: allowed when `occurrences.maximum !== 1`
- **Remaining slots**: dynamically computed — upload button hidden when max occurrences reached
- **Drag-and-drop**: supports dragged images from browser (extracts `src` from dropped HTML, handles both data URIs via blob conversion and remote URLs via `CreateMediaFromUrlRequest`)
- **On success**: wraps uploaded Content in `ContentSummaryAndCompareStatus` with `CompareStatus.NEW`, selects it in dropdown, sets content ID property, validates
- **On failure**: removes failed item from selected options silently

The upload button is added as a child element of the dropdown with `extra-button` class. It's shown/hidden in sync with the dropdown and respects occurrence limits.

#### v6 status

No v6 implementation exists. Needs to be built composing the existing v6 `ContentSelector` with upload functionality. The upload mechanism, MIME type resolution, and drag-and-drop handling are new work.

#### References

- ContentSelector issue: #10017
- ImageSelector issue: #9924
- Legacy: `app/inputtype/selector/MediaSelector.ts`, `app/inputtype/ui/upload/MediaUploaderEl.ts`

<sub>*Drafted with AI assistance*</sub>

Issue #10061, opened 2026-03-17
Image selector: ui & code improvements


reisfmb:
- [x] wrap long paths at the beginning rather than in the end. it's more important to see the ending of the path than the beginning. f.ex. "...erylongpath/myfile.png"
- [x] add a tooltip with full path over the path
- [x] selected options should have status as well
- [x] fix drop-down size when filtering
- [x] disable upload button while uploading
- [x] fix combobox / upload button ring adjustment
- [x] fix multi selection bug after multi upload
- [x] use lib admin ui sortable list

Issue #10017, opened 2026-03-09
Input type: ContentSelector


edloidas:
A searchable content picker that allows selecting content items filtered by type and path. Supports dual view mode (tree hierarchy with expand/collapse and flat search list), debounced search (300ms), server-side pagination (50 items per batch), and publish status badges (ONLINE/OFFLINE/PENDING). Single selection hides the combobox after picking, multiple selection uses staged selection with an Apply button. Selected items can be removed; if a previously saved content item has been deleted, it stays visible with an availability status indicator.

Value type: `REFERENCE` (string content IDs).

#### Config options

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `allowContentType` | `string[]` | all types | Content type names to filter by (e.g. `portal:site`); app key prefix is prepended automatically |
| `allowPath` | `string[]` | `${site}` | Path patterns restricting selectable content |
| `treeMode` | `boolean` | `false` | Start in tree view instead of flat list |
| `hideToggleIcon` | `boolean` | `false` | Hide the tree/flat view toggle button |

Selection mode (single vs multiple) is derived from `occurrences.maximum`: single when max=1, multiple otherwise.

#### v6 status

The v6 `ContentSelector` component and its subcomponents are already implemented in `v6/features/shared/selectors/content/`:
- `ContentSelector.tsx` — facade composing `ContentCombobox` + `ContentSelection`
- `ContentCombobox` — search input, tree/flat toggle (Ctrl/Cmd+Shift+V), virtualized list (48px rows, max 300px dropdown), staged selection in multi mode
- `ContentSelection` — GridList of selected items with status badges and remove buttons
- `useContentComboboxData` — data loading hook (tree navigation with per-parent pagination, flat search with global pagination, stale request handling via request ID)
- `useContentComboboxController` — view state management

What's missing is the **input type registration**: a wrapper component that reads `inputConfig` from the form descriptor, maps config to component props (`allowContentType` → `contentTypeNames`, `allowPath` → `allowedContentPaths`, `treeMode`, `hideToggleIcon`), handles `PropertyArray` with REFERENCE values, and registers in the form system so `ResolvedInputField` can render it.

#### API endpoints

- `treeSelectorQuery` (POST) — tree view: initial load, node expansion, per-parent pagination. Params: `parentPath`, `childOrder`, `contentTypeNames`, `allowedContentPaths`, `applicationKey`, `contentId`, `inputName`, `from`, `size`, `expand: 'summary'`
[...]

PR #10092, opened 2026-03-21
Input type: MediaSelector #10018

reisfmb:


reisfmb:
Will open a single PR for #10061, #10018  and #10017

PR #10136, opened 2026-03-26
Incorrect handling of CSV and DOCX preview

### LINKED ISSUE #10134: Incorrect handling of CSV and DOCX preview
Currently .csv and .docx files get downloaded by browser when they are selected in the browse list. Meda types that are not supported for inline preview by browsers should not be treated as renderable.

alansemenov:


modules/app/src/main/resources/admin/extensions/preview-media/preview-media.js
```diff
@@ -48,9 +83,26 @@ exports.canRender = function (req) {
...
+        // For non-image media, verify the MIME type is browser-renderable
+        const mimeType = __.toNativeObject(
+            bean.resolveMimeType(params.id, params.repository, params.branch, params.archive));
+
+        const renderable = isBrowserRenderable(mimeType);
+
+        log.debug(`Media [CAN_RENDER]: ${renderable} (mimeType: ${mimeType})`);
 
-        return canRender;
+        return renderable;
```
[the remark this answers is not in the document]
pmi: Agree, we want as little js-to-java journeys as possible, especially in a high load pathways

modules/app/src/main/resources/admin/extensions/preview-media/preview-media.js
```diff
@@ -4,6 +4,41 @@ const widgetLib = require('/lib/export/widget');
...
+    if (mimeType.indexOf('image/') === 0) {
+        return true;
+    }
+
+    if (mimeType.indexOf('text/') === 0) {
+        return true;
+    }
+
+    return BROWSER_RENDERABLE_TYPES.indexOf(mimeType) >= 0;
+}
```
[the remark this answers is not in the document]
pmi: Agree, it should be decided in the bean as it makes some decisions already

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
```diff
@@ -95,6 +104,12 @@ public boolean canRender( final String contentId, final String repository, final
         return runInAdminContext( repository, branch, archive, () -> canRenderInContext( contentId, editMode ) );
     }
 
+    public String resolveRenderableMimeType( final String contentId, final String repository, final String branch, final boolean archive,
+                                             final boolean editMode )
+    {
+        return runInAdminContext( repository, branch, archive, () -> resolveRenderableMimeTypeInContext( contentId ) );
```
pmi: `editMode`is not passed down. But these methods seem to be not necessary at all.

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
pmi: We can just add `&& isBrowserRenderable(attachment.getMimeType())` here instead of a whole new `resolveRenderableMimeTypeInContext` and `resolveRenderableMimeType` methods to avoid duplication.

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
```diff
@@ -155,6 +170,40 @@ private <T> T runInAdminContext( final String repository, final String branch, f
             .build().callWith( func );
     }
 
+    private String resolveRenderableMimeTypeInContext( final String contentId )
```
pmi: We have 2 methods now that are very similar: `canRenderInContext` and `resolveRenderableMimeTypeInContext`. I suggest we just reuse the existing `canRenderInContext`

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
```diff
@@ -155,6 +170,40 @@ private <T> T runInAdminContext( final String repository, final String branch, f
...
+        return null;
+    }
+
+    private static boolean isBrowserRenderable( final String mimeType )
+    {
+        if ( mimeType.startsWith( "image/" ) )
+        {
+            return true;
+        }
+        if ( mimeType.startsWith( "text/" ) && !mimeType.equals( "text/csv" ) )
```
pmi: It's better to use BROWSER_RENDERABLE_TYPES whitelist and list all text/... types, because this will allow `text/calendar` for example.

pmi (changes requested):

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
alansemenov: it is used and passed from `canRender` check in `preview-media.js`: https://github.com/enonic/app-contentstudio/blob/issue-10134/modules/app/src/main/resources/admin/extensions/preview-media/preview-media.js#L49. IntelliJ doesn't resolve java bean references well, so `editMode` is highlighted as unused.

modules/app/src/main/java/com/enonic/app/contentstudio/widget/MediaRenderingBean.java
```diff
@@ -84,15 +84,23 @@ public class MediaRenderingBean
 
         ).filter( t -> !t.equals( "video/avi" ) ).collect( Collectors.toSet() );
 
+    // Subset of MEDIA_ATTACHMENT_TYPES that browsers can render inline (won't trigger a download)
+    private static final Set<String> BROWSER_RENDERABLE_TYPES = Set.of(
```
pmi, line 88: `MEDIA_ATTACHMENT_TYPES` includes *text/css*, but `BROWSER_RENDERABLE_TYPES` does not. Browsers can render CSS as plain text so it should be present here as well.

pmi (approved):
Looking good! Just one last finding

PR #10164, opened 2026-03-30
Content Wizard allows to save invalid values #10163

### LINKED ISSUE #10163: Content Wizard allows to save invalid values
```
superType: "base:structured"
abstract: false
final: false
allowChildContent: true
displayName: "long_max"
form:
- type: "Long"
  name: "long"
  label: "Long"
  helpText: "Required long input"
  occurrences:
    min: 1
    max: 1
  min: 1
  max: 10

```
1. Create input type with  a config - Set the maximum value for the long input to 10.
2. Open wizard for content with long input
3. Enter a value greater than the configured limit

**BUG - Save button should be disabled  a value greater than the configured limit**

Image

### **Exception thrown after clicking `Save` button**
```
09:50:05.013 ERROR c.e.x.j.i.e.JsonExceptionMapper - Incorrect content property
java.lang.IllegalArgumentException: Incorrect content property
        at com.enonic.xp.core.impl.content.AbstractCreatingOrUpdatingContentCommand.validateForm(AbstractCreatingOrUpdatingContentCommand.java:297)
        at com.enonic.xp.core.impl.content.AbstractCreatingOrUpdatingContentCommand.validateContentData(AbstractCreatingOrUpdatingContentCommand.java:200)
```

edloidas:
Added `$isFormValid` computed store that derives validity from `$wizardDataValidation`, and wired it into `ContentWizardPanel` and `ContentWizardActions` so Save and Publish Request are disabled when form inputs have validation errors (e.g. value exceeds configured max).

Closes #10163

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
@@ -52,6 +52,7 @@ import {openRenameContentDialog} from '../../v6/features/store/dialogs/renameCon
 import {
     $displayName,
     $isContentFormExpanded,
+    $isFormValid,
     $wizardContentState,
     $wizardDraftName,
     $wizardHasChanges,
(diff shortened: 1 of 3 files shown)


edloidas:
It was discussed, that the actual problem in not the wizard. The reason why "Save" button is disabled, is that we treat invalid values as `null` and handle them accordingly. If input was empty and was filled with invalid value, it will be treated as if wasn't changes `null == null`. This, this must be adjusted in numeric input types directly to treat min/max limitations as invalid value as well (as if it was unable to be parsed)

PR #10172, opened 2026-03-31
Don't apply filters to image based content thumbnails #10147

### LINKED ISSUE #10147: Don't apply filters to image based content thumbnails
https://github.com/enonic/app-contentstudio/issues/9461 broke icon thumbnails

ashklianko:


alansemenov:
Can't we tell content type icons from thumbnails? And only apply filter to the former?

ashklianko:
https://developer.enonic.com/docs/xp/stable/cms/content-types#custom_content_types

Content types may use multicolor svg/png, we can't know how applying filter will result for them.

ashklianko:
Two potential solution:
1. Add white background behind icons in dark mode (looks bad bad guarantees all icons are seen)
2. Introduce dart mode icons in content types: <content_type_name.ext> icon for standard/fallback (as it is now) and <content_type_name-dark.ext> for dark mode

No filters for thumbnails ofc

modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx
```diff
@@ -41,13 +41,13 @@ type Props = {
 
 type BuiltInIconProps = Pick<Props, 'contentType'> & React.ComponentProps<LucideIcon>;
 
-const IMAGE_TYPES = new Set<string>([
+const IMAGE_CONTENT_TYPES = new Set<string>([
```
edloidas, line 44: Not sure all those renames are reasonable, since we do not export them, and do not add more similar variables

ashklianko: For sake of clarity - to specify it is not mime types but our own content types

edloidas:
So the problem is with 3rd party icons and images only. I suggest we try drop-shadow instead of other filters for img. Won't look great either, but better than the square border or full background. E.g.:
```css
filter: drop-shadow(0 0 1px rgba(255,255,255,1));
```
image

ashklianko:
Updated PR to not apply filters to image based content types thumbnails. Any other potential solutions/changes to be discussed after the release

PR #10178, opened 2026-03-31
Content Wizard: tabs with invalid forms are not marked #10162

### LINKED ISSUE #10162: Content Wizard: tabs with invalid forms are not marked
Image

and should be:

Image

ashklianko:


modules/lib/src/main/resources/assets/js/v6/features/store/wizardValidation.store.ts
```diff
@@ -21,6 +22,9 @@ import {createDebounce} from '../utils/timing/createDebounce';
 
 export const $validationVisibility = atom<ValidationVisibility>('none');
 
+// Tracks which tabs have invalid forms. Key is tab value ('content' or mixin name), value is true if invalid.
+export const $invalidTabs = map<Record<string, boolean>>({});
```
edloidas: Either rename it (since you store a list of tabs with boolean values, or change type to `Record<string, true>` if you want to use it as a simplified set replacement.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardValidation.store.test.ts
```diff
@@ -0,0 +1,295 @@
...
+    };
+});
+
+vi.mock('@enonic/lib-admin-ui/ValidationErrorHelper', () => ({
+    ValidationErrorHelper: {
+        isCustomError: () => false,
+    },
+}));
+
+const {validateForm} = await import('@enonic/lib-admin-ui/form2');
```
edloidas: Import at the top. Use `vi.mocked(validateForm)`

edloidas (changes requested):

PR #10196, opened 2026-04-01
Initialize page-editor in inline mode to send link clicked event in C…

### LINKED ISSUE #10157: Initialize page-editor in inline mode
Initialize page-editor in inline mode to send link clicked event in CS

pmi:
…S #10157

### DIFF
--- a/modules/app/src/main/resources/assets/js/page-viewer.ts
+++ b/modules/app/src/main/resources/assets/js/page-viewer.ts
@@ -1,0 +1,4 @@
+import {PageEditor} from '@enonic/page-editor';
+
+PageEditor.init(false);
+// console.info('Page editor started in read only mode.');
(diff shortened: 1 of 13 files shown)


modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
```diff
@@ -371,7 +369,8 @@ export class ExtensionRenderingHandler {
                     break;
                 case 'text':
                     try {
-                        frameWindow.addEventListener('click', (event) => this.frameClickHandler(frameWindow, event));
+                        // moved click handling to page-editor package
+                        // it fires ContentPreviewPathChangedEvent from there
                 } catch (error) { /* error */ }
                     break;
```
alansemenov, line 371: What's the practical point of keeping the empty try/catch here?

modules/app/src/main/resources/assets/js/page-viewer.ts
```diff
@@ -0,0 +1,4 @@
+import {PageEditor} from '@enonic/page-editor';
+
+PageEditor.init(false);
+console.info('Page editor started in read only mode.');
```
alansemenov: Remove all these `console.info`. They don't bear any purpose, except for when in the debugging phase, and pollute the console.

PR #10213, opened 2026-04-08
Minor issues in Enonic UI #10161

### LINKED ISSUE #10161: Minor issues in Enonic UI
- [x] Avatar icon in the toolbar should be of the same size as in the Details panel
- [x] Set focus inside the Display name field after opening the Content Wizard
- [x] Display name input should never have the entire border visible. When it’s focused, only its left border should be visible but be thicker than when the field doesn’t have focus. Similarly, when the input is invalid and is focused, only the left border should be visible but be thick and red. So, basically, the same as now but hide all the borders except for the left one.
- [x] Replace the “collapse form icon” with the same as or max similar to the one in [Figma](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=46788-185140&m=dev)
- [x] Toolbar of HtmlArea is not according to the [design](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=51220-125961&m=dev). Should be white icons on black background, both the toolbar and the footer.
- [x] “New Content” dialog should close on pressing “Esc”
- [x] “Ready for publishing” is not shown in the Details panel when secondary (publish) status is “Modified” - it should.
- [x] For unpublished items secondary status is shown as “New”. This is wrong, it should say “Unpublished”.
- [x] Widget dropdown (in the context panel on the right) should close on a click outside
- [x] When secondary status is “Modified”, show “Modified” instead of “Version History” in the preview toolbar:
Image

- [x] When we collapse action buttons inside the toolbar on mobile resolution, move the dropdown to the left, where “Save” button originally is:

Image

skjulestad:
Resize avatar icon, display name focus on opening content wizard, display name left border only, replace collapse form icon, toolbar HtmlArea design fix, escapeable new content, ready for publishing display in details for modified, widget dropdown closeOnBlur, modified/version history label fix, move mobile toolbar in content wizard.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
+++ b/modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
@@ -23,6 +23,8 @@ export class AiContentDataHelper {
 
     public static TOPIC = '__topic__';
 
+    public static TOPIC_PATH = `/${AiContentDataHelper.TOPIC}`;
+
     private static CONFIG_PREFIX = '__config__';
 
     private data: PropertyTree;
@@ -168,5 +170,3 @@ export class AiContentDataHelper {
         return `__${key}__/${pathParts.slice(3).join('/')}`;
     }
 }
-
-export const AI_CONTENT_TOPIC_PATH = `/${AiContentDataHelper.TOPIC}`;
--- a/modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
+++ b/modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
@@ -168,3 +168,5 @@ export class AiContentDataHelper {
         return `__${key}__/${pathParts.slice(3).join('/')}`;
     }
 }
+
+export const AI_CONTENT_TOPIC_PATH = `/${AiContentDataHelper.TOPIC}`;
--- a/modules/lib/src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus.ts
+++ b/modules/lib/src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus.ts
@@ -13,6 +13,7 @@ import {ContentIconUrlResolver} from './ContentIconUrlResolver';
 import {ContentId} from './ContentId';
 import {ContentInheritType} from './ContentInheritType';
 import {type ContentPath} from './ContentPath';
+import {type ContentState} from './ContentState';
 import {ContentSummary, ContentSummaryBuilder} from './ContentSummary';
 import {ContentSummaryAndCompareStatusHelper} from './ContentSummaryAndCompareStatusHelper';
 
@@ -282,13 +283,19 @@ export class ContentSummaryAndCompareStatus implements ViewItem, Cloneable {
     }
 
     isMovedAndModified(): boolean {
-        const contentState = this.contentSummary?.isReady()
-            ? 'ready'
-            : this.contentSummary?.isInProgress()
-                ? 'in-progress'
-                : null;
+        return CompareStatusChecker.isMovedAndModified(this.getCompareStatus(), this.getContentState());
+    }
 
-        return CompareStatusChecker.isMovedAndModified(this.getCompareStatus(), contentState);
+    private getContentState(): ContentState | null {
+        if (this.contentSummary?.isReady()) {
+            return 'ready';
+        }
+
+        if (this.contentSummary?.isInProgress()) {
+            return 'in-progress';
+        }
+
+        return null;
     }
 
     canBeMarkedAsReady(): boolean {
--- a/modules/lib/src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus.ts
+++ b/modules/lib/src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus.ts
@@ -282,7 +282,13 @@ export class ContentSummaryAndCompareStatus implements ViewItem, Cloneable {
     }
 
     isMovedAndModified(): boolean {
-        return CompareStatusChecker.isMoved(this.getCompareStatus()) && this.contentSummary?.isInProgress();
+        const contentState = this.contentSummary?.isReady()
+            ? 'ready'
+            : this.contentSummary?.isInProgress()
+                ? 'in-progress'
+                : null;
+
+        return CompareStatusChecker.isMovedAndModified(this.getCompareStatus(), contentState);
     }
 
     canBeMarkedAsReady(): boolean {
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/html-area/HtmlAreaInput.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/html-area/HtmlAreaInput.tsx
@@ -325,7 +325,7 @@ const CKEditorWrapper = ({
         <>
             <div data-name={CKEDITOR_WRAPPER_NAME} className={cn(
                 'html-area rounded-sm *:rounded-sm transition-highlight',
-                focused && 'ring-3 ring-offset-3 ring-offset-ring-offset',
+                focused && 'ring-2 ring-offset-0 ring-offset-ring-offset',
                 focused && (hasError ? 'ring-error' : 'ring-ring'),
                 hasError && 'has-error [&_.cke_chrome]:!border-error',
             )}>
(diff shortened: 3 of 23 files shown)


modules/lib/src/main/resources/assets/js/app/ai/AiContentDataHelper.ts
```diff
@@ -168,3 +168,5 @@ export class AiContentDataHelper {
         return `__${key}__/${pathParts.slice(3).join('/')}`;
     }
 }
+
+export const AI_CONTENT_TOPIC_PATH = `/${AiContentDataHelper.TOPIC}`;
```
edloidas: I don't like the legacy code we have, but mixing it with object-oriented approach with more traditional one and the same file is not that great. Let's create another static property inside of this helper and use it.

modules/lib/src/main/resources/assets/js/app/content/ContentSummaryAndCompareStatus.ts
```diff
@@ -282,7 +282,13 @@ export class ContentSummaryAndCompareStatus implements ViewItem, Cloneable {
     }
 
     isMovedAndModified(): boolean {
-        return CompareStatusChecker.isMoved(this.getCompareStatus()) && this.contentSummary?.isInProgress();
+        const contentState = this.contentSummary?.isReady()
+            ? 'ready'
+            : this.contentSummary?.isInProgress()
+                ? 'in-progress'
+                : null;
```
edloidas: For readability I'd rather move this to separate method:
```ts
getContentState(): string | null {
    if (this.contentSummary?.isReady())      return 'ready';
    if (this.contentSummary?.isInProgress()) return 'in-progress';
    return null;
}
```

modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/html-area/HtmlAreaInput.tsx
```diff
@@ -325,7 +325,7 @@ const CKEditorWrapper = ({
         <>
             <div data-name={CKEDITOR_WRAPPER_NAME} className={cn(
                 'html-area rounded-sm *:rounded-sm transition-highlight',
-                focused && 'ring-3 ring-offset-3 ring-offset-ring-offset',
+                focused && 'ring-2 ring-offset-0 ring-offset-ring-offset',
```
edloidas: Are those changes intentional? The HtmlArea would be different from Input and TextArea after this. Better show Alan, I guess

modules/lib/src/main/resources/assets/js/v6/features/shared/items/ContentListItemSelectable.tsx
```diff
@@ -67,6 +68,7 @@ export const ContentListItemSelectable = ({
                 {status && <DiffStatusBadge
                     publishStatus={content.getPublishStatus()}
                     compareStatus={content.getCompareStatus()}
+                    contentState={calcContentState(content.getContentSummary())}
```
edloidas: Probably lacks useMemo

modules/lib/src/main/resources/assets/js/v6/features/shared/lists/content-row/ContentRowStatus.tsx
```diff
@@ -23,6 +24,7 @@ export const ContentRowStatus = ({
                 <DiffStatusBadge
                     publishStatus={content.getPublishStatus()}
                     compareStatus={content.getCompareStatus()}
+                    contentState={calcContentState(content.getContentSummary())}
```
edloidas: Same `useMemo` missing?

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
```diff
@@ -28,6 +31,10 @@ export const DisplayNameInput = (): ReactElement => {
         }
     };
 
+    const handleFocus = useCallback((_event: FocusEvent<HTMLInputElement>): void => {
+        new AiContentOperatorSetContextEvent(AI_CONTENT_TOPIC_PATH).fire();
+    }, []);
```
edloidas: All data is static here. It doesn't need a callback and can be just moved outside as:
```ts
function setAIContext(): void {
    new AiContentOperatorSetContextEvent(AI_CONTENT_TOPIC_PATH).fire();
}

// ...
    onFocus={setAIContext}
```

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
```diff
@@ -1,13 +1,16 @@
...
-import {useState, type ReactElement} from 'react';
+import {type FocusEvent, useCallback, useState, type ReactElement} from 'react';
+import {AI_CONTENT_TOPIC_PATH} from '../../../../../app/ai/AiContentDataHelper';
 import {useI18n} from '../../../hooks/useI18n';
 import {EditableText} from '../../../shared/primitives/EditableText';
 import {$displayName, setDraftDisplayName} from '../../../store/wizardContent.store';
 import {$validationVisibility} from '../../../store/wizardValidation.store';
 
-const DISPLAY_NAME_INPUT_NAME = 'DisplayNameInput';
+export const DISPLAY_NAME_INPUT_NAME = 'DisplayNameInput';
```
edloidas: No need to export it

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
```diff
@@ -1,13 +1,16 @@
...
+export const DISPLAY_NAME_INPUT_SELECTOR = `input[data-component="${DISPLAY_NAME_INPUT_NAME}"]`;
```
edloidas: And this one, if you move all the logic here.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/ContentWizardTabsToolbarElement.ts
```diff
@@ -1,8 +1,20 @@
...
+    focusDisplayNameInput(): boolean {
+        const input = this.getHTMLElement().querySelector<HTMLInputElement>(DISPLAY_NAME_INPUT_SELECTOR);
+
+        if (!input || input.disabled || input.offsetParent === null) {
+            return false;
+        }
+
+        input.focus();
+        return document.activeElement === input;
+    }
```
edloidas: I'm sure this can be done inside our components instead of a legacy element. And we would need that when we move fully to React. 

All you need to make focus work properly in DisplayNameInput is to call focus on layout:
```tsx
useLayoutEffect(() => {
    inputRef.current?.focus();
}, [inputRef.current]);
```
Unfortunately, this is also a hack, because ideally we won't need the `inputRef.current` in dependencies and could use `useEffect`. But due to the complex rendering patter of our app this is more robust solution. Try it.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/html-area/HtmlAreaInput.tsx
alansemenov: The only fix that concerns HtmlArea is this one:

> Toolbar of HtmlArea is not according to the [design](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=51220-125961&m=dev). Should be white icons on black background, both the toolbar and the footer.

I'm not sure if fixing these styles will actually fix colors of HtmlArea's toolbar. We can go through Mathias' fixes in the status meeting tomorrow.

edloidas: The icons fix is below in `modules/lib/src/main/resources/assets/styles/inputtype/text/htmlarea/html-editor.less`. This changes the focus ring around the `HtmlArea`, removing the offset and reducing the ring width.

skjulestad: This was according to design, but reverted after some clarification in the next commit resolving comments currently in progress.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
skjulestad: Moving logic /Users/msk/Documents/EnonicReps/app-contentstudio/modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/ContentWizardTabsToolbarElement.ts requires more changes, keep exported for now, might revisit later?

edloidas (approved):

PR #10218, opened 2026-04-09
Form rendering populates property array and makes Save enabled #10206

### LINKED ISSUE #10206: Form rendering populates property arrays and makes Save enabled
Partially relates to https://github.com/enonic/lib-admin-ui/issues/4330.

When we create a new content in CS we get an empty `.data` field in this content after initial create. Form rendering populates empty property arrays for every input so we could add/remove values, plus we add input occurrences according to `min` config value.

This means that persisted data value is not equal to the rendered value and we get Save button enabled.

As a fix let's make the wizard use rendered content form data as a base value for comparison, however this solution has it's potential flaws. For example: what if at some point we start rendering only visible part of the content form?

Other solution to be discussed: don't persist a new content until we press Save, just open a new content form when triggering content creation. Pros: initial content create version can have a display name and desired values for inputs! Current content creation flow creates empty content template with no display name and data, thus we always need an extra version even for a display name!

ashklianko:


ashklianko:
Some notes about changes made:

## The Problem

When `InputField` (from `@enonic/lib-admin-ui/form2`) renders, it creates `PropertyArray` objects and seeds null values on the draft `PropertyTree` in `useEffect`. These structural additions don't exist in the persisted data. The comparison `ContentDiffHelper.dataEquals(persisted, draft, false)` detects them as real changes, incorrectly enabling Save.

## How It Works

Three private atoms in `wizardContent.store.ts` track snapshot state:

| Atom | Purpose |
|------|---------|
| `$needsRenderedSnapshot` | Flag for content data, armed at init |
| `$mixinsPendingMount` | Mixin names waiting for their lazy tab to mount |
| `$mixinsNeedingSnapshot` | Mixins whose tab has mounted and whose snapshot window is open |

### Arming

`armRenderedSnapshot()` is called only from `initializeWizardContentState()`, never from `setPersistedContent()`. This distinction matters: after a save, the wizard is already rendered — re-arming would risk swallowing the next real edit.

### Snapshotting

`snapshotContentBaseline()` and `snapshotMixinBaseline()` copy the enriched draft tree back to the persisted atom, making the comparison baseline match the rendered state. The enrichment then looks like "no change."

### Why `setTimeout(0)` Alone Doesn't Work

InputField enrichment happens in React `useEffect` (a macrotask). `setTimeout(0)` is also a macrotask. Their execution order isn't guaranteed — the disarm timer can fire *before* React effects, clearing the flag before enrichment occurs.

### Component-Driven Disarm

React guarantees **child effects fire before parent effects**. This is the key invariant the mechanism relies on.

**`ContentForm`** calls `notifyContentFormMounted()` in `useEffect`. By the time this runs, all child `InputField` effects have already enriched the tree. The function schedules a `queueMicrotask` to snapshot (capturing the enrichment) and a `setTimeout(0)` to disarm (safety net if no enrichment occurred).

The `isReady` guard ensures the effect only fires when the form actually renders its children — not on the initial render that returns `null`.

**`MixinView`** calls `notifyMixinMounted(mixinName)` in `useEffect`. Same pattern, but per-mixin. This moves the mixin from the "pending" set to "armed," schedules a snapshot via `queueMicrotask`, and sets up a per-mixin disarm timer. This handles lazy tabs correctly — each mixin gets its own snapshot window starting when its tab actually mounts, not at wizard init.

### Execution Order

```
initializeWizardContentState() → armRenderedSnapshot()
  → $needsRenderedSnapshot = true, $mixinsPendingMount = {seo, og, ...}

React renders ContentForm → FormRenderer → InputField

Effects fire bottom-up:
  1. InputField useEffect → propertyArray.add(null) → tree.onChanged
     → bumpVersion → $wizardDataVersion subscriber → queueMicrotask(snapshot)
  2. ContentForm useEffect → notifyContentFormMounted()
     → queueMicrotask(snapshot) + setTimeout(disarm)

Microtask queue:
  → snapshotContentBaseline() → copies draft to persisted → flag cleared

Macrotask queue:
  → setTimeout fires → flag already cleared → no-op
```

## Related Fixes

### Strict Mixin Comparison

`mixinsEqual()` was changed from `ContentDiffHelper.dataEquals(a, b, true)` to `false`. The `true` variant strips nulls, empty strings, and `false` booleans before comparing, which masked real user changes like adding a new occurrence (which starts as null). The snapshot mechanism now handles the false positives that `ignoreEmptyValues` previously masked.

### Bidirectional Workflow State

The `$wizardHasContentChanges` subscriber previously only downgraded READY → IN_PROGRESS when content was edited. It now also restores READY when all content changes are reverted.

The snapshot functions no longer call `restoreWorkflowIfNeeded()` directly — that could blindly restore READY even when other sections were still dirty (e.g., opening a lazy mixin tab after editing something else). The subscriber handles all workflow restoration and only restores when *all* content sections are clean.

### Auto-Naming Restoration

In `wizardToolbar.store.ts`, the display name subscriber auto-generates a content name. When the user clears the display name back to the persisted value, the persisted content name is now restored instead of generating `ContentUnnamed`.

### Post-Save Mixin Pending Set

`setPersistedContent()` (called after save) adds newly persisted mixins to `$mixinsPendingMount`, so their first lazy mount is properly handled even if they were enabled during the current session.

edloidas:
To be honest this looks complicated AF. Along with hack for display name focus this require serious revision after the whole app moved to React.

edloidas (approved):

PR #10229, opened 2026-04-10
Content online status and Content grid: improvements #10148

### LINKED ISSUE #10148: Content online status and Content grid: improvements
After recent changes for XP 8 we got `publishInfo` that now includes enough info to correctly calculate content's status (with condition that Import and Export operations are fixed and set require publish info fields). Currently we always send a compare request with ids of items we need to calc, thus ContentSummaryAndCompareStatus is created.
The idea is to get rid of this request because compare operation on the backend is expensive, and ditching it can improve performance.

- Investigate on possibility and complexity to remove compare requests. Implement if it is not going to take much time.

ashklianko:


ashklianko:
### Why

After recent backend changes, all content statuses can be calculated client-side from the content's publish timestamps. The expensive `compare` operation that ran on every content load is no longer needed. It was removed from the default data loading path entirely.

For the few places that still need to detect specific changes (e.g. "Moved" status in publish dialog and details widget), a lightweight field-level diff endpoint was introduced — only called where actually necessary.

### 1. Removed compare from default content loading

- `ContentSummaryAndCompareStatusFetcher` — no longer calls compare on every load
- `ContentSummaryOptionDataLoader` — `loadStatuses()` passes through without compare
- `GetDescendantsOfContents` — removed `filterStatuses` that depended on compare

### 2. Introduced targeted field-level diff

For cases where specific change detection is needed, the compare endpoint was replaced with a diff that returns changed field names (path, displayName, data, mixins, language, owner, inherit).

- `ContentResource.compare()` — rewritten with batch `getByIds` and direct field comparison
- `CompareContentResultJson` — `record(String id, Set<String> diff)`
- `ContentDiffField` enum, TypeScript `compare.ts` API, `formatCompareResult.ts` utility
- `PublishDialogItemStatus.tsx` — consumes diff results where needed

### 3. Switched from `ContentSummaryAndCompareStatus` to `ContentSummary`

Without automatic compare, the CSCS wrapper became dead weight. Removed throughout v6 code.

- 5 browse actions — `getCurrentItems()` directly
- Sort dialog — types switched to `ContentSummary`
- ~100+ v6 files — stores, components, selectors, status badges updated

edloidas:
Look at the falling test in [testInputTypes_2](https://github.com/enonic/app-contentstudio/actions/runs/24291353252/job/70929264950?pr=10229) and see if this could be the reason. The other failing test is not related.

PR #10230, opened 2026-04-13
Incorrect DnD behaviour in the Sort dialog #10228

### LINKED ISSUE #10228: Incorrect DnD behaviour in the Sort dialog
We should use the same library for manual sorting of items inside the Sort dialog as we are using in other places where we are using DnD: selected options in selectors, SiteConfigurator etc. Behaviour and styling should be exactly the same: drop shadow for the dragged component, other items in the list rearranging to show where the item will be placed etc.

skjulestad:
Use DnD Core for sort dialog as it is used elsewhere, keep focus functionality for now.


skjulestad:
After some feedback I will look into dropping sortablelist and import from lib-admin UI while maintaining focus and navigation logic.

Issue #10193, opened 2026-04-01
Page Components View component


ashklianko:
https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=44954-132427&m=dev

Notes:
- Add sort icon on the draggable items
- No dots menu on the right of each component (use right click)
- New icons to be used for page components
- Show descriptor dropdown after inserting a new component
- Empty placeholder to be shown when a component is not selected and not focused
- On click on component open inspect panel as before, but don't open closed inspect for empty/non-selected components
- Add invalid icon for the tree items with invalid config, icon must be shown even before rendering the form in inspect panel. (Check server validation errors?)

PR #10243, opened 2026-04-14
Page Components View component #10193

ashklianko:


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/4425

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsContextMenu.tsx
edloidas, line 1: This will be imported from the page editor package, when new version is ready. It's okay to have it as is, but don't do much to customize it.

edloidas (commented):
Try to isolate this component as much as you could. New UI components should come from Enonic UI and Page Editor. Page data model would probably come from the page editor as well.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsContextMenu.tsx
ashklianko: Ok!

PR #10259, opened 2026-04-17
Support new Style descriptor #10258

### LINKED ISSUE #10258: Support new Style descriptor
Asset Delivery via Content Studio Admin Extension

`/_/admin:extension/com.enonic.contentstudio:style/<app-key>/editor.css`

anatol-sialitski:
Depends on https://github.com/enonic/xp/pull/11945

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/style/StyleDescriptorMapper.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/style/StyleDescriptorMapper.java
@@ -36,11 +36,8 @@ public class StyleDescriptorMapper
         final Map<String, ImageStyle> styles = new LinkedHashMap<>();
         final Map<String, ApplicationKey> stylesApps = new LinkedHashMap<>();
 
-        gen.array( "css" );
         for ( StyleDescriptor styleDescriptor : this.styleDescriptors )
         {
-            gen.value( styleDescriptor.getCssPath() );
-
             for ( var styleElement : styleDescriptor.getElements() )
             {
                 if ( styleElement instanceof ImageStyle )
@@ -50,7 +47,6 @@ public class StyleDescriptorMapper
                 }
             }
         }
-        gen.end();
 
         gen.array( "app" );
         for ( StyleDescriptor styleDescriptor : this.styleDescriptors )
@@ -73,7 +69,7 @@ public class StyleDescriptorMapper
 
         gen.value( "element", "image" );
         gen.value( "name", element.getName() );
-        gen.value( "displayName", localizeDisplayName( element, applicationKey ) );
+        gen.value( "displayName", localizeLabel( element, applicationKey ) );
 
          serializeImage( gen, element );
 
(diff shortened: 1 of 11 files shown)


modules/app/src/main/java/com/enonic/app/contentstudio/style/StyleDescriptorMapper.java
```diff
@@ -73,7 +69,7 @@ private void serializeElement( final MapGenerator gen, final ImageStyle element,
 
         gen.value( "element", "image" );
         gen.value( "name", element.getName() );
-        gen.value( "displayName", localizeDisplayName( element, applicationKey ) );
+        gen.value( "displayName", localizeLabel( element, applicationKey ) );
```
rymsha:
```suggestion
        gen.value( "label", localizeLabel( element, applicationKey ) );
```

modules/app/src/main/resources/admin/extensions/style/style.js
```diff
@@ -0,0 +1,19 @@
+exports.get = function (req) {
+    const match = req.path.match(/com.enonic.app.contentstudio:style\/([^/]+)\/editor\.css/);
+    if (!match) {
+        return {
+            status: 400,
```
rymsha:
```suggestion
            status: 404,
```

modules/app/src/main/resources/admin/extensions/style/style.yaml
```diff
@@ -0,0 +1,4 @@
+kind: "AdminExtension"
+title: "Style Extension"
+interfaces:
+  - "generic"
```
rymsha: We don't want styles to be available to any admin tool.

modules/app/src/main/resources/admin/extensions/style/style.js
```diff
@@ -0,0 +1,19 @@
+exports.get = function (req) {
+    const match = req.path.match(/com.enonic.app.contentstudio:style\/([^/]+)\/editor\.css/);
```
rymsha: This is a bit convoluted. Why would extension need to deal with an app name (which is always a constant inside every extension)
Besides this regex is incorrect: does not mask all dots

modules/app/src/main/resources/apis/styles/styles.js
```diff
@@ -1,6 +1,6 @@
 /*global __*/
 
-var portalLib = require('/lib/xp/portal');
+var adminLib = require('/lib/xp/admin');
 
 exports.get = function (req) {
```
rymsha: Old syntax. `export.GET`

BTW why js, not ts?

rymsha (changes requested):

modules/app/src/main/resources/apis/styles/styles.js
anatol-sialitski: Currently, all APIs, extensions, etc. are written in JavaScript, not TypeScript. This can be done later.

modules/app/src/main/resources/admin/extensions/style/style.yaml
anatol-sialitski: What about `contentstudio.style` or `contentstudio.richtextstyle`?

alansemenov: `contentstudio.style` is ok

modules/app/src/main/resources/apis/styles/styles.js
rymsha: `exports.GET`

rymsha (approved):

alansemenov (approved):

PR #10263, opened 2026-04-20
Version History: adjustments (part 4) #10246

### LINKED ISSUE #10246: Version History: adjustments (part 4)
Adjust UI part with the latest XP changes:
- Fetch editorial change version from the backend instead of calculating it in the versions list
- Remove childOrder from getVersion response
- Remove 'permissionsChanged' from getVersions response
- Remove modifier and modified time, they are duplicated of operation fields
- Check 'path' field if it is needed

ashklianko:


modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentPublishInfoResolver.java
```diff
@@ -18,4 +19,16 @@ public ContentPublishInfo resolvePublishInfo( final ContentId contentId, Content
...
+    public boolean versionExists( final ContentId contentId, final ContentVersionId contentVersionId )
+    {
+        try
+        {
+            return contentService.getByIdAndVersionId( contentId, contentVersionId ) != null;
+        }
+        catch ( ContentNotFoundException e )
+        {
+            return false;
+        }
```
rymsha, line 32: it never returns null

modules/lib/src/main/resources/assets/js/app/resource/json/ContentVersionActionJson.ts
```diff
@@ -1,7 +1,9 @@
-
 export interface ContentVersionActionJson {
     operation: string;
     fields: string[];
+    editorial: string | null;
+    editorialExists: boolean | null;
```
rymsha: a bit of a duplicate for dto

ashklianko: not really - we need to handle the case when editorial version is removed, then we are going to mark the original publish version with the publish state label.

PR #10266, opened 2026-04-20
AdminSiteHandler: resolve nearest site if missing #10265

### LINKED ISSUE #10265: AdminSiteHandler: resolve nearest site if missing
When a URL handled by `AdminSiteHandler` (preview/inline/admin/edit modes) points to a path that does not resolve to an existing content, the site is not set on the `PortalRequest`. As a result, site/controller mappings that match virtual URLs (URLs served by a controller mapping on the nearest site without a backing content) are never considered, and admin preview returns 404 for pages that render correctly in live mode. XP's own `SiteHandler` handles this by falling back to `ContentService.findNearestSiteByPath(contentPath)` when the content lookup returns null.

#### Rationale
Admin rendering should be consistent with live rendering: if a mapping can serve the URL in live mode, it should also be reachable from admin preview. Without the nearest-site fallback, mappings on the nearest site cannot run and previewing dynamically-served pages is impossible.

#### References
- XP `SiteHandler`: https://github.com/enonic/xp/blob/master/modules/portal/portal-impl/src/main/java/com/enonic/xp/portal/impl/handler/portal/SiteHandler.java

rymsha:
Falls back to `ContentService.findNearestSiteByPath(contentPath)` in `AdminSiteHandler` when the content lookup returns null, mirroring XP's `SiteHandler`. Applied in both edit and preview/inline/admin branches so site mappings can handle virtual URLs in admin preview the same way they do in live mode. The root-content case (content resolved to root) is preserved as-is.

Closes #10265

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/site/AdminSiteHandler.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/site/AdminSiteHandler.java
@@ -136,7 +136,12 @@ public class AdminSiteHandler
             final Content content =
                 contentById != null ? contentById : callAsContentAdmin( repositoryId, branch, () -> this.getContentByPath( contentPath ) );
 
-            if ( content != null && !content.getPath().isRoot() )
+            if ( content == null )
+            {
+                portalRequest.setSite(
+                    callAsContentAdmin( repositoryId, branch, () -> this.contentService.findNearestSiteByPath( contentPath ) ) );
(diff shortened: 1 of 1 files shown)


pmi (approved):

pmi:
Should we have tests for that ?

rymsha:
I hope this magic will be reimplemented into TypeScript eventually. so no tests for now

PR #10270, opened 2026-04-21
ContextWindow: cross-iframe drag and drop #10191 

### LINKED ISSUE #10191: ContextWindow: cross-iframe drag and drop
Wire up cross-iframe drag and drop for the Insert tab. Users should be able to drag component types from the insertables list into the live page iframe to add new components.

The current implementation uses jQuery UI Draggable with a transparent drag mask over the iframe and a cloned draggable element inside the iframe. In v6, use dnd-kit in the outer window with a `PointerSensor` and a droppable zone covering the iframe rect. When the pointer enters the iframe zone, relay position and drop intent via `postMessage` to the iframe. The iframe handles placement internally.

The transparent drag mask trick (intercepting pointer events before they cross the iframe boundary) is still needed regardless of library choice.

#### Rationale

This is the most complex piece of the ContextWindow. Deferred from the shell implementation to allow the rest of the UI to be built and tested without it.

edloidas:
Made the right-side Context panel open automatically when a component is selected in the page editor, and stopped the left-side wizard form from auto-collapsing when that happens.

- `bridge.ts`: on SELECT/INSPECT, call `setContextOpen(true)` so the existing `$isContextOpen` subscription in `ContentWizardPanel` opens the panel. Export `$inspectedPath` from the page-editor store barrel.
- `ContentWizardPanel.ts`: removed the viewport-width `InspectEvent` listener that auto-collapsed the form on selection. Pruned the now unused `InspectEvent` and `PageNavigationEventSource` imports.
- `PageEditorExtension.tsx` / `usePageEditorTabs.ts`: drive the inspect-tab auto-switch from `$inspectedPath` instead of the boolean `$isInspecting`, so re-selecting a different item re-triggers the switch.
- `useInsertableDrag.ts`: use `document.elementFromPoint` for iframe hit-testing instead of raw bounding-box math.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -178,8 +178,8 @@ importers:
         specifier: ^1.11.20
         version: 1.11.20
       dompurify:
-        specifier: ~3.3.3
-        version: 3.3.3
+        specifier: ~3.4.0
+        version: 3.4.0
       fine-uploader:
         specifier: ^5.16.2
         version: 5.16.2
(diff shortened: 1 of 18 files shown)


edloidas:
Requires new `@enonic/page-editor` version

edloidas:
Closed in favor of #10282

PR #10278, opened 2026-04-21
Item and Option sets #10167 #10250

### LINKED ISSUE #10167: ItemSet
Based on:
- [Lib Admin UI PR](https://github.com/enonic/lib-admin-ui/pull/4382)
- [Content Studio PR](https://github.com/enonic/app-contentstudio/pull/10155)

### LINKED ISSUE #10250: OptionSet

reisfmb:
Depends on https://github.com/enonic/lib-admin-ui/pull/4431


edloidas (changes requested):
Very impressive work, Bruno. 

Before we merge, there are few things I'd like to adjust and fix:
- [x] I suggest we drop arrow navigation for Single Selection and keep all things navigable with Tab instead, like in Multi Selection. Otherwise, we'll face few stylistic and behavioral issues that would be hard to resolve. It's just not worth it and adds complexity in such generic interfaces we can't fully control.
- [x] Inside `*SetOccurrenceView` components we have `<button>` as a header. It lacks the `:focus-visible` styles like other Enonic UI buttons and shows default blue box. But we don't need gap between element and focus ring there. Just make it dark, 2px wide with no gap.
- [x] We have another button inside for the drag handling inside this button. This is incorrect and technically makes it 3rd button in a nesting chain. I'd like you to move it on the same level as the header button, not inside it. It's totally okay for it not toggling the occurrence.

edloidas:
There is also a related occurrence error, but should be fixed in a separate issue https://github.com/enonic/app-contentstudio/issues/10291

edloidas (approved):

PR #10281, opened 2026-04-22
Updating display name changes content path #10253

### LINKED ISSUE #10253: Updating display name changes content path
1. Open a folder that has  'Online' status.
2. Update the display name and click on `Save` button.

Image

### **BUG:**    **Updating display name changes content path and sets online content to “moved” status**

Image

skjulestad:
Only update the path in sync with displayName for non-published content.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/ContentWizardToolbar.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/ContentWizardToolbar.ts
@@ -21,6 +21,7 @@ import {subscribe as subscribeToCollaborators} from '../stores/collaboration';
 import {UrlHelper} from '../util/UrlHelper';
 import {type ContentWizardActions} from '../wizard/action/ContentWizardActions';
 import {ContentWizardToolbarPublishControls} from '../wizard/ContentWizardToolbarPublishControls';
+import {type PublishStatus, PublishStatusChecker} from '../publish/PublishStatus';
 import {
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/app/browse/ContentWizardToolbar.ts
```diff
@@ -166,12 +168,12 @@ class ContentWizardToolbarElement extends V6ContentWizardToolbarElement {
...
     setItem(item: ContentSummaryAndCompareStatus): void {
         this.item = item;
-        const publishStatus = item?.getPublishStatus() ?? $wizardToolbar.get().publishStatus ?? null;
+        const publishStatus = this.resolveToolbarPublishStatus(item);
         const contentPath = this.resolveToolbarContentPath(item);
         setWizardToolbarPublishStatus(publishStatus);
         setWizardToolbarContentPath(contentPath);
         setWizardToolbarCanRenameContentPath(!!item?.getPath());
-        setWizardToolbarIsContentOnline(!!item?.isOnline());
+        setWizardToolbarIsContentOnline(publishStatus != null && PublishStatusChecker.isOnline(publishStatus));
```
ashklianko: 1. `publishStatus != null` is redundant
2. Doesn't take  into account  online Scheduled and Expired content: path should not change for them too

ashklianko (changes requested):

Issue #10279, opened 2026-04-22
Page Editor modernization


edloidas:
The Page Editor is a core part of Content Studio responsible for visual editing of pages, regions, and components. As the CS admin UI is migrated to the new v6 Preact-based stack, the Page Editor and its surrounding integration points need to work seamlessly with the modernized UI. This epic tracks the coordination of work required to align the Page Editor with the new UI architecture — including its dedicated `@enonic/page-editor` package, event bus integration, reuse of components from `@enonic/ui`, Nanostores-based state synchronization, and theming via Tailwind v4.

#### Rationale

Modernizing the Page Editor is a prerequisite for a cohesive v6 experience. Keeping the legacy editor wired into the new admin UI leads to inconsistent styling, duplicated state, and drift between class-based and Preact-based code paths. This umbrella issue groups the individual migration, integration, and polish tasks so progress can be tracked in one place.

#### Notes

- Clicking a component directly inside the Page Editor canvas should not trigger a scroll — the component is already in view. Scrolling should only happen when navigation comes from outside the canvas (PageComponentsView tree, search, etc.) and the component is currently outside the viewport.

<sub>*Drafted with AI assistance*</sub>

PR #10282, opened 2026-04-22
Page Editor modernization #10279

### LINKED ISSUE #10191: ContextWindow: cross-iframe drag and drop

### LINKED ISSUE #10269: InsertPanel: add Apply button
Add an Apply button to `InsertPanel` in the page editor context widget that lets users explicitly commit pending changes to the page. Currently the panel only lists draggable items (part, layout, text, fragment) with no confirmation action. The Apply button should trigger the same page-apply flow used elsewhere in the editor, giving users explicit control over when changes take effect.

### LINKED ISSUE #10280: Migrate to new version of @enonic/page-editor package
Upgrade `@enonic/page-editor` from `^0.0.6` to the new version in both `modules/app/package.json` and `modules/lib/package.json`. The new package release introduces API changes required for the modern v6 UI and cleans up the existing initialization surface. The primary integration point is `modules/app/src/main/resources/assets/js/page-editor.ts`, which calls `PageEditor.initEditor` with `resolveUrl`, `checkPageReloadRequired`, and `sanitizeHtml` handlers — these need to be re-validated against the updated API. The migration should also revisit the `ComponentPath` identity workaround and verify event bus flows, component reload behavior, and fragment sanitization still work end-to-end.

#### Rationale

Staying on `0.0.6` blocks downstream modernization work tracked in the Page Editor modernization epic. The new package version is where the v6-aligned APIs land, so the migration is a prerequisite for the rest of the initiative.

#### Implementation Notes

- Bump the dependency in both `modules/app/package.json` and `modules/lib/package.json`.
- Re-check `page-editor.ts` handlers against the new type signatures.
- Validate `page-viewer.ts` still works with the updated package.
- Run `pnpm -C ./modules/lib run check` and a full `./gradlew deploy -x test -Penv=dev` smoke test.

### LINKED ISSUE #10292: PageComponentsView: tree selection does not select component in Page Editor
When a node is selected in the components tree (`PageComponentsView`), `inspectItem(path)` is called and `PageNavigationMediator` notifies an `INSPECT` event, but the selection is not reflected in the Page Editor iframe — the matching component on the canvas is not highlighted. Selecting from the `InspectPanel` works correctly, so the issue is in how the tree dispatches the navigation event (or what the Page Editor listens for from the tree path).

#### Rationale

Tree-driven navigation is a primary way to locate components on complex pages. Without the visual selection mirrored on the canvas, users lose context of which element they're editing.

#### References

- `modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsView.tsx`
- `modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/InspectPanel.tsx`

### LINKED ISSUE #10293: InspectPanel: switching to Inspect after adding a new component only works once
After inserting a new component in the Page Editor, the first navigation to the Inspect panel works as expected (if nothing was selected before). Inserting a second component before the selected one does not switch the panel. This breaks the typical insert-and-configure flow when authoring multiple components in sequence.

#### Steps to reproduce

1. Open a page in Page Editor.
2. Insert a new component → Inspect opens correctly.
3. Insert another component (placing it before the selected component) → Selected index remain unchanged → Inspect does not switch.

#### Rationale

Authors regularly add multiple components in a row; Inspect must respond on every insertion.

#### References

- `modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/InspectPanel.tsx`

### LINKED ISSUE #10295: InspectPanel: Apply and descriptor changes reload the page instead of hot-swapping the component
When the user changes a descriptor or presses Apply in the `InspectPanel`, the entire page is reloaded in the Page Editor iframe. The selected component loses its visual selection and any UI state, and the user is briefly disoriented while the full page rebuilds. Instead, only the affected component should be re-rendered (or hot-swapped) in place, preserving the current selection and component state. The previous Page Editor implementation already supported this targeted update flow.

#### Rationale

Full reloads break editing flow, hide which component the user is working on, and degrade perceived performance. Targeted updates keep the user oriented and make the editor feel responsive.

#### References

- `modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/InspectPanel.tsx`

### LINKED ISSUE #10296: InspectPanel: text component editing — auto-focus HtmlArea on select and restore Apply button
Two text-component-specific issues in `InspectPanel`:

1. When a text component is selected, the panel opens but the inline `HtmlArea` is not focused. Focus should move into the editor automatically so the user can start typing immediately.
2. The Apply button is missing for the text component editor. While inline edits are reflected in the iframe instantly, an Apply button is still required to persist the change to the saved content (mirroring the previous behavior). Apply must save without a full iframe reload — same behavior as in the related descriptor/Apply issue.

#### Rationale

Text editing is one of the most frequent actions in Page Editor. Missing focus and persistence controls force extra clicks and break parity with the legacy editor.

#### References

- `modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/InspectPanel.tsx`
- `modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/text/`

### LINKED ISSUE #10297: PageComponentsView: text components should display their content instead of default "Text" label
Text components in the `PageComponentsView` tree always render the static label "Text" regardless of the component's content. The displayed name should be a snippet of the actual text content; the "Text" fallback should appear only when the component is empty. This matches the legacy page components view behavior.

#### Rationale

Authors rely on the tree to identify components quickly. A column of identical "Text" labels makes pages with multiple text components unnavigable.

#### References

- `modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsItem.tsx`

edloidas:
- Added support for the new Page Editor dependency format.
- Made InserPanel elements draggable and synced them with new page editor.
- Added Apply button to InspectPanel.

---

_This branch depends on a local relative npm-page-editor dependency, while new version is not released yet._


pmi:
Depends on https://github.com/enonic/npm-page-editor/pull/29

pmi (changes requested):
Wrote a review in [npm-page-editor repo](https://github.com/enonic/npm-page-editor/pull/29#pullrequestreview-4156623231).

In a nutshell, this goes against the idea of UI agnostic event based handling.

modules/app/src/main/resources/assets/js/page-editor.ts
```diff
@@ -1,13 +1,68 @@
...
+import {UriHelper} from '@enonic/lib-contentstudio/app/rendering/UriHelper';
+import {PageHelper} from '@enonic/lib-contentstudio/app/util/PageHelper';
+import {PageState} from '@enonic/lib-contentstudio/app/wizard/page/PageState';
+import {ComponentPath as EditorComponentPath, EditorEvent, EditorEvents, PageEditor} from '@enonic/page-editor';
+import DOMPurify from 'dompurify';
 
 PageEditor.init(true);
-// console.info('Page editor started in edit mode.');
 
-const eventHandler = new EditorEventHandler();
```
pmi, line 8: Why did you need to delete the class and put the code in place ?

modules/app/src/main/resources/assets/js/page-editor.ts
```diff
@@ -1,13 +1,68 @@
...
+    try {
+        const response = await fetch(resolveComponentUrl(path));
+
+        if (!isExisting && needsPageReload(path, response.headers)) {
+            PageEditor.reloadPage();
+            return;
+        }
+
+        const html = sanitizeComponentHtml(await response.text(), path);
+        PageEditor.renderComponent(path, html);
```
pmi, line 35: That is a nice idea to reduce amount of PE exports, but we talked that PE should not know anything about rendering. Different UI frameworks will do it differently. It is only site engine who will use this approach.

modules/app/src/main/resources/assets/js/EditorEventHandler.ts
```diff
@@ -1,127 +0,0 @@
...
-        const newElement: Element = this.wrapLoadedComponentHtml(htmlAsString, componentView.getType());
-        const parentView = componentView.getParentItemView();
-
-        const createViewConfig: CreateItemViewConfig<RegionView> = new CreateItemViewConfig<RegionView>()
-            .setLiveEditParams(parentView.getLiveEditParams())
-            .setParentView(parentView)
-            .setPositionIndex(componentView.getPath().getPath() as number)
-            .setElement(newElement);
-
-        const newComponentView = componentView.createView(componentView.getType(), createViewConfig) as ComponentView;
```
pmi, line 104: What was the reason you needed to move rendering from here back into PE ? This is a completely different task from migrating PE into enonic UI.

pmi (changes requested):

modules/app/src/main/resources/assets/js/page-editor.ts
edloidas: It's a bit of a tricky part. But let me explain it here for all the related parts in both pulls.

I believe most of the confusion actually comes from the function names and by just renaming them would conclude it.

As you may notice, the loading flow is the same. But here's what the old code did differently:
- It was [receiving](https://github.com/enonic/app-contentstudio/blob/master/modules/app/src/main/resources/assets/js/page-editor.ts#L11) the `view` from the `ComponentLoadRequest` Editor event (not just the `path`).
  - The view itself, as you know, is an object that mixes rendering and state, and exposes methods to manipulate both.
  - We no longer use views to render components on the page, and most of their state manipulation and utility methods are either gone or obsolete.
- Then it was manipulating that view object to render what it needed:
  - First, the loading state — by asking view to show the [spinner](https://github.com/enonic/app-contentstudio/blob/master/modules/app/src/main/resources/assets/js/EditorEventHandler.ts#L58).
  - Then, the success state — by asking the same view to [create](https://github.com/enonic/app-contentstudio/blob/master/modules/app/src/main/resources/assets/js/EditorEventHandler.ts#L104)/render a new instance with a different configuration, and then asking it to [replace](https://github.com/enonic/app-contentstudio/blob/master/modules/app/src/main/resources/assets/js/EditorEventHandler.ts#L106C9-L106C52) the old view in the DOM.

But now, with the separation of rendered components and state in the Page Editor, views are gone and we only get a path here.

So previously we were receiving a "wrapper" around the path that held state and knew everything about DOM manipulation, rendering new views, and sending events on changes. We could tell it: _"hey, you handle this path"_, so _"mark yourself as loading"_ or _"update your HTML."_

Now we're doing roughly the same thing, but instead we ask the `PageEditor` class: "whatever is at that path, I need it marked as loading" or "I need this HTML at that path." The functions could actually be renamed to better reflect this: `renderComponent` → `replaceHtml`, `renderLoadingComponent` → `markLoading`/`showLoading`, `renderErrorComponent` → `markError`/`showError` (which could've lived in PE, but I kept them here for the same reason: to avoid rendering anything directly).

I could have made it even closer to the original by returning a new view as an object with the path in its state and those 3 manipulation functions. Or I could have hacked the legacy views into calling the new functions internally — but that would have just been uglier and more complicated.

modules/app/src/main/resources/assets/js/EditorEventHandler.ts
edloidas: Explained above.

modules/app/src/main/resources/assets/js/page-editor.ts
pmi: This is a good change code-wise, but not so good concept-wise.

I repeat for the 3rd time that PE should NOT render anything. No matter the name of the function. 
It used to do it when it was site engine specific, but we're trying to make it agnostic now: throws event and forgets about it.

pmi (approved):

pmi:
We need to move with the release, so let's merge it and adress issues later.

modules/app/src/main/resources/assets/js/page-editor.ts
edloidas: So, regarding this intermediate release:
- We merge this and PE.
- We come back after beta to continue with the changes.
Right?

---

As for the rendering part:

This is kind of a compromise to mirror the old behaviour, where the view was doing the rendering — except that we only have a path now. Same shape, new ways to update the path. Not the only temporary compromise we have, either: we should be adding a lib entry to PE to provide shared components (context menu, drag, styles) that we currently duplicate in CS, but we just don't have time to do it properly.

I think we might need to sit down for some pair programming or whiteboarding. Still, I'll share some thoughts so we can come back to them later — no need to reply now.

---

Correct me if I'm wrong — what you want to see is the current page-editor side injecting its own HTML without any help from PE?

I see quite a few options with different levels of complexity. For example — do you think PE still handles the rendering internally (like with regions, layouts, drop zones and other placeholders), but all communication goes through events instead of function calls? Or do we actually move the UI part over to CS?

For the second, the bare minimum we'd need is a function that points to the correct `HTMLElement` on the page, so CS can inject its UI into it. Having those shared components and CSS I mentioned before would be very helpful here. Maybe some utilities — pure functions to correctly create and update shadow roots — would also be nice. But that would also require moving some placeholders to CS and mimicking some functionality: creating shadow roots, placing components there, and compiling CSS for that.

PR #10285, opened 2026-04-22
Project selection dialog appears after a project is deleted #10272

### LINKED ISSUE #10272: Project selection dialog appears after a project is deleted
After deleting a project Content Studio for some reason opens the modal dialog for selecting a project.

skjulestad:
Reimplement legacy logic, deleting current project redirects to the parent project then first project in project lists. If now other projects open create project dialog.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/utils/cms/projects/projectSelection.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/utils/cms/projects/projectSelection.ts
@@ -1,0 +1,27 @@
+import {type Project} from '../../../../../app/settings/data/project/Project';
+
+type ProjectLike = Pick<Project, 'getName' | 'getParents'>;
+
+export function resolveActiveProjectId(
+    projects: Readonly<ProjectLike>[],
+    projectIdFromUrl: string | undefined
+): string | undefined {
+    if (projects.length === 1) {
+        return projects[0].getName();
+    }
+
+    return projects.some((project) => project.getName() === projectIdFromUrl) ? projectIdFromUrl : undefined;
+}
+
+export function resolveActiveProjectIdAfterDeletion(
+    remainingProjects: Readonly<ProjectLike>[],
+    deletedProject: ProjectLike | undefined
+): string | undefined {
+    const remainingByName = new Map(remainingProjects.map((project) => [project.getName(), project]));
+
+    const parentProject = (deletedProject?.getParents() ?? [])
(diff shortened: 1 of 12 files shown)


ashklianko (changes requested):
Important moment: not all projects are available to the user: we built entire project tree down from the project up to the top root project, including the projects that are not available to the user. 

So we can't switch to the first project in the list, we must switch to the first _available_ project, and if no projects available show corresponding message.

ashklianko:
Try by creating a user that has access only to a single project deep in the project hierarchyt

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/projects/projectSelection.ts
```diff
@@ -0,0 +1,99 @@
+import {type Project} from '../../../../../app/settings/data/project/Project';
+
+type ProjectLike = Pick<Project, 'getName' | 'getParents'> & Partial<Pick<Project, 'getDisplayName'>>;
+
+function isAvailableProject(project: ProjectLike): boolean {
+    return typeof project.getDisplayName !== 'function' || !!project.getDisplayName();
+}
```
edloidas: Why do you need `ProjectLike` type and why is it in such form?
And why do you need to have a functions that will check for optional `getDisplayName` function and try to get a name out of the object?

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/utils/cms/projects/projectSelection.ts
skjulestad: My mistake, first version was too generic.

The idea with `ProjectLike` was to keep this helper typed by the few project methods it uses, instead of depending on the full `Project` object.

But making `getDisplayName` optional was wrong. That came from keeping old lightweight test mocks working, and it made the helper assume that a project without `getDisplayName()` is available.

For this bug, that is not correct. Project availability is based on display name, so the helper should require `getDisplayName()` and check it directly.

I changed it to:

```
type ProjectLike = Pick<Project, 'getName' | 'getParents' | 'getDisplayName'>;

function isAvailableProject(project: ProjectLike): boolean {
    return !!project.getDisplayName();
}
```

Tests now include getDisplayName() as well, with a case where parent has empty display name and gets skipped. So current version matches the actual rule: no display name means project unavailable.

edloidas (approved):

PR #10289, opened 2026-04-22
New content dialog: double 'createMedia' request is sent #10244

### LINKED ISSUE #10137: Exception Occurs when using drag and drop in New Content Dialog
When using drag and drop (DnD) in the New Content modal dialog, the same content is attempted to be added twice.

Image

**However, this issue does not occur when adding media content via file upload with a specified file path** — in that case, the content is added only once as expected.

Image

### LINKED ISSUE #10244: New content dialog: double 'createMedia' request is sent
Try uploading an image via NewContentDialog: two requests are sent. 
Can be reproduced on Nightly.

reisfmb:
## Problem:
Dropping a media file on the New Content dialog fired two createMedia requests. The server rejected the second as a duplicate (500), while the first succeeded (200).

## Cause: 
Two drop handlers were both reacting to the same drop:
  1. DropZone (inside the Media tab) — calls uploadMediaFiles via onFiles.
  2. NewContentDialog's Dialog.Content — outer onDrop that also calls uploadMediaFiles.
  
The drop event ran on DropZone and then bubbled up to the dialog, so the same files were uploaded twice.

## Fix:
  - DropZone.handleDrop now calls e.stopPropagation(), so when the inner zone handles the drop, the outer dialog handler does not run again.
  - As a small cleanup, isDragging was moved out of local useState in NewContentDialog into the newContentDialog store.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
@@ -6,6 +6,7 @@ import {useI18n} from '../../../hooks/useI18n';
 import {
     $newContentDialog,
     closeNewContentDialog,
+    setIsDragging,
     uploadMediaFiles,
 } from '../../../store/dialogs/newContentDialog.store';
 import {DropZone} from '../../DropZone';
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
```diff
@@ -14,17 +15,17 @@ const NEW_CONTENT_DIALOG_MEDIA_TAB_NAME = 'NewContentDialogMediaTab';
...
     tabName: string;
-    isDragging?: boolean;
 };
 
 export const NewContentDialogMediaTab = ({
     tabName,
-    isDragging = false,
 }: NewContentDialogMediaTabProps): ReactElement => {
-    const {parentContent} = useStore($newContentDialog);
+    const {parentContent, isDragging} = useStore($newContentDialog);
```
edloidas: I'd put `isDragging` into it's own atom store. Or pass keys you are using here to the useStore hook.

edloidas: First option look more performant and optimal.

edloidas (changes requested):

ashklianko (approved):

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/new-content/NewContentDialogMediaTab.tsx
reisfmb: Yes, but the gain to create its own atom store is minimal, and adds a new file instead of a new line on the existing store... I'd go with the second one: "pass keys you are using here to the useStore hook".

edloidas: I meant you can keep it the same store file. We do a poor work on separation of entities in some cases and put too many props into the maps for no reason. But passing keys works too.

edloidas (approved):

PR #10290, opened 2026-04-22
Selector inputs: add error borders #10219

### LINKED ISSUE #10219: Selector inputs: add error borders
If there are errors on the self managing occurrences selectors input types, we should add error borders.

reisfmb:


modules/lib/src/main/resources/assets/js/v6/features/hooks/useSelectorInputHasError.ts
```diff
@@ -0,0 +1,13 @@
+import {Occurrences} from '@enonic/lib-admin-ui/form/Occurrences';
+import {getOccurrenceErrorMessage, useI18n, useValidationVisibility} from '@enonic/lib-admin-ui/form2';
+import {OccurrenceValidationState} from '@enonic/lib-admin-ui/form2/descriptor';
+
+export const useSelectorInputHasError = (occurrences: Occurrences, errors: OccurrenceValidationState[]) => {
```
edloidas: It's a very fragile hook. It requires validation visibility context. That's why we should place it into all hooks (I know, the reorganization is long overdue), so put it into form/input-types/hooks . And useSelectorInput as well.

modules/lib/src/main/resources/assets/js/v6/features/hooks/useSelectorInputHasError.ts
```diff
@@ -0,0 +1,13 @@
...
+    const t = useI18n();
+    const visibility = useValidationVisibility();
+    const occurrenceError = getOccurrenceErrorMessage(occurrences, errors, t);
+
+    if (visibility !== 'all') return false;
+
+    return occurrenceError != null || errors.some((error) => error.breaksRequired || error.validationResults.length > 0);
```
edloidas: Most of the code can be replaced with visibility check + new [hasOccurrenceError](https://github.com/enonic/lib-admin-ui/blob/master/src/main/resources/assets/admin/common/js/form2/utils/validation.ts#L57) function from lib-admin-ui to avoid passing and doing meaningless translation there.

edloidas (changes requested):

edloidas:
@claude fix failing tests

edloidas:
You are not Claude, Bruno. But looks like that integration is off and `@claude[agent]` should be used.

edloidas (approved):

ashklianko (approved):
Good job!

PR #10300, opened 2026-04-24
Fixing a positioning issue with datepicker trigger icon

ashklianko:


edloidas (commented):
Looks sketchy. Why can't it be done with Trigger alone? Should it be fixed in Enonic UI maybe?

ashklianko:
that's to make a nice focus look, so border rings not overlap:

image

edloidas (approved):

PR #10308, opened 2026-04-25
Error when saving content with inputs #10254

### LINKED ISSUE #10254: Error when saving content with inputs
1. Open a content with long, double, dateTime inputs
2. Enter a value then click on Save 

**BUG**  - Error when saving content with inputs

Image

```
Uncaught Error: Expected type of added Property to be [Long], got: Long
    at assert (main.js:6:458527)
    at PropertyArray.addProperty (main.js:1:71816)
    at main.js:1:74082
    at Array.forEach (<anonymous>)
    at PropertyArray.copy (main.js:1:74067)
    at main.js:1:86368
    at ObjectHelper.objectPropertyIterator (main.js:1:17265)
    at PropertySet.copy (main.js:1:86301)
    at new PropertyTree (main.js:3:9944)
    at ContentWizardPanel.assembleViewedContent (:8080/admin/com.enonic.app.contentstudio/main/_/asset/com.enonic.app.contentstudio:0000004a16a6a580/951.js:1:293389)
```

ashklianko:
- ref


ashklianko:
## What was done

Net: ~1500 lines removed from legacy `ContentWizardPanel.ts`, ~50 lines added in the new service.

Stripped legacy wizard plumbing out of `ContentWizardPanel` / `ContentWizardActions` and replaced it with the v6 store/service equivalents:

- **Mixins fetching** — moved out of `ContentWizardPanel` into a new `wizardMixins.service.ts`. It listens for `setPersistedContent` (via the new `onWizardPersistedContentSet` hook in `wizardContent.store`), fetches descriptors, and writes them to the store. A token guards against stale fetches racing each other.
- **Unsaved-changes / content-state tracking** — `ContentWizardActions` no longer listens to legacy panel events (`onDataChanged`, `onContentNamed`). It subscribes directly to `$wizardHasChanges` and `$wizardContentState`, and `hasUnsavedChanges()` reads from the v6 store.
- **Rename / content-path click** — the dialog flow moved into `ContentWizardToolbar` itself. It calls `openRenameContentDialog`, then `setDraftName` + `saveAction.execute()`. The legacy `onContentPathClick` / `compareVersionsPreHook` props and the `ContentNamedEvent` class are gone.
- **App-removed mixin cleanup** — draft mixins are now filtered by the app key encoded in the mixin name, not by descriptor lookup. Descriptors can be missing for apps added after the wizard loaded, which previously left orphan mixins in the draft.
- Removed the now-unused `isPathAvailable` flag from the toolbar store; nothing in the new flow sets it.

## Why

The legacy panel kept its own copies of state — unsaved-changes flags, mixin descriptors, persisted content — and fired its own events. The v6 stores were the new source of truth, but the legacy path was still writing to and reading from the old state in parallel. That dual source of truth is what caused #10254: on save legacy 'hasUnsavedChanges' method was doing something wrong with the content's properties

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1150,21 +752,6 @@ export class ContentWizardPanel
...
-
-    private handleSiteConfigApply() {
-        const siteConfigApplyHandler = (event: ContentRequiresSaveEvent) => {
-            if (this.isCurrentContentId(event.getContentId()) && this.hasUnsavedChanges()) {
-                this.setMarkedAsReady(false);
-                this.saveChanges();
-            }
-        };
-
-        ContentRequiresSaveEvent.on(siteConfigApplyHandler);
```
edloidas, line 1165: Since this is gone, we have no listeners for this event. But there are a lot of places it's fired. We should either remove them, or have a listener somewhere else, if it's still needed.

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -1968,233 +1297,16 @@ export class ContentWizardPanel
...
-
-            return this.layoutWizardStepForms(content).then(() => {
-                if (this.params.localized) {
-                    this.onRendered(() => {
-                        NotifyManager.get().showFeedback(i18n('notify.content.localized'));
-                        this.renderAndOpenTranslatorDialog();
-                    });
-                }
-
-                AI.get().setDataTree(this.contentWizardStepForm.getData());
```
edloidas, line 2018: Have you checked with AI? We no longer do updates on layout. Might not be current concern or might work anyway, but still.

modules/lib/src/main/resources/assets/js/app/browse/ContentWizardToolbar.ts
```diff
@@ -359,7 +360,38 @@ class ContentWizardToolbarElement extends V6ContentWizardToolbarElement {
...
+            this.config.onContentRenameConfirmed?.(newName);
+
+            const saveAction = this.config.actions.getSaveAction();
+            saveAction.setEnabled(true);
+            saveAction.execute();
+        });
+    }
+
+    private normalizeContentPathName(pathName: string): string {
+        return pathName?.startsWith(ContentUnnamed.UNNAMED_PREFIX) ? '' : pathName;
```
edloidas: Should probably be a utility function under utils/

modules/lib/src/main/resources/assets/js/v6/features/services/wizardMixins.service.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    unsubscribe = onWizardPersistedContentSet((content: Content) => {
+        void loadDescriptors(content.getContentId());
+    });
+}
+
+export function cleanupWizardMixinsService(): void {
+    unsubscribe?.();
+    unsubscribe = null;
+    pendingToken += 1;
```
edloidas, line 53: Better just reset `pendingToken = 0`, but not really an issue.

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
```diff
@@ -249,61 +172,34 @@ export class ContentWizardPanel
...
-
-    private applicationAddedListener: (applicationConfig: ApplicationConfig) => void;
-
-    private applicationRemovedListener: (event: ApplicationRemovedEvent) => void;
-
-    private applicationUninstalledListener: (event: ApplicationEvent) => void;
-
-    private applicationStoppedListener: (event: ApplicationEvent) => void;
-
-    private applicationStartedListener: (event: ApplicationEvent) => void;
```
edloidas, line 266: Will we still reload wizard when app uninstalled?

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
ashklianko: Bringing it back to avoid potential bugs.

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
ashklianko: AI interaction need to be re-implemented in coherence with our latest changes/stores. Should not be handled here at all together with the other AI interactions that were preserved for future reference

modules/lib/src/main/resources/assets/js/app/wizard/ContentWizardPanel.ts
ashklianko: We'll bring it back if needed, there are plenty of actions that reload the editor, plus we have a reload button on the preview toolbar.

PR #10311, opened 2026-04-26
Replace legacy Page Components View with the new one #10309

### LINKED ISSUE #10309: Replace legacy Page Components View with the new one
Detached PCV needs a touch too:

Image

ashklianko:


modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsView.tsx
```diff
@@ -160,7 +164,7 @@ export const PageComponentsView = (): ReactElement => {
 
     return (
         <div data-component={PAGE_COMPONENTS_VIEW_NAME} className="flex flex-col gap-1 py-2">
-            <h3 className="text-base font-semibold">{i18n('field.components')}</h3>
+            {showTitle && <h3 className="text-base font-semibold">{i18n('field.components')}</h3>}
```
edloidas: Legacy i18n still used here. Please, use hook and move to the top.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/layout/DetachedPageComponentsView.tsx
```diff
@@ -0,0 +1,179 @@
...
+};
+
+const writeCollapsed = (value: boolean): void => {
+    if (typeof window === 'undefined') return;
+    if (value) {
+        window.localStorage.setItem(COLLAPSED_STORAGE_KEY, 'true');
+    } else {
+        window.localStorage.removeItem(COLLAPSED_STORAGE_KEY);
+    }
+};
```
edloidas: `app.store.ts` is our global state store synced with local storage. Use it instead of manual local storage management.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsView.tsx
```diff
@@ -35,7 +35,11 @@ const PAGE_COMPONENTS_VIEW_NAME = 'PageComponentsView';
 // ? tree is rebuilt with changed positional keys.
 const animateLayoutChanges = ({isSorting}: {isSorting: boolean}): boolean => isSorting;
 
-export const PageComponentsView = (): ReactElement => {
+export type PageComponentsViewProps = {
+    showTitle?: boolean;
+};
+
+export const PageComponentsView = ({showTitle = true}: PageComponentsViewProps = {}): ReactElement => {
```
edloidas: Maybe keep it false and just use `<PageComponentsView showTitle>`? Doesn't look that we need to show it as a most common option, since we only use it in two places.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/layout/DetachedPageComponentsView.tsx
```diff
@@ -0,0 +1,179 @@
...
+            writeCollapsed(next);
+            return next;
+        });
+    }, []);
+
+    const handlePointerDown = useCallback((event: ReactPointerEvent<HTMLDivElement>): void => {
+        if (event.button !== 0) return;
+        // ? Skip drag start when the press lands on an interactive child (e.g. close button),
+        // ? otherwise pointer capture on the header swallows the click.
+        if ((event.target as Element).closest('button, [role="button"]') != null) return;
```
edloidas: Better avoid `as`, we'll ban it. Something like this with `instanceof` might do:
```ts
const isIntereactive = target instanceof Element && target.closest('button, [role="button"]') != null
if (isIntereactive) return;
```

edloidas (changes requested):

edloidas (approved):
Looks good. Though the component itself still need some more work. Let's discuss it.

PR #10335, opened 2026-04-28
Make 'language' field editorial #10331

### LINKED ISSUE #10331: Make 'language' field editorial
This is a CS part of https://github.com/enonic/xp/issues/11975

To be done:

- Update ContentResource and related classes to bring 'language' back from metadata update to regular content update
- In Edit Settings dialog: if language gets updated send an update content request
- When making Restore operation don't include 'owner' fields
- When doing localisation always do empty 'updateMetadata' call to remove required inheritance fields. If this layer's language differs from the parent's language do a separate 'updateContent' call with the language.

ashklianko:


ashklianko:
Requires https://github.com/enonic/xp/pull/11977

PR #10338, opened 2026-04-28
enonic/xp#11975

edloidas:


rymsha:
WIP @ashklianko

ashklianko:
being fixed in https://github.com/enonic/app-contentstudio/pull/10335

PR #10340, opened 2026-04-28
Missing separator #10339

### LINKED ISSUE #10339: Missing separator
For SplitActionButton.tsx we need to add separator between actions groups to clearly state actions are coming from different sources.

Image

skjulestad:
For SplitActionButton.tsx we need to add separator between actions groups to clearly state actions are coming from different sources.


edloidas (changes requested):
Too complicated and mixes presentation (separator) with semantics (actions). And you even need to invent ID for it.
I suggest we stick with `Action[][]`

edloidas (approved):

edloidas:
Nice one!

PR #10358, opened 2026-04-29
Import content widget #10356

### LINKED ISSUE #10356: Import content widget
widget that import note exports as it is content
- removes publish data
- adds `content.import` action and prevents vacuuming

rymsha:


ashklianko:
Things to do/discuss:

1. Add handling on a new operation type 'content.import'
2. enable/disable export depending on num of items selected (only enable when 1 item is selectef)
3. Need to consider: root/all root items export/import
4. Need to consider: Export params - workflow/parent project/publish fields (now all the publish fields are erased)
5. Fix failing ui test

rymsha:
- No root import/export
- show popup with destination on export
- multiple export and import files if multiple selected
- prefix exports with `<insance-name>-<project-id>...`
- option to keep `publish.firtst` (default)
- move Unknown to "Show all activities"

ashklianko:
This widget is ready for usage, however it is not decided yet where should it be located: in CS or CS+. Additionally the previous import widget needs to be removed simultaneously from Data Toolbox app.

rymsha:
It is decided - it is part of CS

alansemenov (approved):
I'll merge this but the new widget is missing two important functions that DT's widget has: deletion of imports and especially download of exports. Both can still be done inside the DT, so it's ok for now. But we should consider adding both to the new widget.

PR #10382, opened 2026-04-30
Inspect panel Apply button: mirror Save action enable state #10381

### LINKED ISSUE #10381: Inspect panel Apply button: mirror Save action enable state
The Apply button in the page editor's Inspect panel maintained its own dirty and validity tracking that diverged from the wizard's Save action. As a result, after editing a form value and reverting it to the original, Apply stayed enabled because the dirty flag was never cleared. Apply was also disabled while inline validation errors were present, even though those errors are purely cosmetic and Save itself does not gate on them. The button now mirrors the wizard Save action's enabled state via `Action.onPropertyChanged`, dropping the duplicated dirty/valid bookkeeping in `inspect-panel.store`, `useInspectFormTracking`, `useInspectTextTracking`, and `TextEditor`. This fixes the revert case for free, since the wizard already computes `$wizardHasChanges` from a structural diff of persisted vs draft state.

#### Rationale

The Save action is the authoritative source for "are there unsaved changes that can be persisted?". Duplicating that logic in the inspect panel was both incorrect (no baseline comparison on revert) and over-restrictive (gated on validation). Mirroring Save unifies the two surfaces and keeps Apply consistent with the toolbar.

edloidas:
Replaced the inspect panel's bespoke dirty/valid bookkeeping with a reactive subscription to the wizard `saveAction.onPropertyChanged`. Apply now enables exactly when Save does — fixes the case where reverting a form value to its original left Apply enabled, and removes the validation gate so inline errors no longer block Apply.

- Dropped `$inspectFormDirty` / `$inspectFormValid` and their setters from `inspect-panel.store`.
- Simplified `useInspectFormTracking` and `useInspectTextTracking` to track only form presence.
- Removed baseline-vs-current dirty tracking from `TextEditor`.
- Removed the redundant `setInspectFormDirty(false)` from `ApplyButton` after `requestReloadComponent`.

Tracking Save button was chosen to have a simplified implementation, since  Save action's enabled state is derived from four inputs (`$wizardHasChanges`, `persistedContent.isEditable()`, `this.userCanModify`, `!persistedContent.isDataInherited()`), and only 1 is a store. Should be updated once all our states will be in stores.

Closes #10381

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/store/inspect-panel.store.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/store/inspect-panel.store.ts
@@ -9,9 +9,7 @@ export const $inspectSaveAction = atom<Action | null>(null);
 
 export const $inspectFormPresent = atom<boolean>(false);
 
-export const $inspectFormDirty = atom<boolean>(false);
-
-export const $inspectFormValid = atom<boolean>(true);
+const $inspectSaveActionEnabled = atom<boolean>(false);
 
 //
 // * Computed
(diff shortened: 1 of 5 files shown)


ashklianko (changes requested):
Noticed a problem with non empty part config: after changing it's value Apply button doesn't get disabled anymore 

Image

ashklianko:
Same on Nightly, so the PR is not issue's origin but related

edloidas:
This actually should be fixed. It's indeed a problem on Nightly, but this fix addresses this issue as well.
I double-checked this. It might be that your cache was not updated and you were checking the old version.

PR #10390, opened 2026-05-04
config.store: use explicit initConfig(scriptId) #10389

### LINKED ISSUE #10389: config.store: use explicit initConfig(scriptId)
The v6 config store populated `$config` at module load by reading `document.currentScript.getAttribute('data-config-script-id')` and looking up the JSON element by that id. This pattern has two problems that block reuse outside the Content Studio main bundle:

1. `document.currentScript` is `null` when the loader script is `type="module"` — modules are not classic scripts and there is no "currently executing script" reference. Any consumer that wants to ship as an ES module bundle hits this immediately.

2. The id is taken from a `data-*` attribute on the loader script. That forced the id to be discoverable from a single specific entry point. Other apps that want to reuse v6 stores cannot pass their own id — the resolution is hardcoded to "ask currentScript".

The store now exposes `initConfig(scriptId: string)`. Eager loading at module-load is removed; `$config` starts empty and is populated when the bootstrap explicitly calls `initConfig` with the id of its own inline JSON config script. Content Studio's own bootstraps in `main.ts` and `settings.ts` call `initConfig(configScriptId)` right after `CONFIG.setConfig(...)`, before any v6 code runs, so existing consumers see no behavioral change.

This unblocks two things:
- The v6 config store can be initialized from an ES-module bundle embedded as an extension or widget, without the `currentScript` workaround.
- A consumer app that ships its own JSON config under a different script id (e.g. `my-widget-config`) can call `initConfig('my-widget-config')` and reuse `$config` and everything derived from it.

`scriptElement.innerText` is replaced with `textContent` while the function is being rewritten — `innerText` triggers layout and respects CSS visibility, neither of which is meaningful inside a `<script>` element.

Note: the legacy lib-admin-ui `CONFIG` singleton (`CONFIG.setConfig`) is still populated separately by each bootstrap. That is intentional during the v6 migration — `CONFIG` is the full untyped object used by `app/` (legacy), `$config` is the typed subset used by `v6/`. Both read the same JSON; consolidation is out of scope for this change.

#### Rationale

Without this change, no ES-module-based extension can reuse `@enonic/lib-contentstudio/v6/features/store/config.store`, because the eager `loadConfig()` runs at import and silently fails when `currentScript` is null. With explicit `initConfig`, ownership of "where the config comes from" moves from the store to the bootstrap, which is the only place that knows the answer.

ashklianko:


ashklianko:
Required for https://github.com/enonic/app-contentstudio-plus/pull/1769

edloidas (approved):

PR #10391, opened 2026-05-04
Show the Project creation dialog when no projects are present #10353

### LINKED ISSUE #10353: Show Project creation dialog when no projects are present
[Only system.admin and cms.admin roles have permissions to create new projects.](https://developer.enonic.com/docs/xp/stable/api/lib-project#create)

When we open CS and no projects are present we need to immediately open project creation dialog (this is how it used to work previously).

Good to have, but can be made as a separate task after this one, requires refactoring:
- Load empty CS so we could use menus and switch to other apps. As an option - open Settings app and disable CS tab until a project created.

skjulestad:
Show get access or create a project based on admin rights, gate a new button in settings, show browser toolbar with user and menu button as well as sidebar with setting, title and default project icon. Replace deprecated ProjectContext where possible.


ashklianko (approved):
Good work, Mathias!
Only added few small improvements from my side.

PR #10392, opened 2026-05-04
Move attachment textContent to media data

### LINKED ISSUE #10409: Move attachment textContent to media data
CS part for https://github.com/enonic/xp/pull/12021

rymsha:
Follow-up to PR #10388. PR #10388 added `sha512` and `textContent` to `AttachmentJson` (Java + TS). The `sha512` half is correct — it's a binary-level concern. The `textContent` half is misplaced: it's the Tika-extracted text from a media binary, so it's meaningful only for media content types (PDFs, docs, …), not for every attachment. Non-media attachments (page banners, source attachments, …) always carried a null `textContent`, which is misleading both in the JSON preview and in any consumer code that reads it.

The XP follow-up at https://github.com/enonic/xp/pull/12021 moves `textContent` off `Attachment` and onto the media content's own data at `data.media.text`. This PR is the contentstudio-side correction.

Changes:

- `AttachmentJson.java` — drop `getTextContent()`. Keep `getSha512()`.
- `AttachmentJsonTest.java` — drop the `textContent` assertions; rename `nullSha512AndTextContent` to `nullSha512`. Keep the sha512 ones.
- `Attachment.ts` — drop the `textContent` field, `getTextContent()`, `setTextContent()`, the `fromJson` assignment, and the `toJson` output. Keep `sha512`.
- `AttachmentJson.ts` — drop the optional `textContent?` field.
- `ContentResource.java` — drop `.text(sourceAttachment.getTextContent())` in `createAttachment(...)`, since `CreateAttachment.text(...)` and `Attachment.getTextContent()` are removed upstream.

Surfacing `textContent` on the media-data view: no contentstudio code change needed. The content's `data` tree (including `data.media.text` once the XP PR lands) is already rendered as JSON in the inspect/preview panels — no place in contentstudio currently special-cases `data.media`.

Sequencing: this PR's TS surface tolerates `textContent` being absent from `AttachmentJson` (the field on the XP side is removed entirely; `AttachmentJson.ts` already had it as optional, so it's forward- and backward-compatible). The Java side compiles against the published XP today and will continue to compile once the new XP snapshot lands, since we're only removing calls.

References

- https://github.com/enonic/app-contentstudio/pull/10388 (the PR being amended)
- https://github.com/enonic/xp/pull/12021 (paired XP follow-up)

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/json/content/attachment/AttachmentJson.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/json/content/attachment/AttachmentJson.java
@@ -41,11 +41,6 @@ public class AttachmentJson
         return this.attachment.getSha512();
     }
 
-    public String getTextContent()
-    {
-        return this.attachment.getTextContent();
-    }
-
     @JsonIgnore
     public Attachment getAttachment()
     {
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentResource.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentResource.java
@@ -1755,7 +1755,6 @@ public final class ContentResource
             .name( attachmentName )
             .mimeType( mediaFile.getContentType().toString() )
             .byteSource( mediaFile.getBytes() )
-            .text( extractedData.getText() )
             .build();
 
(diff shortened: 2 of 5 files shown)


alansemenov:
This needs to be linked to an issue

ashklianko (approved):

PR #10408, opened 2026-05-05
Move Image Editor to preview panel #10175

### LINKED ISSUE #10175: Move Image Editor to preview panel
* Make the editor work by fixing https://github.com/enonic/app-contentstudio/issues/10182
* Remove special handling of the media preview where we don't open the preview panel when editing images
* When editing an image, show image editor over the preview panel (only when Automatic preview mode is selected). Preferably skip the request to preview widget in this case, as we won't show the preview.
* When switching from Automatic to any other preview mode, preview panel should be shown instead of the image editor
* In mobile resolution, when we don't show the preview, show image editor inlined in the form, as now

reisfmb:
Image editor now appears under the preview toolbar in Automatic preview mode for image content, instead of inside the form.

**Behavior**
- Auto mode + image content -> editor renders over the preview area; the preview-widget request is skipped;
- Switch to another widget -> editor unmounts, iframe preview loads normally;
- Mobile / preview hidden -> editor falls back to inline form rendering;
- Removed the legacy special handling of the media preview where we don't open the preview panel when editing images.

**Key Changes**
- New `LiveViewImageEditor` component mounted into the preview frame container;
- New `$isLiveViewImageEditorActive` (computed) gates the iframe visibility and the preview-widget request;
- New `$isPreviewPanelVisible` atom store, shared between wizard (ContentWizardPanel) and browse (ContentItemPreviewPanel), drives whether ContentForm excludes the ImageUploader input to avoid double rendering;
- FormRenderer accepts an optional `excludeInputTypes` prop;
- Image-uploader minor visual updates.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/image-uploader/ImageUploaderInputImage/ImageUploaderInputCropSvg.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/image-uploader/ImageUploaderInputImage/ImageUploaderInputCropSvg.tsx
@@ -69,13 +69,30 @@ export const ImageUploaderInputCropSvg = (): ReactElement => {
     return (
         <svg
             viewBox={`0 0 ${dimensions.w} ${dimensions.h}`}
-            className="w-full h-auto cursor-crosshair"
+            className="w-full h-full cursor-crosshair"
+            style={{maxWidth: dimensions?.w, maxHeight: dimensions?.h}}
             onClick={handleClick}
             onMouseMove={handleMouseMove}
         >
             <image href={base64Image} width={dimensions.w} height={dimensions.h} />
 
-            {!rect && <rect x={0} y={0} width={dimensions.w} height={dimensions.h} fill="black" fillOpacity={0.5} />}
+            {!rect && (
+                <>
+                    <rect x={0} y={0} width={dimensions.w} height={dimensions.h} fill="black" fillOpacity={0.5} />
+                    {/* TODO: remove this harcoded text when crop is behavior gets updated */}
+                    <text
+                        x={dimensions.w / 2}
+                        y={dimensions.h / 2}
+                        textAnchor="middle"
+                        dominantBaseline="middle"
+                        fill="white"
+                        fontSize={Math.min(dimensions.w, dimensions.h) * 0.04}
+                        style={{pointerEvents: 'none'}}
+                    >
+                        Click the image to draw cropping area
+                    </text>
+                </>
+            )}
 
             {rect && (
                 <>
@@ -84,7 +101,7 @@ export const ImageUploaderInputCropSvg = (): ReactElement => {
                         fillRule="evenodd"
                         fill="black"
                         strokeWidth={strokeWidth}
-                        fillOpacity={0.4}
+                        fillOpacity={0.5}
                     />
                     <rect
                         x={rect.x1}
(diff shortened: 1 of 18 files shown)


modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/image-uploader/ImageUploaderInputImage/ImageUploaderInputCropSvg.tsx
```diff
@@ -69,13 +69,30 @@ export const ImageUploaderInputCropSvg = (): ReactElement => {
...
+                    <text
+                        x={dimensions.w / 2}
+                        y={dimensions.h / 2}
+                        textAnchor="middle"
+                        dominantBaseline="middle"
+                        fill="white"
+                        fontSize={Math.min(dimensions.w, dimensions.h) * 0.04}
+                        style={{pointerEvents: 'none'}}
+                    >
+                        Click the image to draw cropping area
```
ashklianko, line 92: Replace hardcoded text with i18n

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -128,6 +128,8 @@ export const $wizardDataValidation = map<FormDataValidation>({});
 
 export const $isContentFormExpanded = atom<boolean>(true);
 
+export const $isPreviewPanelExpanded = atom<boolean>(true);
```
ashklianko: Not used?

ashklianko (changes requested):
Image

- Buttons on the right all have different heights! 
- Apply button must have the same background as 'Reset' and 'Close'
- Focus: when entering mode and immediately pressing Apply - focus is not set to the center. Initialize focus state when entering focus mode, or make Apply use the same default point that the SVG displays.

alansemenov:
> Apply button must have the same background as 'Reset' and 'Close'

Do we have a more fitting style for "Reset"? In the legacy version it used to be a link and looked very different from "Apply". We definitely don't want "Reset" and "Apply" to look the same.

ashklianko:
Not really, and we better stick to our defined styles:
image

alansemenov:
Something like this would be good for Reset - no background by default

image

modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/image-uploader/ImageUploaderInputImage/ImageUploaderInputCropSvg.tsx
reisfmb: This is a temporary improveemnt. We'll change how crop works later on, and remove this text, so Alan requested it to not be translated.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
reisfmb: Yeah, initially I added the atom in this store, but later on I migrated to its own store. I'll remove, good catch Sasha, thanks.

ashklianko:
A question: is it intended that we are triggering save immediately after pressing 'Reset'? It works same way on Nightly, but on the QA (Legacy) it does not save it. @alansemenov

alansemenov:
No, it shouldn't save but it's not related to this task. There are lots of problems with image editor, so let's fix those in separate tasks.

ashklianko (approved):

Issue #10397, opened 2026-05-04
Incorrect icon/name for Page node in the Page Component tree


alansemenov:
Both icon and name are currently hardcoded to globe and "Page", but it should show controller name that is used for rendering of the page.

<img width="1450" height="469" alt="Image" src="https://github.com/user-attachments/assets/5913049a-b63f-4360-928f-3ed12e13aa54" />

It works correctly in the legacy version:
<img width="678" height="278" alt="Image" src="https://github.com/user-attachments/assets/562b4893-6600-45c0-97ab-d847597e58e2" />

PR #10413, opened 2026-05-06
Display name won't auto-grow #10399

### LINKED ISSUE #10398: Fieldset border breaks in between the inputs
The vertical line on the left side shouldn't break:

Image

### LINKED ISSUE #10399: Display name won't auto-grow
Display name should wrap to another line and auto-grow vertically, as shown in [the design](https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?node-id=46788-185499&m=dev)

skjulestad:
Display-name grows upon editing but stays single line else, field-set border stays as a continues line between inputs, fixed name and icon for page node in page component tree.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -218,8 +218,8 @@ importers:
         specifier: ^10.0.0
         version: 10.0.0(@dnd-kit/core@6.3.1(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)
       '@enonic/ui':
-        specifier: ~0.50.0
-        version: 0.50.0(@radix-ui/react-slot@1.2.4(@types/react@19.2.14)(react@19.2.4))(focus-trap-react@11.0.6(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(lucide-react@1.14.0(react@19.2.4))(preact@10.29.1)(react-dom@19.2.4(react@19.2.4))(react-virtuoso@4.18.3(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)(tw-animate-css@1.4.0)
+        specifier: ~0.49.0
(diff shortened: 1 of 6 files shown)


edloidas:
This shouldn't be in a single PR. `Display name won't auto-grow #10399` and `Incorrect icon/name for Page node in the Page Component tree #10397` are too breaking, and I can't just approve the PR because of them, since they require changes. I see no reason why those 4 fixes should've been on one branch — they're pretty much separate things. Please don't do that in the future. If it's possible to commit them separately, please do so.

skjulestad:
> This shouldn't be in a single PR. `Display name won't auto-grow #10399` and `Incorrect icon/name for Page node in the Page Component tree #10397` are too breaking, and I can't just approve the PR because of them, since they require changes. I see no reason why those 4 fixes should've been on one branch — they're pretty much separate things. Please don't do that in the future. If it's possible to commit them separately, please do so.

I got asked to do this in one pr :') would have made single or per issue of else. Will note for future, scope may have been larger than assumed

edloidas:
No worries. But better split them, if they are not that simple. Or just ask me first, if it's me who is going to review it in the middle of the night. Alan may have just thought that the changes were trivial. But sine they are not, it's better to review them separately.

ashklianko:
I am on it, no need to review it in the middle of the night btw

ashklianko (commented):
1. On blur when DisplayName becomes a single line again we need to show truncate dots '...' if the name is long, otherwise it is not obvious that there's more of display name
2. Pressing 'Enter' in display name behaves wrong: it blurs the input. Compare with any text input where pressing Enter does nothing

ashklianko (changes requested):
1. On blur when DisplayName becomes a single line again we need to show truncate dots '...' if the name is long, otherwise it is not obvious that there's more of display name
2. Pressing 'Enter' in display name behaves wrong: it blurs the input. Compare with any text input where pressing Enter does nothing

ashklianko:
One more thing: let's leave EditableText as it is, now it becomes overcomplicated and hard to use. Do this specialised logic directly in DIsplayNameInput. We'll extract this logic to a separate component if needed.

ashklianko (approved):

PR #10421, opened 2026-05-06
Sort dialog items #10420

### LINKED ISSUE #10420: Sort dialog items
Sort dialog items inverse now results in white on white(unreadable).

skjulestad:
Sort dialog now opts out of inverse style.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/ItemLabel.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/ItemLabel.tsx
@@ -7,6 +7,7 @@ export type ItemLabelProps = {
     'icon': ReactNode;
     'primary': ReactNode;
     'secondary'?: ReactNode;
+    'inverseTone'?: boolean;
     'data-component'?: string;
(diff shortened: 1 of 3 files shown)


edloidas (commented):
The fix is correct (technically), but the implementation is not an optimal one. We do not need to control tone at all here. It is supposed to be an addition to parts, that reverse it's background. We don't do that in sort list item, we don't need to customize it, since it is used in sort dialog only, and it must not be forced by separate property just for the tone.

edloidas:
The fix is rather much simpler and would only require changes to the item.
`bg-unset` is set in `SortDialogMainContent` for `SortContentListItem` which makes any changes in background meaningless. I assume it was done to get rid of the bg change on selection and hover. This also makes all `selected`, `isMovable`, `isFocused` meaningless and all the changes of bg inside the item as well.
So we just unset background and don't mark it selected. That's it.

edloidas (approved):

ashklianko:
This fix was only treating symptoms, was going to write similar comments.

PR #10441, opened 2026-05-07
Invalid content ID #10216

### LINKED ISSUE #10216: Invalid content ID in edit URL shows errors instead of redirecting to browse view
When navigating to the content edit route with an invalid or non-existent content ID (e.g. `/features/edit/some-incorrect-id`), the application displays errors instead of gracefully falling back to the plain browse view (`/features/`). The expected behavior is to silently discard the invalid ID and load the browse view as if no edit route was specified.

This affects both the modern v6 code and the legacy code paths.

### LINKED ISSUE #10433: Broken context after switching project in the opener tab
1. Open content tree in any project in tab A
2. Open any content for edit (tab B)
3. Go back to tab A, and switch to another project
4. Go back to tab B and refresh the page

Image

ashklianko:
When the edit URL contains a broken content ID, the app now navigates to browse and shows a localized toast — without a full page reload.

`preLoadApplication` validates the wizard URL before the SPA mounts. On rejection it sets a module-scoped flag and calls `history.replaceState` to swap the URL in place to browse. `startApplication` then re-evaluates `isContentWizardUrl()` against the rewritten pathname and routes into `startContentBrowser`, which consumes the flag and shows the toast.

- No full page reload — `history.replaceState` instead of `window.location.replace`. Smooth transition, broken URL doesn't enter browser history.
- No sessionStorage bridge — a `boolean` flag survives the synchronous transition.
- Single routing decision point — `isContentWizardUrl()` against the live URL. The previous gating flag in `startApplication` is gone.
- Reuses `text.content.not.found` instead of adding a new phrase. No `Exception` parsing, no severity switch, no duplicated message extraction.
- `clearFavicon()` only on success — favicon stays untouched on failure instead of leaving a placeholder.
- Fixes a latent bug in `ContentWizardPanel.doLoadData` — `.finally` ran `initLiveEditModel(currentItem)` even after rejection, masking the original error with a secondary crash. Switched to `.then(() => loadedContent)` so rejections propagate cleanly.

Closes #10216

### DIFF
--- a/modules/app/src/main/resources/assets/js/main.ts
+++ b/modules/app/src/main/resources/assets/js/main.ts
@@ -171,6 +171,10 @@ const faviconCache: Record<string, HTMLElement> = {};
 
 const iconUrlResolver = new ContentIconUrlResolver();
 
+// ? dataPreloaded is set inside the preload .then so that startContentWizard can
+// ? skip favicon/title work that already happened. Safe because startApplication
+// ? is gated on preLoadPromise — by the time the wizard runs, the request has
+// ? resolved and the flag is correct.
 let dataPreloaded: boolean = false;
 
 let invalidEditUrlNotificationPending: boolean = false;
@@ -221,6 +225,9 @@ function preLoadApplication(): Promise<void> {
     const shouldPreloadTabData: boolean = !Body.get().isRendered() && !Body.get().isRendering();
 
     if (wizardParams.contentId) {
+        // ? Request fires regardless of body render state so background tabs with
+        // ? invalid bookmarks also get URL cleanup before the user focuses them.
+        // ? Favicon/title updates remain gated on shouldPreloadTabData.
         return Promise.resolve(
             new GetContentByIdRequest(wizardParams.contentId).setRequestProjectName(wizardParams.projectName).sendAndParse()
                 .then((content: Content) => {
@@ -237,6 +244,7 @@ function preLoadApplication(): Promise<void> {
                 .catch(() => {
                     invalidEditUrlNotificationPending = true;
                     history.replaceState(null, '', UrlHelper.createContentBrowseUrl(wizardParams.projectName));
+                    wizardParams = undefined;
                 })
         ).then(() => undefined);
     }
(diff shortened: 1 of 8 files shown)


ashklianko:
Going to fix https://github.com/enonic/app-contentstudio/issues/10433 here, partially related

ashklianko:
## Summary

Replaces the indiscriminate `localStorage` sync of `$projects.activeProjectId` with a dedicated `$defaultBrowseProjectId` atom that is only written on explicit user action, and rolls back the now-redundant `projectName` plumbing in the wizard data loader.

## Background

`projects.store.ts` previously synced the entire `activeProjectId` field via `syncMapStore`, persisting to `localStorage` on every change of any key in the `$projects` map.

Two problems:

1. `localStorage` is shared across tabs, but `activeProjectId` is per-tab state driven by the URL. Tabs silently overwrote each other's persisted project; an empty-URL tab on reload could land on a project chosen in a different tab.
2. The listener fired for changes to unrelated keys (`projects`, `loaded`, `resolved`, …), producing redundant `localStorage` writes that re-encoded the same `activeProjectId` over and over.

## What changed

### `projects.store.ts`

- Dropped `syncMapStore($projects, …)`. Added a separate `$defaultBrowseProjectId` atom synced via `syncAtomStore('defaultBrowseProjectId', { loadInitial: true })`. The atom is only updated by explicit `.set(…)` calls — no automatic mirroring from the map.
- `setActiveProject` is now internal. Added a public `selectProject(project)` command — the UI write path. It updates the active project and persists the default browse id. Programmatic flows (init, fallback resolve, URL-driven changes) keep using internal helpers and never write storage.
- Renamed `initializeActiveProjectFromUrl` → `initializeActiveProject`. New init logic:
  - URL has a project → use URL.
  - URL empty and not a wizard URL → fall back to `$defaultBrowseProjectId`.
  - Wizard URL → no storage fallback (wizards always carry an explicit project; falling back to a different one would be wrong).
- `updateActiveProjectAfterDeletion` and `removeProject` keep the persisted default consistent: if the deleted project equals `$defaultBrowseProjectId`, the value is updated to the resolved fallback or cleared. Deleting an unrelated project leaves storage untouched.

### `ProjectSelectionDialog.tsx`

Switched from `setActiveProject` to `selectProject` so user-driven selections in the dialog are persisted as the default browse project.

### Wizard data loader — revert of redundant `projectName` plumbing

With the store init now reliably driven by the URL, `ProjectContext` is correctly populated by the time `ContentWizardDataLoader` runs. The defensive `projectName` parameter plumbing introduced earlier becomes redundant inside the loader. Reverted in:

- `ContentWizardDataLoader.ts` — `loadContent`, `loadSite`, `loadDefaultModels`, `loadParentContent`, and the `loadDataForNew` / `loadDataForEdit` orchestration.
- `page/DefaultModelsFactory.ts` — `projectName?` removed from `DefaultModelsFactoryConfig` and the `setRequestProjectName(…)` call.
- `ContentWizardPanel.ts` — third argument removed from the `loadDefaultModels(…)` call site.

`ContentWizardPanelParams.projectName` is **kept**: `main.ts`'s `preLoadApplication` runs before the project store finishes initializing and still needs an explicit project name read from the URL.

## Why this design

- Persist only what is genuinely "the user's last UI choice", and treat it as a *fallback*, not a source of truth.
- URL stays authoritative; storage is consulted only when the URL has nothing to say.
- Storage writes happen only on explicit UI selection or deletion-driven fallback — no incidental writes from unrelated state.
- Wizard URLs deliberately skip the storage fallback to avoid silently switching to a different project than the URL implies.

PR #10449, opened 2026-05-08
Part descriptors dropdown is not shown if the app is missing #10411

### LINKED ISSUE #10411: Part descriptors dropdown is not shown if the app is missing
### With app:
**tutorial.nxp** app was used to create a page and it has a **heading** part.
As long as app is added to the site, part descriptor dropdown is present in contenxt window inspect panel.
Image

### Without app:
But as soon as app is deleted from the site, the descriptor is hidden, despite the fact that it is present in page json.
 Image

### Expected:
This should work same way it works for page templates/controllers in [PageTemplateAndControllerSelector.ts](https://github.com/enonic/app-contentstudio/blob/master/modules/lib/src/main/resources/assets/js/app/wizard/page/contextwindow/inspect/page/PageTemplateAndControllerSelector.ts):
- Part/Layout descriptor dropdown should always be visible if the descriptor is present in the page json. It should show that descriptor as an option in the list ( even though it's not present in the app ) with following fields:
  - title: <descriptor name>
  - subTitle: Invalid descriptor
- If there are no descriptors and nothing is present in page json, a no descriptors message is displayed instead, similarly to [PageInspectionPanel.ts](https://github.com/enonic/app-contentstudio/blob/master/modules/lib/src/main/resources/assets/js/app/wizard/page/contextwindow/inspect/page/PageInspectionPanel.ts).

edloidas:
When a Part or Layout component on a page references a descriptor whose application has been removed from the site, the inspect panel now surfaces the persisted key as a synthetic invalid option instead of hiding the dropdown entirely. The chip shows the cached descriptor's display name when available, otherwise the local key name; the description renders as `<Application is missing>` in error red at 12px. The form continues to render whenever the cached descriptor is still available.

Closes #10411


ashklianko:
@edloidas is it ready for testing? one of the tests is failing, not sure if it is time to review it

ashklianko (approved):

PR #10453, opened 2026-05-10
Decouple ContextView from widget composition and host policies #10415

### LINKED ISSUE #10415: Decouple ContextView from widget composition and host policies
ContextView served as both the shell that renders the context panel and the policy that decides what goes into it. In a single class it constructed the built-in widgets for the wizard and browse hosts, fetched contributed widgets from the `contentstudio.contextpanel` interface, listened to install/uninstall events to keep that set in sync, implemented `PageNavigationHandler` and ran the page-editor activation state machine, and reacted to `ContentServerEventsHandler` for item-driven refreshes. Variation between hosts was reduced to a single `editorMode: boolean` constructor parameter, which was enough for browse vs. wizard but not for any other consumer — Content Studio Plus, in particular, needs the same shell with a different widget set and without the page-editor and contributed-widget machinery. Because all of those concerns were initialised from the constructor against private state, neither subclassing nor configuration provided a clean opt-out. This change separates the shell from the policy: ContextView becomes a passive host that renders a widget list it is given and exposes a small composition API; concrete policies — built-in widget construction, contributed-widget loading, page-editor activation — move into focused collaborators that hosts assemble explicitly. ContentBrowsePanel and ContentWizardPanel are migrated to the new shape; behaviour in both hosts is preserved.

#### Rationale

ContextView conflated three orthogonal responsibilities — rendering, widget-set composition, and event-driven activation — and accumulated dependencies on every event source any current host happened to need. The result was a class with one configuration knob and three implicit clients, where adding a fourth (Content Studio Plus) required either forking or accepting unrelated behaviour. Splitting along responsibility lines turns hosts into peers that compose what they need, removes the implicit subclass-or-fork contract, and leaves the rendering surface small enough to be replaced by a React component without dragging policy along.

#### Architecture

- **Shell (`ContextView`)** — owns rendering and the lifecycle of an externally supplied widget list: active widget, default widget, currently inspected item, and the item-driven refresh subscriptions. Exposes `setWidgets`, `addWidget`, `insertWidget`, `removeWidgetByKey`, `replaceWidgetByKey`, `setDefaultWidget`, `activateDefaultWidget` as the composition API. No knowledge of which widgets exist, where they come from, or how navigation drives them.
- **Built-in widget factory (`buildDefaultContextWidgets`)** — produces the standard `properties`, `versions`, `dependencies`, and (optionally) `pageEditor` widget views for hosts that want them. Hosts may use it whole, partially, or not at all.
- **Contributed-widget integration (`loadCustomContextWidgets`, `watchCustomContextWidgets`)** — encapsulates the `contentstudio.contextpanel` interface and the `ApplicationEvent` lifecycle (install / uninstall / start / stop / update, debounced). Lifetime is bound to the host's `ContextView` via `onRemoved`. Hosts that don't expose contributed widgets simply don't call into it.
- **Page-editor controller (`PageEditorContextController`)** — implements `PageNavigationHandler` and owns the `InspectEvent` / `PageEventsManager` / `PageNavigationMediator` subscriptions plus the rule for promoting the page-editor between default and active. Constructed only by hosts that embed the page editor.

#### Host migration

- `ContentBrowsePanel.createContextView()` instantiates `ContextView`, assembles widgets with `buildDefaultContextWidgets`, calls `setWidgets`, and opts into `loadCustomContextWidgets` + `watchCustomContextWidgets`.
- `ContentWizardPanel.createWizardAndDetailsSplitPanel()` does the same with `editorMode: true` and additionally constructs `PageEditorContextController` over the page-editor / properties / versions widgets.
- Content Studio Plus and any future host follow the same pattern with their own widget list and their own choice of collaborators.

ashklianko:


ashklianko:
Required for https://github.com/enonic/app-contentstudio-plus/issues/1777

edloidas (approved):

PR #10457, opened 2026-05-11
Display Name field validation blocks mouse down events #10225

### LINKED ISSUE #10225: Display Name filed validation blocks mouse down events
When the wizard page for creating new content is loaded, clicking in the form (same React root as the Display Name, e.g. the mode toggle button) does not ignore the mouse down events and runs validation instead.

1. Open wizard for content with dropdown (content selector or custom relationship selector)
2. Click on Mode toggle in the dropdown

**BUG  - Mode toggle button does not switch modes on initial wizard load**

https://github.com/user-attachments/assets/db21a679-33d7-440d-87d6-21ef87f36217

skjulestad:
DisplayName keeps same height after editing and single line when not interacted with. If error validation pushes button out of cursor it is ok as we give visual feedback that something is wrong.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
@@ -152,11 +152,12 @@ export const DisplayNameInput = (): ReactElement => {
                     onClick={startEditing}
                     onFocus={startEditing}
                     className={cn(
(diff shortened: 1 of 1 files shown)


ashklianko (changes requested):
Can you also make DisplayName not jump when switching between empty/non-empty display name?

ashklianko:
https://github.com/user-attachments/assets/a0956beb-6908-4adf-ba6f-80708a9d90ba

ashklianko:
Occurrence validation is out of scope of this issue and should be reconsidered on later stages, thus the current fix is ok

ashklianko (approved):

PR #10490, opened 2026-05-13
Image Editor improvements #10422

### LINKED ISSUE #10422: Image Editor improvements
Several issues, feel free to split into subtasks. All these worked correctly in the legacy version, the goal is to make it work exactly like before.

* "Reset" shouldn't appear until the image has been modified. Clicking on one of the tools doesn't modify the image so Reset shouldn't appear right away.
* "Reset" shouldn't save the content
* After modifying an unchanged image and clicking "Reset", "Save" should always go from enabled to disabled state.
* Focal point control should be draggable. Current behaviour is that it can be moved only by clicking on the image, but it should also be possible to drag it (the cursor should change to "move" over the circle)

reisfmb:


ashklianko:
I keep noticing that fully rotating image or flipping it back and forward makes 'Save' button remain enabled. I presume it is because those props are initially empty/missing and when we rotate/flip it creates a non-empty prop -> Save is enabled. 
This is probably not related to this task, but @reisfmb please check it, and if it is a property set related thing then we just approve and merge this improvement.

reisfmb:
@ashklianko 

From my understanding, this is indeed property set related thing. 

`setRotationInPropertySet` and `setFlipInPropertySet` do handle the return-to-identity case correctly, i.e, when the orientation cycles back to 1, they call `set.removeProperty('orientation', 0)` rather than writing 1. 

The underlying lib-admin-ui property set fires a change event on every mutation regardless of whether the final state matches the original, so the "Save" button enables on the first rotate/flip and stays enabled for the rest of the session, even if you end up back at orientation 1

ashklianko (approved):

PR #10499, opened 2026-05-14
Remove ProjectContext

### LINKED ISSUE #10448: Remove ProjectContext
CS classes use two sources of truth for the active project - one in projects.store and another one in legacy ProjectContext. It spawns bugs periodically when classes uses wrong project value.

Remove ProjectContext after replacing all it's usages

alansemenov:


ashklianko:
NB: Required for CS+ changes , https://github.com/enonic/app-contentstudio-plus/issues/1777

PR #10529, opened 2026-05-19
Project Wizard: ambiguous API calls #10521

### LINKED ISSUE #10521: Project Wizard: ambiguous API calls
Several issues that need to be fixed.

1. For every project update 4 API calls are made: `modify`, `modifyLanguage`, `modifyPermissions` and `modifyReadAccess`. The problem is that the first one - `modify` - also contains `language`, `permissions` and `readAccess` fields in its data object. So, either we update these fields twice, or they are not updated by `modify`. In either case, these fields must be removed from `modify` as they are updated by dedicated APIs.
 
2. `modifyLanguage`, `modifyPermissions` and `modifyReadAccess` are called always, even when those fields are not changed in the Wizard. This is especially bad for `modifyReadAccess` which always updates all content items in the project.

3. As mentioned above, `modifyReadAccess` is a heavy operation which may take time. It already returns taskId but we don't track progress of this task. So, if (and only if) read access has been changed and we call `modifyReadAccess` as a part of project update, we need to immediately close the modal dialog and show the progress bar dialog (like we do for content delete, move etc.). Use the existing `dialog.projectAccess.applying` i18n resource in the progress bar dialog.

4. In the legacy version we used to show confirmation whenever Access mode was changed (see below). This got lost in the new version, but we need to get it back. Use the existing `dialog.projectAccess.confirm` i18n resource in the confirmation. This confirmation should be shown when user clicks "Next" after changing Access mode. If user clicks "Cancel" in the confirmation dialog, we should keep him on the current step. this behaviour is a bit similar to what we have in the Permissions wizard when selecting "Replace all child permissions" in the 2nd step.
 
Image

Please create subtasks for each of these issues.

reisfmb:


ashklianko (changes requested):
After latest project api changes `modifyLangugage` calls the same `ProjectService.modify` method, thus there's no sense in even having a separate endpoint for it: language must be update via `modify` call.

One more thing to check: I tried to throttle network and see the progress for changing access mode, but the progress just stays at zero and then jumps immediately to 100%. Try checking on a project with many contents too.

ashklianko:
Same for Create: language now can be set in create params, no need in a separate call

reisfmb:
Didn't know about that update. Good call @ashklianko, will update accordingly.

About the progress, I tried the same and the behavior was the same as you described... then I tested on a project with +1k items, and the progress bar fills correctly.

ashklianko:
Made a few little changes:
- Removed redundant 'modifyLanguage' endpoint and modified tests
- Changing read access when editing a project is now remembered, thus if you come back to the earlier step and go forward again you won't be asked for the confirmation again

ashklianko (approved):

PR #10541, opened 2026-05-20
Publish Tree button is incorrectly enabled in content browse toolbar #9979

### LINKED ISSUE #9979: Publish Tree button is incorrectly enabled in content browse toolbar
1. Open a **Parent** project with item that has "Publish" and "Publish Tree" actions enabled.
    Image
2. Select that **one** item via checkbox.
3. Switch to **subproject** of that project, where that item is editable
    Image
4. You will see that "Publish Tree" action is enabled, but it should not, since selection is empty.
    Image

skjulestad:
Publish Tree button is incorrectly enabled in the content browse toolbar, actions now reset with an explicitly empty list upon opening a new project to avoid stale actions.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
+++ b/modules/lib/src/main/resources/assets/js/app/browse/ContentBrowsePanel.ts
@@ -361,7 +361,9 @@ export class ContentBrowsePanel
 
         onActiveProjectChanged(() => {
             this.selectionWrapper.deselectAll();
-            this.updateActionsAndPreview();
+            this.getBrowseActions().updateActionsEnabledState([]);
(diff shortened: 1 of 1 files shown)


ashklianko:
Can't reproduce it in master, is the scenario in the issue correct?

edloidas:
Yeah, me neither. It's probably irrelevant now. 
@skjulestad were you able to reproduce it on the latest CS/XP?

skjulestad:
I was able to reproduce it multiple times with master from 14 today

skjulestad:
@edloidas @ashklianko it seems I may not be able to reproduce it after pulling master today, but I want this to stay open for now so I can test it in depth later if possible.

alansemenov:
Closing for now, as no one can reproduce

PR #10548, opened 2026-05-21
Language Selector #10131

### LINKED ISSUE #10131: Language Selector - First character not captured in search input
When focusing the input with the mouse and pressing any key, the dropdown opens, but the typed character is not inserted into the input.

This issue does not occur in  `Owner dropdown` , where the input behaves as expected.

Steps to Reproduce:

1. Open the `Edit Settings` dialog.
2. Try typing text in the `language search input`- Norway for example

### **Expected Result:**
Typed characters, including the first one, should appear in the input field.

### **Actual Result:**
The first character is not displayed, and key presses only open the dropdown without inserting text.

Image

skjulestad:
Let language selector show result from first character set, don't display selected language as placeholder to reflect other selectors.


alansemenov:
@skjulestad Note that this problem also appears inside Language selector in the Project Wizard dialog. Check that your fix solves that one as well.

skjulestad:
@alansemenov all usecases of LanguageSelector.tsx works fine now! 😁 

https://github.com/user-attachments/assets/6911eb38-e50b-4a6d-8d14-99df6dbc8ed6

edloidas (approved):

PR #10551, opened 2026-05-21
Incorrect context menu when selecting several items #10087

### LINKED ISSUE #10087: Incorrect context menu  when selecting several items
When selecting multiple content items with different statuses (e.g., Published and Offline) in the Browse Panel grid, the context menu displays options that are not appropriate for the selected combination.

Steps to reproduce :
1. Open the filter panel and enter the name of the first content item(published). Select its checkbox.
2. Enter the name of another content(work in progress) item and select its checkbox.
3. Open `Context Menu`

**Actual Result:**
The context menu shows options corresponding only to one of the selected items instead of combining them.

Image

Image

### **Expected Result:**
Both `Publish` and `Unpublish` menu items should be displayed simultaneously in this case.

skjulestad:
Publish and unpublish action is now based on all items instead of first selected item, replaced [0] with some().


ashklianko (approved):

ashklianko (changes requested):
We discussed the issue with enabling publish/unpublish actions

edloidas (approved):

PR #10565, opened 2026-05-22
No clear function in the Filter panel #10562

### LINKED ISSUE #10562: No clear function in the Filter panel
There's no longer a simple way to reset settings in the Filter panel. We do have "x" inside the search input, but it only clears the input, not the rest of the settings.

We will solve this by having a header in the filter panel with title "Search" (i18n `panel.filter.search`) and right-aligned `Clear` (`panel.filter.clear`) link. I see there's also "x" icon in design, but we don't need it. "Clear" should reset the panel to its default state, including emptying the search input.

https://www.figma.com/design/qkrCEWujVxkjZBd2aI8YXI/content-studio-reskin?t=NPjobyGDaTAIVZts-0

reisfmb:


ashklianko:
- Hooks reordered and put to the top
  - Clear button replaced — native <button> swapped for Button from @enonic/ui 
  - Cosmetic — removed extra blank line between hits row and aggregations list.

ashklianko (approved):

PR #10573, opened 2026-05-25
"Mark as Ready" button flickers before changing to "Publish"

### LINKED ISSUE #10572: "Mark as Ready" button flickers before changing to "Publish"
Several issues:
* After "Mark as Ready" is clicked, it goes to "Publish", then back to "Mark as Ready", then to "Publish" again
* On a slow instance the button might not change to "Publish" at all (and Publish dialog won't open). After page refresh it would be correctly set to "Publish"

Support ticket: https://enonic.zendesk.com/agent/tickets/9151

alansemenov:


alansemenov:
the fix is confirmed by requester

ashklianko (approved):

PR #10587, opened 2026-05-26
Enonic UI: DeleteDialog key navigation #9565

### LINKED ISSUE #9565: Enonic UI: DeleteDialog key navigation
DeleteDialog.tsx has a flaw with keyboard navigation: all the content items in it's list are treated as focusable items, thus when you use a keyboard on a long list there's no way to tab out from the list to action buttons before or after it. 
This is fixed by separating the concept of an active and a focused item, can be seen in our new UI lib components (Listbox and TreeList).

skjulestad:
Separated list focus from active item behavior so keyboard users can arrow through content/reference actions and tab out of long lists. Added shared ContentReferenceGridList and reused it in both delete and unpublish dialogs for consistent navigation and focus styling.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -2924,6 +2924,12 @@ packages:
     peerDependencies:
       postcss: ^8.4.32
 
+  cssnano-utils@5.0.3:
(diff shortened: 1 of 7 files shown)


ashklianko (changes requested):
Several moments I notice:
- Home/End key are not working
- ContentListItemWithReference is dead code now
- focusVisible param is only recently added, I would avoid using not so long existing params

It looks to me that in general `renderContentRow` is kinda overcomplication. We can now miss styling changes in ContentListItem and we won't see them in your list. But if there are no other options then ok.

ashklianko (approved):

PR #10589, opened 2026-05-26
Invalid content icon is missing tooltip (wizard and grid) #690

### LINKED ISSUE #690: Add tooltip to icons for invalid content (in wizard and in grid)
1. Open new  folder-wizard.
2. Move cursor to `toolbar-state-icon`

image

**Expected**: 'Invalid content' message should appear

The same issue in the grid:
image

skjulestad:
Add tooltip to invalid content icon.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/icons/StatusIcon.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/icons/StatusIcon.tsx
@@ -1,6 +1,8 @@
-import {cn} from '@enonic/ui';
+import {cn, Tooltip} from '@enonic/ui';
 import {cva, type VariantProps} from 'class-variance-authority';
 import {CircleAlert, CircleCheck, CircleX, Info, type LucideIcon} from 'lucide-react';
+import {type ComponentProps, type ReactElement} from 'react';
+import {useI18n} from '../../hooks/useI18n';
 
 const statusIconVariants = cva(
     [
@@ -33,8 +35,9 @@ type Status = VariantProps<typeof statusIconVariants>['status'];
 
 type Props = {
     status: Status;
+    withTooltip?: boolean;
     'data-component'?: string;
-} & React.ComponentProps<LucideIcon>;
+} & ComponentProps<LucideIcon>;
 
 
 function getIcon(status: Status): LucideIcon {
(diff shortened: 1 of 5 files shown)


modules/lib/src/main/resources/i18n/phrases.properties
```diff
@@ -331,6 +331,7 @@ status.outofdate=Out-of-date
 status.published=Published
 status.online=Online
 status.offline=Offline
+status.invalid=Invalid content
```
ashklianko: No need to add a new one, there's a 'field.invalid' entry in common.properties in lib-admin, as Alan mentioned in the task

ashklianko (commented):
Image

I don't see the point in adding a tooltip only for the invalid icon: we also have 'Ready' and 'In progress' icons

ashklianko:
Also the task title says '(wizard and grid)', where in the wizard are we expecting this tooltip to appear? @alansemenov

edloidas:
@ashklianko I assume it should be in the Toolbar:
image

modules/lib/src/main/resources/assets/js/v6/features/shared/icons/WorkflowContentIcon.tsx
```diff
@@ -12,10 +13,18 @@ export function WorkflowContentIcon({
...
     return (
         <span className='relative inline-flex items-center'>
             <ContentIcon className='size-6' {...props} />
-            {status && <StatusIcon status={status} className='absolute -top-0.75 -right-0.75' />}
+            {status && (
+                <StatusIcon
+                    status={status}
+                    className='absolute -top-0.75 -right-0.75'
+                    invalidTooltip={status === 'invalid' ? invalidTooltip : undefined}
+                />
```
edloidas: There is absolutely no point in:
- Checking the status being invalid, as you already do it in the StatusIcon
- Adding a customization option via prop and then name it `invalidTooltip`

modules/lib/src/main/resources/assets/js/v6/features/shared/icons/StatusIcon.tsx
```diff
@@ -48,20 +49,30 @@ function getIcon(status: Status): LucideIcon {
...
+    if (status === 'invalid' && invalidTooltip) {
+        return (
+            <Tooltip delay={300} value={invalidTooltip}>
+                {icon}
+            </Tooltip>
+        );
+    }
 
-    return <Icon data-component={componentName} className={classNames} aria-label={status} {...props} />;
+    return icon;
```
edloidas: Just wrap the whole icon in tooltip and that's it.

edloidas (changes requested):
The implementation is suboptimal:
- Wrap the whole icon in tooltip without conditions. It uses the `asChild` by default so it will be merged, and won't create a new div.
- If tooltip value is empty it won't be shown, so noting is set from outside = nothing to show
- Just call it `tooltip`
- Keep the check for whether we need to show the tooltip in the parent element.

alansemenov:
> Also the task title says '(wizard and grid)', where in the wizard are we expecting this tooltip to appear?

The bug should have referred all workflow icons, not just the invalid one. All of the workflow icons (in progress, ready, invalid) are supposed to have tooltips everywhere they are shown.

skjulestad:
Adjusted to add tooltip with existing translations for all statuses now in both use cases of StatusIconWorkflowContentIcon and ContentWizardToolbar :)

edloidas (approved):

PR #10594, opened 2026-05-26
Lock non-localized content wizard inputs #10212

### LINKED ISSUE #10212: Lock non-localized content wizard inputs
Content/Mixin forms must be read-only until the current content was localised

ashklianko:


edloidas (approved):
Looks good. Though it's a bit confusing that we have `enabled` property in FromRenderer. We have to refactor it later.

PR #10595, opened 2026-05-26
Grid keeps selecting an item after delete or reset selection #10034

### LINKED ISSUE #10034: Grid keeps selecting an item after duplicate or reset selection
### **Duplicate scenario**:  
1. Select an item in the grid
2. Click on Duplicate button

https://github.com/user-attachments/assets/c2a78a81-5c95-4c71-836e-2cbff383bd96

### **Refresh Icon scenario**:  

1. Click on the checkbox  in a row item (the item gets checked)
2. Click on Refresh icon in the toolbar 

https://github.com/user-attachments/assets/300725e4-7ef6-49f6-be4e-a5048d73a160

The first item gets selected in Grid after  clicking on Refrsh  icon

**Expected -  the same row should be highlighted after  clicking  on Refresh  icon**

### LINKED ISSUE #10138: Content Item gets unchecked after clicking Refresh con
1. Click on a checkbox, select a item
2. Click on `Refresh` icon

Image

BUG  -  Content Item gets unchecked after clicking Refresh con
            And anothe element gets  highlighted

Image

skjulestad:
Active and checked items are now cleared upon, duplicate, move, delete and refresh. Replaced deprecated resetSelection() with clearSelection() and setActive().

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/TreeListToolbar.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/TreeListToolbar.tsx
@@ -5,23 +5,39 @@ import {RefreshCcw} from 'lucide-react';
 import {type ReactElement, useMemo} from 'react';
 import {activateFilter, fetchRootChildrenIdsOnly, getFilterQuery} from '../../../api/content-fetcher';
 import {useI18n} from '../../../hooks/useI18n';
-import {$isFilterActive} from '../../../store/active-tree.store';
+import {$activeRawFlatNodes, $isFilterActive} from '../../../store/active-tree.store';
 import {clearProjectContentCache} from '../../../store/content.store';
-import {$isAllLoadedSelected, $isNoneSelected, $selectionCount, clearSelection, selectAll} from '../../../store/contentTreeSelection.store';
+import {$activeId, $isAllLoadedSelected, $isNoneSelected, $selectionCount, clearSelection, selectAll, setActive} from '../../../store/contentTreeSelection.store';
 import {$rootLoadingState, resetTree} from '../../../store/tree-list.store';
 
 type TreeListToolbarProps = {
     enabled?: boolean;
 };
 
-const handleReload = (): void => {
-    if ($isFilterActive.get()) {
-        const query = getFilterQuery();
-        if (query) void activateFilter(query);
-    } else {
-        resetTree();
-        clearProjectContentCache();
-        void fetchRootChildrenIdsOnly();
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/TreeListToolbar.tsx
```diff
@@ -5,23 +5,35 @@ import {RefreshCcw} from 'lucide-react';
...
+
 const handleReload = (): void => {
+    const activeId = $activeId.get();
+
     if ($isFilterActive.get()) {
         const query = getFilterQuery();
-        if (query) void activateFilter(query);
+        if (query) void activateFilter(query).finally(() => restoreActiveAfterReload(activeId));
     } else {
+        clearSelection();
```
edloidas: This would reset the selection on reload, while we should actually preserve it.

modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/TreeListToolbar.tsx
```diff
@@ -5,23 +5,35 @@ import {RefreshCcw} from 'lucide-react';
...
-import {$isAllLoadedSelected, $isNoneSelected, $selectionCount, clearSelection, selectAll} from '../../../store/contentTreeSelection.store';
+import {$activeId, $isAllLoadedSelected, $isNoneSelected, $selectionCount, clearSelection, selectAll, setActive} from '../../../store/contentTreeSelection.store';
 import {$rootLoadingState, resetTree} from '../../../store/tree-list.store';
 
 type TreeListToolbarProps = {
     enabled?: boolean;
 };
 
+const restoreActiveAfterReload = (activeId: string | null): void => {
+    if (activeId && $activeRawFlatNodes.get().some((node) => node.nodeType === 'node' && node.id === activeId)) {
```
edloidas: Move this out of `if`, it's too complicated to read there.

edloidas: Can also do early return

modules/lib/src/main/resources/assets/js/v6/features/views/browse/tree/TreeListToolbar.tsx
edloidas: Since we suddenly need to wait for events, make the function async and keep structure flat. You can do `restoreActiveAfterReload` at the very end

edloidas (changes requested):

PR #10598, opened 2026-05-27
Inputs are not updated on server event #10393

### LINKED ISSUE #10393: Inputs are not updated on server event
See the decription  in  https://github.com/enonic/app-contentstudio/issues/7132

1. Open a same content item with HtmlArea in two browser tabs, add some new text into one of them and save - HtmlArea in the other tab is not updated.
2. The same issue occurs with the Text Area, tag input, double input, etc

### LINKED ISSUE #10525: Reset does not work after localization
After localizing content, clicking the Reset button has no effect.

Expected: Reset should revert the localized override and restore the inherited value.
Actual: Clicking Reset does nothing.

#### Rationale
Reset is the primary way to undo a local override in a localized layer. With it broken, users cannot cleanly revert localized fields back to inheriting from the parent.

ashklianko:
-This pull doesn't fix/change Page handling


ashklianko:
Going to show a notification when the form/mixins get updated with server changes

ashklianko:
Also fixes https://github.com/enonic/app-contentstudio/issues/10525

edloidas (approved):

PR #10604, opened 2026-05-27
Focus is not set inside DisplayName for existing content

### LINKED ISSUE #10603: Focus is not set inside DisplayName for existing content
DisplayName gets focused only when its empty. It should always be focused but value inside the input should not be selected as now when the input is clicked. Cursor should always be set in the last position inside the input.

alansemenov:


modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/DisplayNameInput.tsx
```diff
@@ -99,7 +99,8 @@ export const DisplayNameInput = (): ReactElement => {
 
     const handleEditorFocus: FocusEventHandler<HTMLTextAreaElement> = useCallback((): void => {
         setAIContext();
-        requestAnimationFrame(() => inputRef.current?.select());
+        // Place the caret at the start instead of selecting the value.
+        requestAnimationFrame(() => inputRef.current?.setSelectionRange(0, 0));
```
edloidas: Why do we need caret at the start and not the end? It doesn't look useful at all.
Also, meaningless Claude comment since recent;y. Claude like those, let's try to minimize amount of such.

ashklianko: Yes, and the caret in the start is not consistent with browser's behaviour: when you navigate using tab focus is always set to the end

ashklianko (approved):

PR #10611, opened 2026-05-27
Load mask in the preview panel covers modal dialog #10610

### LINKED ISSUE #10610: Load mask in the preview panel covers modal dialog
Image

ashklianko:
- legacy styles in lib-admin mask.less set z-index that we need to override in CS


edloidas (commented):
I think it might be possible to put it inside the `load-mask` without presenting new class. We won't be using load masks in the code anymore so it might be considered safe. Also, ain't there more load-masks out, like in wizard as well?

ashklianko:
This one fixes both browse and wizard preview panels. Not changing it in the load-mask styles to avoid potentially breaking load mask in other places before the release. Can be removed when legacy load masks removed

edloidas (approved):

PR #10620, opened 2026-05-27
Fix focal point in content icons #10618

### LINKED ISSUE #10618: Focal point is not taken into account in content icons
instead of /rest-v2/cs/cms/enonic-homepage/content/content/icon/7c144556-e0ae-4260-b60f-6b2da3b0b633?size=128 icons media:image api should be use instead. this would also include fix to respect focal point.

contentstudio admin tool must declare use of media:image api

rymsha:
## Summary

Icons for image media in Content Studio were served by the legacy
`/rest-v2/cs/<project>/<layer>/content/icon/<id>?size=128` endpoint,
which does not honor the Media's focal point — so an image with an
off-center focal point was cropped from the geometric center.

- `ContentIconUrlResolver` now builds icon URLs for image content via
  `PortalUrlGeneratorService.imageUrl(...)` using the universal
  `media:image` API at scale `block(128,128)` with a `ts` cache buster.
- `media:image` is added to the admin tool's `apis:` allowlist in
  `modules/app/src/main/resources/admin/tools/main/main.yaml`.
- `JsonObjectsFactory` injects `PortalUrlGeneratorService` and forwards
  it to the resolver; `ContentResourceTest` is updated for the new
  constructor signature.
- Custom thumbnail attachments still resolve via the legacy
  `ContentIconResource`, since `media:image` rejects non-image content
  types.

Closes #10618

## Test plan

- [ ] Build the app and deploy on a fresh XP install with Content Studio.
- [ ] Create or use a Media:image content and set a focal point off-center.
- [ ] Open the content browser and verify the icon thumbnail is cropped
      around the focal point (not the geometric center).
- [ ] In DevTools Network, confirm the icon URL is a `media:image` API
      URL (e.g. `/api/media:image/<project>:<branch>/<id>:<hash>/block-128-128/<name>?ts=...`)
      and returns 200 with the expected scaled image.

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentIconUrlResolver.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentIconUrlResolver.java
@@ -2,28 +2,39 @@ package com.enonic.app.contentstudio.rest.resource.content;
 
 import jakarta.servlet.http.HttpServletRequest;
 
-import com.enonic.xp.app.ApplicationNotFoundException;
 import com.enonic.app.contentstudio.rest.resource.ResourceConstants;
 import com.enonic.app.contentstudio.rest.resource.schema.content.ContentTypeIconResolver;
 import com.enonic.app.contentstudio.rest.resource.schema.content.ContentTypeIconUrlResolver;
+import com.enonic.xp.app.ApplicationNotFoundException;
 import com.enonic.xp.attachment.AttachmentNames;
+import com.enonic.xp.branch.Branch;
 import com.enonic.xp.content.Content;
 import com.enonic.xp.content.Media;
 import com.enonic.xp.context.ContextAccessor;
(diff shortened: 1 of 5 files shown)


modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentIconUrlResolver.java
```diff
@@ -79,9 +95,7 @@ private String makeIconPath( final Content content )
 
     private String getProjectName()
     {
-        final String project =
-            ContextAccessor.current().getRepositoryId().toString().replace( ProjectConstants.PROJECT_REPO_ID_PREFIX, "" );
-        return project;
+        return ContextAccessor.current().getRepositoryId().toString().replace( ProjectConstants.PROJECT_REPO_ID_PREFIX, "" );
```
anatol-sialitski:
```suggestion
        return ProjectName.from( ContextAccessor.current().getRepositoryId() ).toString();
```

anatol-sialitski (approved):

PR #10622, opened 2026-05-27
Fix icon cropping in v6 (flags, application, content) #10621

### LINKED ISSUE #10621: Fix icon cropping in v6 (flags, application, content)
Icons in v6 are rendered with incorrect fit modes. Flag icons use a default background sizing that lets the flag image tile/repeat inside the circle instead of fully covering it. Content and application icons use `object-cover`, which crops images (e.g. an uploaded image content's thumbnail) instead of fitting them inside the icon box.

Flags should fully cover their circular container with no repeat. Images in content (and application icons) should use `object-contain` with `crop=false` so the full image is visible without distortion or cropping. For non-image content, the legacy `ContentIconUrlResolver` defaults `crop = !isImage()` (CMS-4677); the v6 `createImageUrl(url, { size, crop })` exposes the same params and should follow the same convention.

Note: `getIconUrl()` may return a URL with backend-appended query params (e.g. cache-busting timestamps). Both the v6 `createImageUrl` and the legacy resolver strip the existing query string before re-encoding, so only the client-supplied `size` and `crop` survive.

Affected files in v6:
- `modules/lib/src/main/resources/assets/js/v6/features/shared/icons/FlagIcon.tsx`
- `modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ApplicationIcon.tsx`
- `modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx`

edloidas:
Fix icon rendering so flags fully cover their circle and content/application icons stop being cropped.

- Flag icon: added `bg-cover bg-no-repeat` so the flag fills the circle without tiling
- Application icon: switched to `object-contain` to stop cropping uploaded icons
- Content icon: unified to `object-contain` with default `crop=false` so images render without cropping or distortion
- `Flag.ts`: added `ar-SA` mapping and inlined `getCodeMap()` reference to `NONSTANDARD_CODES`
- `PreviewContextMenuElement`: dropped redundant `size-full` class

Closes #10621

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/locale/Flag.ts
+++ b/modules/lib/src/main/resources/assets/js/app/locale/Flag.ts
@@ -8,14 +8,13 @@ export interface FlagData {
     classCode: string;
 }
 
-export class Flag
-    extends DivEl {
-
+export class Flag extends DivEl {
     private static CODE_NONE: string = 'none';
 
     private static CODE_UNKNOWN: string = 'unknown';
 
     private static NONSTANDARD_CODES: NonstandardCodes = Object.freeze({
+        'ar-SA': 'sa',
         'ast-es': 'es',
         'af-na': 'na',
         'af-za': 'za',
@@ -110,7 +109,7 @@ export class Flag
         sr: 'rs',
         sv: 'se',
         uk: 'ua',
-        zh: 'cn'
+        zh: 'cn',
     });
 
     private countryCode: string;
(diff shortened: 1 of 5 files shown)


edloidas:
#### Content

| Before | After |
| --- | --- |
| image | image |

#### Flags

| Before | After |
| --- | --- |
| image | image |

alansemenov:
Awesome

PR #10638, opened 2026-05-28
Enonic UI: Issue List modal dialog - Incorrect option selected in Assignee filter dropdown #9758

### LINKED ISSUE #9758: Enonic UI: Issue List modal dialog  -  Incorrect option selected in Assignee filter dropdown
1. Create an issue for an user
2. Log in using the username
3. Open Issue List modal dialog

**BUG**  - All option is selected in the filter dropdown:

Image

### **Expected option** : 

Image

edloidas:
Defaulted the Issue List dialog filter to `assignedToMe` so the dropdown lands on a meaningful slice of data instead of the disabled `All` option when the user has limited visibility.

Simplified `IssueDialogsManager.openListDialog` and its `main.ts` caller — dropped the unused `assignedToMe` flag from `ShowIssuesDialogEvent`. Also tightened `IssueList` empty-state styling.

Closes #9758

### DIFF
--- a/modules/app/src/main/resources/assets/js/main.ts
+++ b/modules/app/src/main/resources/assets/js/main.ts
@@ -377,8 +377,8 @@ async function startApplication() {
 
     CreateIssuePromptEvent.on((event) => IssueDialogsManager.get().openCreateDialog(event.getModels()));
 
-    ShowIssuesDialogEvent.on((event: ShowIssuesDialogEvent) =>
-        IssueDialogsManager.get().openListDialog(event.getAssignedToMe()));
+    ShowIssuesDialogEvent.on(() =>
+        IssueDialogsManager.get().openListDialog());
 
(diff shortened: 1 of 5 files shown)


ashklianko (changes requested):
Regression: with the new assignedToMe default, the dialog can open on the Closed tab showing stale tickets even when there are open issues in all. Repro: user has assignedToMe = {open: 0, closed: 3} and all = {open: 5, closed: 0}. Before this PR the dialog opened on all / open and the user immediately saw the 5 live issues. After this PR syncFilterWithTotals treats assignedToMe as "non-empty" (it has closed counts), skips the fallback, and syncTabWithTotals then flips the tab to closed — so the user lands on a screen of old closed tickets and has to switch both the filter and the tab by hand.

edloidas:
Fixed. Now we go through the open issues in order first.

ashklianko (approved):

ashklianko:
Nice!

PR #10651, opened 2026-05-29
Issue dialog #10637 #10640

### LINKED ISSUE #10637: IssueDialogDetailsContent comments are not scrollable
Create an issue and open Comments tab. You'll see something like this on small screen:
Image

While this is intentionally done to shrink the comments section and allow the input to stay visible. But there are 2 problems on small screens:
- "No comments yet" text is hidden there, since it's flex shank to 0. We should keep this original size, when we show empty text and there are no comments.
- Add the `italic` class and remove the `text-sm` (keep it default size) for empty text.
- Text area is arguably large, considering it auto expands. Let's reduce it to 2 rows by default
- If possible, we should also limit the max size of the input, as it's impossible to expand it automatically indefinitely.

### LINKED ISSUE #10640: Deleting comment closes the issue dialog
Clicking on Delete or Cancel in the comment deletion confirmation dialog results in closing the issues dialog:

https://github.com/user-attachments/assets/82d8812d-6bba-4108-a07e-778f88100064

skjulestad:
IssueDialogDetailsContent comments are not scrollable #10637
Deleting comment closes the issue dialog #10640

Prevent the dialog from closing when clicking within the delete comment dialog, make comments scrollable as well as adding height break points and let the issue dialog fit shorter screens.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -221,7 +221,7 @@ importers:
         specifier: ^10.0.0
         version: 10.0.0(@dnd-kit/core@6.3.1(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)
       '@enonic/ui':
-        specifier: ~1.0.0-beta.9
+        specifier: ~1.0.0-beta.7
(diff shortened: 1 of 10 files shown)


ashklianko (changes requested):
Discussed with Mathias a few more things to do and check

skjulestad:
Fixed focus visible for comment list, adjusted layout after as well as handling focus after leaving a comment with key enter.

ashklianko (approved):

PR #10663, opened 2026-05-29
Content slug is missing a tooltip with full path #10645

### LINKED ISSUE #10645: Content slug is missing a tooltip with full path
We need to show full content path in a tooltip for the content slug in the editor header. The tooltip should also be correctly styled (now it isn't).

Image

skjulestad:
Display full content path for slug in content wizard, details widget and rename dialog description. Display path name as placeholder in rename dialog. Fix unnamed handling, use <unnamed> when content/child/parent is missing path name.


modules/lib/src/main/resources/assets/js/v6/features/utils/cms/content/paths.ts
```diff
@@ -17,6 +20,40 @@ function getContentPath(item: PathLikeContent): string | undefined {
...
+    return path.getElements?.() ?? path.toString().split('/').filter(Boolean);
+}
+
+export type ContentPathDisplayValues = {
+    pathLabel: string;
+    fullPath: string;
+};
+
+export function getUnnamedContentPathLabel(): string {
+    return `<${i18n('field.unnamed')}>`;
```
edloidas: Our utilities must be pure functions. Using i18n adds side effects to it, making it depends on the external state, and mixes presentation part (that should go into hook) with pure logic.

modules/lib/src/main/resources/assets/js/app/browse/ContentWizardToolbar.ts
```diff
@@ -306,6 +310,17 @@ class ContentWizardToolbarElement extends V6ContentWizardToolbarElement {
...
+    private resolveToolbarContentFullPath(item: ContentSummaryAndCompareStatus, contentPath: string): string {
+        const path = item?.getPath();
+        if (!path) {
+            return '';
+        }
+
+        const currentPathName = contentPath || path.getName();
+        const currentPath = ContentPath.create().fromParent(path.getParentPath(), currentPathName).build();
+        return getContentPathDisplayValues(currentPath).fullPath;
+    }
```
edloidas: This function does not use any object properties and can be static. More to that, it should not be in ContentWizardToolbar at all.
Generally, you don't even need to pass it to toolbar component. Just move it to the store

edloidas (changes requested):

ashklianko:
Checking some moments with name/path updates

edloidas (approved):

PR #10679, opened 2026-06-01
Non-unique path should immediately make the content invalid #10469

### LINKED ISSUE #10469: Non-unique path should immediately make the content invalid
When creating content with a name that already exists, clicking the “Mark as Ready” button in the wizard toolbar results in an error.
1. Open wizard for new content
2. Eneter a name that already exists
3. Click on `Mark as ready` button

BUG  - The content becomes Ready for publishing after  clicking  on "Mark as ready" button, but Save button remains enabled

Image

### **Expected behavior:**
We run the path uniqueness check (`contentsExistByPath`) right when the path is generated (debounced), not when the modal dialog is opened. If another content with such path already exists, we should mark the content as invalid and disable saving. We should also highlight the path in the header as invalid so that user understands where the problem is coming from. User can then click it to open the modal dialog and see the error explanation there.

reisfmb:


ashklianko (changes requested):
- We should not block Rename dialog when the name collides, this dialog is the one that actually helps to fix it. 
- We used to show the error text right next to the name input in legacy, but it is not there. anymore. Let's show notification right after the check if the name collides. Pick a better warning text.
- While the name check is running it is possible, especially with slow connection, to click 'Mark as ready'. Let's block the action while the check is on

reisfmb:
The behavior you are describing conflicts from my understanding on what was intended for this task.

About your points:

---

### 1 
The [description](https://github.com/enonic/app-contentstudio/issues/10469#issue-4428460603) specifically says: 
> User can then click it to open the modal dialog and see the error explanation there.

I've checked that with @alansemenov and he pointed out where the click should be: "it's not clicking on the invalid icon, it's clicking on the invalid path". By the "invalid path" I understood the draft content name at the header.

### 2 
This also differs from what was asked... the notification should come after clicking on the invalid path.


### 3
Good point!

---

@ashklianko, please confirm the expected behavior with Alan.

ashklianko:
"...we should mark the content as invalid and disable saving. We should also highlight the path in the header as invalid so that user understands where the problem is coming from. User can then click it to open the modal dialog and see the error explanation there."

So this consists of 2 things to address:
1. Need to change the name design in header so it becomes obvious that it is invalid (I am now looking for the appropriate design)
2. User still can click the rename dialog, and immediately see:
image

I believe you misunderstood the assignment a bit :)

alansemenov:
I don't think we should show notifications. This will result in a new notification popup every time there's a problem with uniqueness.

ashklianko:
@reisfmb instead of notification let's show red text 'Unavailable' next to Name button

ashklianko:
@reisfmb or just use 'text-error' on the name button

ashklianko:
A failing test is not related to the pull, merging

ashklianko (approved):

PR #10723, opened 2026-06-03
Global key shortcuts must toggle on opening/closing modal dialogs #10687

### LINKED ISSUE #10687: Global key shortcuts must toggle on opening/closing modal dialogs
1. Open any dialog, let's say Delete dialog
2. Press 'Alt + M' (open Move dialog shortcut) or '⌘ + E' on mac  'Ctrl + E' on Win  

Result: 2nd dialog is open over the 1st one or new tab opened.

ashklianko:
# Global shortcuts should not fire while a v6 dialog is open

## Summary

Global keyboard shortcuts (browse toolbar, wizard and settings actions) kept firing while a v6 `@enonic/ui` modal dialog was open. A shortcut pressed inside a dialog could trigger a background action and, for example, open a second dialog on top of the first. Legacy `ModalDialog` never had this problem because `open()` shelves the global `KeyBindings` and `close()` restores them; the v6 React dialogs never touch `KeyBindings`, so the gap reappeared.

This adds `KeyBindingsGuard` — a small singleton that watches the DOM for open v6 modals and shelves the global bindings while at least one is open, restoring them when the last one closes. It is mounted from the v6 roots (`App.tsx` for browse + wizard, `SettingsAppShell.tsx` for settings).

**The change is self-contained and trivial to revert — one new file plus two mount points — so it can simply be dropped once the legacy `KeyBindings`/`Action` layer is removed and the action layer is refactored.**

Closes #<issue>

## Why it was done this way

The fix is intentionally scoped to app-contentstudio only (no changes to `lib-admin-ui` or `@enonic/ui`).

- Shelving is the single lever that covers everything. The shortcuts are bound through shared `lib-admin-ui` code: browse and settings via `AppPanel`, the wizard independently via `ContentWizardPanel` — there is no single app-side owner to scope. `KeyBindings.shelveBindings()` does a global `Mousetrap.reset()`, so one observer suppresses every shell's shortcuts at once, mirroring exactly what legacy `ModalDialog` already does.

- Singleton via the window-global `Store`. Studio (`main`) and settings (`settings`) are separate entry bundles, each shipping its own copy of this module. A module-level flag would not dedupe across them and produced two observers / a double shelve. The `Store` singleton is window-global, so there is exactly one observer per window.

- Keyed on standard ARIA (`role="dialog"` + `aria-modal="true"`). Consumers frequently override `data-component` on their `Dialog.Content`, so matching on it was unreliable. Legacy `ModalDialog` sets neither attribute and shelves the bindings itself, so it is correctly left untouched.

- DOM observation instead of per-dialog wiring. There is no central open-dialog registry, and wiring ~35 dialogs (or a `Dialog.Root` wrapper + import swaps) is heavier and easy to forget for future dialogs. A single observer covers current and future dialogs automatically.

## Known limitations

- Contract on `@enonic/ui` markup. Detection depends on `@enonic/ui` modals rendering `role="dialog"` + `aria-modal="true"`. If that markup changes, the shortcuts silently leak again — there is no runtime error and no test guarding the contract yet.

- Workaround, not a source-level fix. It reacts to the DOM rather than to an explicit dialog open/close lifecycle.

- Content Studio Plus is not covered. It is a separate application; its own shortcuts/dialogs would need the same handling, which is only possible from a shared layer (`lib-admin-ui` / `@enonic/ui`).

- Each new v6 shell must mount the guard. Forgetting to mount `<KeyBindingsGuard />` in a future shell would silently reintroduce the leak.

- No automated tests for the guard.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/services/keyBindingsGuard.service.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/services/keyBindingsGuard.service.ts
@@ -1,0 +1,67 @@
+import {KeyBindings} from '@enonic/lib-admin-ui/ui/KeyBindings';
+
+// Standard ARIA markers of @enonic/ui modals. Keyed on role + aria-modal (not
+// data-component, which consumers override); legacy ModalDialog sets neither.
+const MODAL_DIALOG_SELECTOR = '[role="dialog"][aria-modal="true"]';
+
+let started = false;
+let shelved = false;
+let pendingUnshelve = false;
+
+const hasOpenModal = (): boolean => document.querySelector(MODAL_DIALOG_SELECTOR) != null;
+
+const sync = (): void => {
+    if (hasOpenModal()) {
+        pendingUnshelve = false;
+        if (!shelved) {
+            KeyBindings.get().shelveBindings();
+            shelved = true;
+        }
+        return;
+    }
+
+    if (!shelved || pendingUnshelve) {
+        return;
+    }
+
+    // A dialog that swaps its content can briefly remove its modal from the DOM;
+    // defer the unshelve so such a remount does not flap the bindings.
+    pendingUnshelve = true;
+    queueMicrotask(() => {
+        if (!pendingUnshelve) {
+            return;
+        }
+        pendingUnshelve = false;
+        if (!hasOpenModal()) {
+            KeyBindings.get().unshelveBindings();
+            shelved = false;
+        }
+    });
+};
+
+/**
+ * Bridges the v6 React dialogs to the legacy global shortcut system.
+ *
+ * Legacy `ModalDialog` shelves all active `KeyBindings` while open so global
+ * shortcuts (e.g. the browse toolbar's `mod+e`) cannot fire behind a modal. The
+ * `@enonic/ui` dialogs never touch `KeyBindings`, so without this guard those
+ * shortcuts stay live and one pressed inside a dialog can open another on top.
+ * Shelving is global (`Mousetrap.reset`), so a single observer covers every
+ * shell in the window - browse, wizard, settings and archive.
+ *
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/v6/features/services/keyBindingsGuard.service.ts
edloidas, line 1: We don't need it to be a component and mount it. Look at `App.tsx` `initialize` function, that mounts WebSocket. We can do it from here. In the future we will move it to the top, where we mount the root component anyway.

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/KeyBindingsGuard.tsx
```diff
@@ -0,0 +1,79 @@
...
+        pendingUnshelve = false;
+        if (!hasOpenModal()) {
+            KeyBindings.get().unshelveBindings();
+            shelved = false;
+        }
+    });
+};
+
+const initKeyBindingsGuard = (): void => {
+    if (Store.instance().get(GUARD_STORE_KEY)) {
```
edloidas: Are you using Store to sync with CS+?

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/KeyBindingsGuard.tsx
ashklianko: To sync with Settings and CS+

ashklianko: Got what you meant, good point!

edloidas (approved):

PR #10726, opened 2026-06-03
Publish menu is updated incorrectly #10716

### LINKED ISSUE #10716: Publish menu is updated incorrectly
1. Open `Work in progress` content
2. Click on `Publish` menu item

Image

3. Click on `Mark as ready`  then click on `Publish`  buttons

**BUG  - Status and default action are not updated in the toolbar** 

Image

ashklianko:


modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -678,6 +678,10 @@ function applyWorkflowFromServer(nextWorkflow: WorkflowState | null): void {
     }
 }
 
+export function applyServerSideWorkflowState(state: WorkflowState | null): void {
+    applyWorkflowFromServer(state);
+}
+
```
edloidas: Just export and use `applyWorkflowFromServer`

edloidas (changes requested):

PR #10743, opened 2026-06-04
Status icons should not expose tooltips on keyboard tab focus #10742

### LINKED ISSUE #10742: Status icons should not expose tooltips on keyboard tab focus
Status icons in Content Studio can currently become part of keyboard tab navigation and show their tooltip on focus. This affects decorative status indicators such as publish/readiness badges in item lists, where keyboard users should tab to actionable controls rather than the icon tooltip trigger.

After @enonic/ui supports configurable Tooltip triggers, Content Studio should use hover-only tooltips for StatusIcon and keep decorative SVG icons out of the tab order. StatusIcon should pass `trigger="hover"` to Tooltip and set the rendered icon to `focusable="false"` with `tabIndex={-1}`. ContentIcon’s built-in SVG icon should also be removed from tab navigation.

#### Rationale
These status icons are informational/decorative in the affected list rows and should not create extra keyboard stops. Tooltip focus behavior remains available in @enonic/ui by default for interactive components, while Content Studio can opt out only for this specific StatusIcon usage.

Drafted with AI

skjulestad:
This pr is dependant on: https://github.com/enonic/npm-enonic-ui/pull/466

Fix status icon tooltip focus

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx
@@ -19,7 +19,7 @@ import {
     FolderOpen,
     Globe,
     ImageIcon,
-    LucideIcon,
+    type LucideIcon,
     Presentation,
     Shapes,
     SplinePointer,
@@ -107,7 +107,7 @@ export const ContentIcon = ({className, contentType, url, imageSize = 64, crop =
         );
     }
 
-    return <BuiltInIcon className={cn('size-6', className)} contentType={contentType} strokeWidth={1.75} />;
+    return <BuiltInIcon className={cn('size-6', className)} contentType={contentType} strokeWidth={1.75}/>;
 };
 
 ContentIcon.displayName = 'ContentIcon';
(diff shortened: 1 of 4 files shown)


modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx
```diff
@@ -107,7 +107,7 @@ export const ContentIcon = ({className, contentType, url, imageSize = 64, crop =
         );
     }
 
-    return <BuiltInIcon className={cn('size-6', className)} contentType={contentType} strokeWidth={1.75} />;
+    return <BuiltInIcon className={cn('size-6', className)} contentType={contentType} strokeWidth={1.75} tabIndex={-1} />;
```
edloidas: We have both a temporary fix here and the correct one that adds `trigger` property. So we either fix it with this, or wait until new Enonic UI release, update the package.json and move this to draft until t's available.
I'll go with second option. But don't forget to update this PR yourself, when it's ready.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/shared/icons/ContentIcon.tsx
edloidas: Version `1.0.1` released

edloidas (approved):

PR #10767, opened 2026-06-08
“Save” button is missing after content localization. #9555

### LINKED ISSUE #9555: “Save” button is missing after content localization.
1. Navigate to a layer , select then click on Edit on the inherited site  then click on `Localize` button in the wizard 
2. Update the description for example

https://github.com/user-attachments/assets/6d2044b5-d0f9-4da8-8924-e04e245fda95

**BUG**  - Save button should appear after any changes in wizard 

**Refresh the page in browser - Save button gets visible** 

Image

ashklianko:


modules/lib/src/main/resources/assets/js/app/wizard/action/ResetContentAction.ts
```diff
@@ -41,6 +41,11 @@ export class ResetContentAction
             .then(() => new GetContentByIdRequest(contentId).sendAndParse())
             .then((content) => {
                 wizardPanel.replacePersistedContent(content);
+
+                const wizardActions = wizardPanel.getWizardActions();
+                wizardActions.setContent(wizardPanel.getContent()).refreshState();
+                void wizardActions.refreshActions();
```
edloidas: I'd suggest putting it into next then or at least return the `wizardActions.refreshActions()` (even though the actual method calls no real promises), so the potential errors would be caught.

edloidas (changes requested):

edloidas (approved):

PR #10773, opened 2026-06-09
"Edit" button gets focus after toggling filter panel #10727

### LINKED ISSUE #10727: "Edit" button gets focus after toggling filter panel
See the video:

https://github.com/user-attachments/assets/6e0b8d41-37cc-4e15-b8b3-9ce8442669ed

skjulestad:
The browse toolbar now exposes a typed focus handle for `SearchToggle`. When the filter panel closes, focus returns to the search toggle; when it opens, focus remains on the filter search input.

### DIFF
--- a/modules/app/package.json
+++ b/modules/app/package.json
@@ -25,7 +25,7 @@
     "@enonic/lib-admin-ui": "workspace:*",
     "@enonic/lib-contentstudio": "workspace:*",
     "@enonic/page-editor": "^1.0.0",
-    "@enonic/ui": "^1.0.1",
+    "@enonic/ui": "^1.0.4",
     "@radix-ui/react-slot": "^1.2.3",
     "chart.js": "^4.5.1",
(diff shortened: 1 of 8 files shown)


edloidas (changes requested):
This is too much of a hack and creates tech debt. Imagine the legacy wrappers are gone, and you need to solve it in new UI. So, why do double work, if you can do it properly right now?
Please, think how it can be done without ugly query selectors with React-native tools and our new code.

edloidas:
Two fixes landed in `@enonic/ui` that were connected to this issue, that were making current simple solution not working correctly:
- https://github.com/enonic/npm-enonic-ui/issues/474 - Portal (when applying tooltip to the search toggle) and secondary hidden toolbar for width calculations for proper wrapping were assigned same IDs as the regular toolbar items.
- https://github.com/enonic/npm-enonic-ui/issues/476 — hidden items were handled by arrow key navigation.

PR #10774, opened 2026-06-09
Content form doesn't respect server-side validation #10762

### LINKED ISSUE #10762: Content form doesn't respect server-side validation
Content on the screenshot is invalid, which icon in the details panel correctly shows. However, form validation doesn't respect that: icon in the header is incorrect and the field with server-side validation error is not highlighted as invalid.

Image

Image

ashklianko:
- Feed content server validation errors into form2 via ServerErrorsProvider; drop built-in system errors (errorCode `system:*`) in favour of clearer client messages and keep custom app errors. Clear them per occurrence on edit and per field/set on occurrence add/remove/move, and refresh from the saved content after save (postUpdatePersistedItem now reads the server response).


ashklianko:
Requires https://github.com/enonic/lib-admin-ui/pull/4535

edloidas (approved):

edloidas:
A bit too much comments. Asking the same as in lib-admin-ui PR: sometimes they are not needed at all and sometimes renaming function would just make them obsolete.

PR #10780, opened 2026-06-09
Combobox search does not display 'No matching items' message #10749

### LINKED ISSUE #10749: Combobox search does not display  'No matching items'  message
1. Open `Inspect Panel` with component  selector 
2. Enter any text into the search input

**BUG** - 
The Combobox search input should display the “No matching items” message when no options are available after entering text into the search input.

Image

### **Expected: **   “No matching items”  is displayed: 

Image

edloidas:
## Changes

- Show the "No matching items" label (`field.option.noitems`) in the selector popup when a search filters out all options — previously the dropdown rendered an empty popup.
- Applied the empty-state fallback to all three Inspect Panel selectors: `ComponentDescriptorSelector` (the reported case), `PageControllerSelector`, and `FragmentContentSelector`, which share the identical pattern.
- Added render tests for `ComponentDescriptorSelector` covering the no-match search, populated results, and no-descriptors-at-all states.

Closes #10749

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/SelectorPopup.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/context/widget/page-editor/inspect/SelectorPopup.tsx
@@ -1,0 +1,36 @@
+import {Combobox, Listbox} from '@enonic/ui';
+import type {ReactElement, ReactNode} from 'react';
+
+const SELECTOR_POPUP_NAME = 'SelectorPopup';
+
+type SelectorOption = {
+    key: string;
+};
+
+type SelectorPopupProps<T extends SelectorOption> = {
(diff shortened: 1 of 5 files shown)


ashklianko (approved):

ashklianko:
The change is right, but adding similar handler to every dropdown seems like overkill. Would be nice if we could just pass an empty list label in a centralized way

edloidas:
Aside from test and formatting changes that is shown really awkwardly in Git we add just a small part:
```tsx
filteredOptions.length === 0 ? (
    <div className="p-4 text-subtle">{noMatchingLabel}</div>
)
```
So the only thing to move outside - is a popup wrapper with options and this code. I wasn't really sure, since it wasn't looking as a win. 
Anyway, move that part to a different component.

ashklianko (approved):

PR #10789, opened 2026-06-09
Incorrect behaviour of dependency exclusion in the Publishing dialog #10781

### LINKED ISSUE #10781: Incorrect behaviour of dependency exclusion in the Publishing dialog
1. Install Superhero app
2. Open posts/featured-image
3. Remove the existing featured image and upload a new one instead
4. Click "Mark as Ready".
5. Publishing wizard comes up

Image

Several issues here:
1. The new image that was uploaded is nowhere to be seen and will only be shown after clicking on the "Show excluded" link. "Show excluded" should never be shown unless user has excluded something and clicked "Apply".
2. The text in the empty list says "No dependent items" which is not true - there are unpublished dependent items. 
3. The hidden item won't appear even after "Include child items" is turned on, which is even more confusing.

This incorrect behaviour is caused by `publishingWizard.excludeDependencies` config setting which was changed to be "true" by default in Content Studio 6.0. 

This is how exactly the same case looked like in the legacy version, with `publishingWizard.excludeDependencies` set to "true". The only thing that was wrong was that "Hide excluded" was enabled before anything was excluded.

Image

Fix:
1. Return the logic for displaying dependencies we had in the legacy version. We spent quite some time in the past polishing this logic for both modes, so it's important that we are using exactly what we used to have. 
2. The text (`field.publish.dependencies.empty`) is completely wrong and should not be used. If all dependencies are excluded, the text to use is `dialog.dependencies.allExcluded`

### LINKED ISSUE #10802: Replace "Show excluded" toggle with an inline button in the Publishing dialog
The "Show excluded" / "Hide excluded" control in the Publishing dialog is a toggle button that keeps a highlighted background while pressed. It should become a regular click button with the thin inline look of the "Copy from {project}" button in the Project dialogs (the shared `InlineButton`). The shared `SplitList.SeparatorToggle` subcomponent is replaced with `SplitList.SeparatorButton`, which renders `InlineButton` and exposes a plain `onClick` instead of `pressed`/`onPressedChange`. The flipping "Show excluded" / "Hide excluded" label already communicates the state, so the pressed background adds nothing. Since the control lives in the shared `SplitList` component, any dialog that adopts it gets the same appearance.

#### Rationale

A control with both a pressed state and a state-dependent label double-encodes the same information, and `aria-pressed` combined with a changing label is an ARIA anti-pattern. Rendering the shared `InlineButton` aligns the control with other inline actions and makes the separator row thinner.

edloidas:
Restores the legacy dependency-display logic in the v6 Publishing dialog. Dependants auto-excluded by the `publishingWizard.excludeDependencies` config are now listed unchecked instead of being hidden, with transitive dependants of excluded items kept out of the list (legacy `calcVisibleIds` via `getNextDependants()`).

The "Show/Hide excluded" toggle now appears only when there are toggleable excluded dependants, defaults to the shown state, and is hidden while the selection is being edited. Toggling "Include child items" or removing a main item resets dependant exclusions and re-runs the auto-exclude pass, so previously excluded children are re-included. The misleading `field.publish.dependencies.empty` text was removed; an all-excluded list now shows `dialog.dependencies.allExcluded`.

Additionally, the "Show/Hide excluded" control is no longer a toggle button: `SplitList.SeparatorToggle` is replaced with `SplitList.SeparatorButton`, which renders the shared `InlineButton` (the thin "Copy from" button style from the Project dialogs) with a plain `onClick` — the flipping label alone conveys the state.

Closes #10781
Closes #10802

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -32,6 +32,9 @@ importers:
       '@enonic/page-editor':
         specifier: ^1.0.0
         version: 1.0.0
+      '@enonic/ui':
+        specifier: ^1.0.1
+        version: 1.0.2(@radix-ui/react-slot@1.2.4(@types/react@19.2.14)(react@19.2.4))(focus-trap-react@11.0.6(@types/react-dom@19.2.3(@types/react@19.2.14))(@types/react@19.2.14)(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(lucide-react@0.545.0(react@19.2.4))(preact@10.29.0)(react-dom@19.2.4(react@19.2.4))(react-virtuoso@4.18.3(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)(tw-animate-css@1.4.0)
       '@radix-ui/react-slot':
         specifier: ^1.2.3
         version: 1.2.4(@types/react@19.2.14)(react@19.2.4)
@@ -178,11 +181,11 @@ importers:
         specifier: ^1.2.0
         version: 1.2.4(@types/react@19.2.14)(react@19.2.4)
       dayjs:
-        specifier: ^1.11.20
-        version: 1.11.20
+        specifier: ^1.11.21
+        version: 1.11.21
       dompurify:
-        specifier: ~3.4.5
-        version: 3.4.5
+        specifier: ~3.4.8
+        version: 3.4.8
       fine-uploader:
         specifier: ^5.16.2
         version: 5.16.2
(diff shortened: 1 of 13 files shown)


ashklianko (approved):

alansemenov (changes requested):
Image

"Show/hide excluded" is probably a toggle button now, but it takes too much attention (and it's not really toggle if the caption changes).

Can we change this to a regular button? This one, with transparent background: https://enonic.github.io/npm-enonic-ui/?path=/story/components-button--text

edloidas:
Well, it is a toggle, isn't it. We don't execute any additional request or do computations. We show and hide the exclusions.
But we can change it to regular button. Do you want to do it in this PR?

alansemenov:
> But we can change it to regular button. Do you want to do it in this PR?

Yes please

PR #10792, opened 2026-06-10
Components rendering errors are not displayed in the PCV #10675

### LINKED ISSUE #10675: Components rendering errors are not displayed in the PCV
1. Open a site , insert  a component that contains errors

**Current validation behavior:**

Image

### **Expected behavior: the item should be marked as invalid**.

Image

ashklianko:
## Summary

Components that fail to render in the live edit are now marked as invalid in the v6 page components tree, so a broken part/layout/fragment is visible in the outline even though it does not affect form validity.

## What changed

- `LiveEditPageProxy` reads the render-error snapshot from the page editor protocol's `ready` message (`errorPaths`) and forwards the paths via `PageEventsManager.notifyPageRenderErrors`.
- The page editor bridge stores them in a dedicated `$renderErrorComponentPaths` fact store; `PageComponentsView` merges them into each node's visual `invalid` flag.
- Incremental tracking supplements the snapshot: `component-loaded` clears a mark, `component-load-failed` adds one.
- Render-error marks are cleared on page reload and on component add/remove/move.
- Bumped `@enonic/page-editor` to 2.0.0-beta.1 (`modules/lib` and `modules/app`), which introduces the `ready.errorPaths` payload.

## Why / how it works

```
editor posts ready{errorPaths} -> LiveEditPageProxy -> PageEventsManager.notifyPageRenderErrors -> bridge -> $renderErrorComponentPaths -> PageComponentsView -> OctagonAlert on the node
```

The marker reuses the existing invalid visual but is kept separate from form validation: render errors are a real render failure, so they show regardless of validation visibility and never feed `$invalidComponentPaths` / `setWizardFormValidation`.

## Notable findings / decisions

- **Positional paths drift.** Node ids are positional (`/main/0`). The error set is a snapshot taken at render time; add/remove/move shift sibling indices, and most edits are partial reloads (no fresh snapshot), so the marks would stick to the wrong position. A full positional remap would duplicate the tree's index-shift logic (especially for move) at notable risk, and is overkill for an edge case. We clear the marks on structural changes instead; they are re-reported on the next full render. Move is covered because it fires both component-removed and component-added; config updates do not shift positions and are left untouched.
- **The wire contract carries only error paths, not full component records.** The host already owns the page model; the editor reports just the DOM-derived render-error state. See enonic/npm-page-editor#86.

Closes #10675

Requires `@enonic/page-editor` >= 2.0.0-beta.1 (enonic/npm-page-editor#86)

### DIFF
--- a/modules/lib/src/main/resources/assets/js/page-editor/event/LiveEditPageViewReadyEvent.ts
+++ b/modules/lib/src/main/resources/assets/js/page-editor/event/LiveEditPageViewReadyEvent.ts
@@ -1,33 +1,11 @@
 import {ClassHelper} from '@enonic/lib-admin-ui/ClassHelper';
-import {type ComponentPath} from '../../app/page/region/ComponentPath';
 import {IframeEvent} from '@enonic/lib-admin-ui/event/IframeEvent';
-
-export interface LiveEditComponentRecord {
-    path: ComponentPath;
-    type: string;
-    parentPath?: string;
-    children: readonly string[];
-    empty: boolean;
-    error: boolean;
-    descriptor?: string;
-    loading: boolean;
-}
 
 export class LiveEditPageViewReadyEvent
     extends IframeEvent {
 
     constructor() {
         super();
-    }
-
-    getComponents(): LiveEditComponentRecord[] {
-        const data = this.getData();
-        if (data == null || !('components' in data)) {
-            return [];
-        }
-
-        const {components} = data;
-        return Array.isArray(components) ? components : [];
     }
 
     static on(handler: (event: LiveEditPageViewReadyEvent) => void, contextWindow: Window = window) {
--- a/modules/lib/src/main/resources/assets/js/page-editor/event/LiveEditPageViewReadyEvent.ts
+++ b/modules/lib/src/main/resources/assets/js/page-editor/event/LiveEditPageViewReadyEvent.ts
@@ -1,11 +1,33 @@
 import {ClassHelper} from '@enonic/lib-admin-ui/ClassHelper';
+import {type ComponentPath} from '../../app/page/region/ComponentPath';
 import {IframeEvent} from '@enonic/lib-admin-ui/event/IframeEvent';
+
+export interface LiveEditComponentRecord {
+    path: ComponentPath;
+    type: string;
+    parentPath?: string;
+    children: readonly string[];
+    empty: boolean;
+    error: boolean;
+    descriptor?: string;
+    loading: boolean;
+}
 
 export class LiveEditPageViewReadyEvent
     extends IframeEvent {
 
     constructor() {
         super();
+    }
+
+    getComponents(): LiveEditComponentRecord[] {
+        const data = this.getData();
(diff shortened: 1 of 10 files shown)


modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsView.tsx
```diff
@@ -53,6 +53,8 @@ export const PageComponentsView = ({showTitle = false}: PageComponentsViewProps
     const componentsLabel = useI18n('field.components');
     const pageVersion = useStore($pageVersion);
     const invalidComponentPaths = useStore($invalidComponentPaths);
+    const renderErrorComponentPaths = useStore($renderErrorComponentPaths);
+    console.log(renderErrorComponentPaths);
```
edloidas: Remove log

modules/lib/src/main/resources/assets/js/page-editor/event/LiveEditPageViewReadyEvent.ts
```diff
@@ -1,13 +1,38 @@
 import {ClassHelper} from '@enonic/lib-admin-ui/ClassHelper';
+import {type ComponentPath} from '../../app/page/region/ComponentPath';
 import {IframeEvent} from '@enonic/lib-admin-ui/event/IframeEvent';
 
+// Serializable view of a live edit component, carried in the event data once
+// the page view is ready. Mirrors the page editor's component registry record
+// without the (non-serializable) DOM element.
```
edloidas: Lots of useless comments in commit

modules/lib/src/main/resources/assets/js/v6/widgets/inspectors/model/page-editor/bridge.ts
```diff
@@ -55,10 +57,21 @@ export function initPageEditorBridge(options?: InitPageEditorBridgeOptions): voi
...
+        // re-reports them on the next ready.
+        setRenderErrorComponentPaths([]);
+    };
     mgr.onBeforeLoad(onBeforeLoad);
     cleanups.push(() => mgr.unBeforeLoad(onBeforeLoad));
 
+    const onRenderErrors = (paths: ComponentPath[]) => {
+        setRenderErrorComponentPaths(paths.map((path) => path.toString()));
+    };
+    mgr.onPageRenderErrors(onRenderErrors);
```
edloidas, line 74: What about `mgr.onComponentLoadFailed` and `mgr.onComponentLoaded` in case we mutate the iframe and try loading component from elsewhere, and it fails (add to paths) or component loaded and errors are gone (remove from paths)?

edloidas (changes requested):

edloidas:
I think we will postpone with this a bit and with page editor part, as there will be massive rework of the PE itself and how it works, that would require more changes to these PRs.

ashklianko:
Switching to draft then

edloidas (approved):

PR #10793, opened 2026-06-10
Enonic UI: Fix sorting in issue details dialog #10791

### LINKED ISSUE #10791: Enonic UI: Fix sorting in issue details dialog
Items in issue details dialog is not properly sorted.

https://github.com/user-attachments/assets/43fa6150-1446-4adc-be19-b6f046c47177

This is the same issue in three different browser tabs opened within minutes of each other, order of items seem "random".

skjulestad:
Add server-side `_path` ordering for `resolveByIds` and use it when loading issue dialog items, so issue details keep correct path order.

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/json/PublishRequestJson.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/json/PublishRequestJson.java
@@ -1,6 +1,6 @@
 package com.enonic.app.contentstudio.rest.resource.content.json;
 
-import java.util.Set;
+import java.util.List;
 import java.util.stream.Collectors;
 
 import com.enonic.app.contentstudio.json.issue.PublishRequestItemJson;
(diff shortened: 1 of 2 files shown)


edloidas (commented):
I'm rather skeptical about this solution, as it's done on client part. You say that data is received from the server unsorted. But we may have lots of data there that will be loaded in batches. But if you do sorting on a client, you can't guarantee, that you won't get any content in next batch that will be placed somewhere between already sorted content. And this fact makes this approach useless.

edloidas:
We have somewhat relative issue merged recently https://github.com/enonic/app-contentstudio/pull/10810
Please, see how this can be done by sorting things on server side yourself to try to consult Sasha how properly implement that kind of sorting.

ashklianko:
Good job Mathias, I'll take it from here since as Mikita mentioned there was relative tasks recently regarding performance of dialogs that have dependant items.

ashklianko:
The issue was in Java part of CS backend, in a classical problem with items order in List vs Set.

edloidas (approved):

PR #10803, opened 2026-06-11
Disable inherited non-localized content editing #10728

### LINKED ISSUE #10728: Disable inherited non-localized content editing
Currently we disable only inputs inside content form of an Inherited content which hasn't been localised yet. This doesn't work very well for items that don't have inputs: folders, collapsed sets etc.

We should (while keeping the inputs disabled) apply shader mask to the entire form. opacity-30 should then be removed from inputs and applied to the entire form in each tab.

Now:
Image

Should be:
Image

skjulestad:
Updates inherited non-localized handling across wizard tabs and page inspection.

Adds shared `useInheritedNonLocalized()` hook and applies it to content tabs, detached page components, page controller selector, customize button, and page config form.

The page components list now disables drag, click, toggle, hover affordance, and context menu when content is inherited and data-inherited.

Keeps existing opacity treatment on tab contents while blocking editing/actions in page-specific controls.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/hooks/useInheritedNonLocalized.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/hooks/useInheritedNonLocalized.ts
@@ -1,0 +1,10 @@
+import {useStore} from '@nanostores/preact';
+import {$wizardToolbar} from '../store/wizardToolbar.store';
+
+export function useInheritedNonLocalized(): boolean {
+    const {isContentInherited, isContentDataInherited} = useStore($wizardToolbar, {
+        keys: ['isContentInherited', 'isContentDataInherited'],
+    });
+
+    return isContentInherited && isContentDataInherited;
+}
--- a/modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/ContentWizardTabs.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/ContentWizardTabs.tsx
@@ -1,6 +1,7 @@
 import {Tab} from '@enonic/ui';
 import {useStore} from '@nanostores/preact';
 import {type ReactElement, useEffect, useState} from 'react';
+import {useInheritedNonLocalized} from '../../../hooks/useInheritedNonLocalized';
 import {useI18n} from '../../../hooks/useI18n';
 import {
     $contentTypeDisplayName,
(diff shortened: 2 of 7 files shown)


modules/lib/src/main/resources/assets/js/v6/features/hooks/useInheritedNonLocalized.ts
```diff
@@ -0,0 +1,10 @@
+import {useStore} from '@nanostores/preact';
+import {$wizardToolbar} from '../store/wizardToolbar.store';
+
+export function useInheritedNonLocalized(): boolean {
+    const {isContentInherited, isContentDataInherited} = useStore($wizardToolbar, {
+        keys: ['isContentInherited', 'isContentDataInherited'],
+    });
+
+    return isContentInherited && isContentDataInherited;
+}
```
edloidas, line 10: No need for a hook. This can just be computed store atomic value that you can use just the same way, but via `useStrore` hook.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/ContentWizardTabs.tsx
```diff
@@ -30,9 +31,11 @@ export const ContentWizardTabs = ({tabListAction}: ContentWizardTabsProps): Reac
     const xDataTabs = useStore($mixinsTabs);
     const invalidTabs = useStore($invalidTabs);
     const validationVisibility = useStore($validationVisibility);
+    const isInheritedNonLocalized = useInheritedNonLocalized();
     const pageTabLabel = useI18n('field.page');
     const [activeTab, setActiveTab] = useState('content');
     const showErrors = validationVisibility === 'all';
+    const inheritedContentClassName = isInheritedNonLocalized ? 'opacity-30' : undefined;
```
edloidas, line 38: What you did there is you added 30% opacity to every tab. But you didn't remove the opacity from existing disabled elements, and that is wrong, as they as a result will have 0.7*0.3 opacity.
You need to find a way to apply opacity and disable content consistently.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsItem.tsx
```diff
@@ -73,17 +75,26 @@ export const PageComponentsItem = ({
 
     const handleToggleClick = (e: MouseEvent<HTMLButtonElement>): void => {
         e.stopPropagation();
+        if (disabled) {
+            return;
+        }
```
edloidas, line 80: Excessive guards, as this is already handled by `pointer-events: none;` and by `disabled` attribute.

modules/lib/src/main/resources/assets/js/v6/features/views/wizard/content-wizard-tabs/page-components/PageComponentsItem.tsx
```diff
@@ -73,17 +75,26 @@ export const PageComponentsItem = ({
...
+
         onToggle(node.id);
     };
 
     const handleClick = (): void => {
+        if (disabled) {
+            return;
+        }
```
edloidas, line 88: Same here. Not needed.

edloidas (changes requested):

edloidas:
Closed in favor of https://github.com/enonic/app-contentstudio/pull/10866

PR #10818, opened 2026-06-12
Lazy-load dependants and drop per-content permission checks #10814

### LINKED ISSUE #10814: Request Publishing dialog is not lazy-loading dependencies
A continuation of https://github.com/enonic/app-contentstudio/issues/10796

Fixing the same problem in Request publishing dialog, Issue create, Issue list dialog

ashklianko:
Extends the publish dialog's windowed dependency loading and server-resolved publish status to the Request Publish, New Issue, and Issue Details dialogs, and removes the per-content `effectivePermissions` lookups behind the assignee selector.

## Assignee selector

- Removed the per-content effective-permissions filtering in `useAssigneeSearch` (and its `publishableContentIds` / `useRootFallback` inputs). It fired one `GetEffectivePermissionsRequest` per publishable content on every dependency reload, flooding the server, and filtered by per-content `PUBLISH` ACE — inconsistent with how the server actually validates approvers (project `OWNER`/`EDITOR`/admin roles).
- The selector now loads all users; the server filters invalid approvers on submit and the existing "some assignees lack rights" warning surfaces the result — matching legacy behavior.
- Deleted the now-unused `useIssuePublishTargetIds` hook; affects Request Publish, New Issue, and Issue Details.

## Dependant loading (windowing)

- Dependant summaries now load lazily in windows of 36 via `onEndReached`, instead of fetching every summary in one request. Toggling "include children" no longer pulls the full dependency tree's summaries at once.
- Counts (`createCount` / `itemsCount`) and `publishable` are derived from the full dependant **id** set, so they stay correct regardless of how many summaries are loaded.
- Request Publish now sources its publishable count from the server (`getPublishable()`, status != `EQUAL`, permission-independent) instead of computing `isOnline` over loaded summaries. Note: this couples the dialog to the backend `publishableContents` field.

## Shared helper

- Extracted the windowing logic into `utils/cms/content/dependantWindow.ts` (`createDependantWindowLoader`, `fetchDependantWindowSlice`, `mergeDependantWindow`, `pruneDependantWindow`, `orderSummariesByIds`, `DEPENDANT_LOAD_SIZE`), removing ~120 lines of duplication across the three stores and giving a single place to later wire row virtualization. (The publish store keeps its atom-based implementation for now.)

## Correctness & UX

- A first-window summary fetch that fails transiently (resolve succeeds, summaries return empty) now surfaces the dialog's error state instead of a silent empty list with no retry path.
- Stale-result guards on lazy loads (`instanceId` / `dependenciesRequestId`) plus reconciliation against the current id set prevent re-adding dependants removed mid-fetch.
- Required (non-deselectable) dependants now disable only the checkbox, not the whole row — visual parity with the publish dialog.

Closes #10814


edloidas (approved):
There is rather big optimizations missing as well: the dependants use `resolveByIds` for every ID it loads and does not use our content cache. Cache already exist and will drastically improve load time on non-local fetches. Either implement it here or create a new task

ashklianko:
This one and other related tasks are rather hot fixes, further optimisations to be done in a separate task.

PR #10820, opened 2026-06-12
Incorrect progress reporting in Publish, Unpublish and Archive dialogs

### LINKED ISSUE #10819: Incorrect progress reporting in Publish, Unpublish and Archive dialogs
Progress dialogs for long-running content operations report progress incorrectly or not at all, making them look frozen on large content trees (tested with a published site of 2569 items). The following problems were observed:

  - Archiving a published site shows "Archiving N item(s)..." with an empty progress bar for a long time (while items are being unpublished), then the bar instantly fills up to 100% and stays there for roughly the same time until the operation completes. The unpublish step is not reflected in the dialog at all, and the progress total only counts the selected items, not their descendants.
  - Unpublishing a site reports all items in a single progress callback: the bar jumps to its final value early and then remains stuck while the server runs per-item post-processing for several minutes.
  - Publishing a site has a near-minute pause with an empty progress bar before the first progress event arrives.
  - During the initial resolution phase of all these operations, the dialogs give no feedback whatsoever, so it is impossible to tell whether the operation is running or the dialog has frozen.
  - While an operation started from a dialog is running, the same dialog keeps re-resolving its own dependency list in response to the content events produced by that operation (e.g. repeated `resolveForUnpublish` requests of ~100 kB each), generating significant unnecessary server load.

  #### Rationale
Editors archiving, publishing or unpublishing large sites cannot tell whether the operation is progressing, stuck, or failed, and the redundant resolution requests add avoidable load on the server during the heaviest operations.

alansemenov:
Reworked progress reporting for the Archive, Unpublish and Publish operations and the corresponding dialogs.

  - Tasks now report the current phase through the task progress info as a machine-readable marker (`{"phase":"unpublish","count":N}`), with all user-facing texts localized client-side. A shared `TaskPhases` helper produces the markers, and `useTaskProgress` exposes `phase` and `phaseTotal` to the dialogs.
  - The archive task resolves the full descendant tree up front, explicitly unpublishes the published items first and then archives, reporting both operations on a single cumulative progress scale. The dialog shows "Unpublishing N item(s)..." followed by "Archiving N item(s)..." with a continuously filling bar.
  - Unpublishing (both standalone and within archive) processes items deepest-first in batches of 50, so every item produces its own progress callback and the server-side post-processing is spread across the operation instead of running silently after the bar is full.
  - The publish task emits its phase marker at the first actual push callback, so the silent stretch (dependency resolution, validity checks, version preparation) is reported as a resolving state.
  - Until the first phase marker arrives, dialogs show "Resolving items..." with an animated ellipsis and no progress bar; the bar appears when real progress starts. `ProgressDialogContent` hides the bar when `progress` is undefined.
  - Dialog stores no longer reload their resolved dependency lists in response to content events caused by their own running task.

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/archive/ArchiveContentProgressListener.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/archive/ArchiveContentProgressListener.java
@@ -1,10 +1,11 @@
 package com.enonic.app.contentstudio.rest.resource.archive;
 
 import com.enonic.xp.archive.ArchiveContentListener;
+import com.enonic.xp.content.PushContentListener;
 import com.enonic.xp.task.ProgressReporter;
 
 public final class ArchiveContentProgressListener
-    implements ArchiveContentListener
+    implements ArchiveContentListener, PushContentListener
 {
     private final ProgressReporter progressReporter;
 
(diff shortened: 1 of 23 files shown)


ashklianko (approved):
Looks good

PR #10824, opened 2026-06-12
#10822 Auto preview cannot skip browser-only failures

### LINKED ISSUE #10822: Auto preview cannot skip browser-only failures
The automatic preview mode picks the first widget whose probe succeeds. The probe is answered from the server's network view, but some liveview widgets (e.g. app-liveview-iframe's same-origin wrapper) only learn whether they can render once their document probes from the browser — the server and browser views can disagree (XP in Docker, remote XP with a local dev server).

Today such a widget either blocks auto mode on its error view, or — if Content Studio reacted to a late failure — the user would see the widget flash before auto mode advances. The probe contract is synchronous, so the browser-side verdict needs an equally seamless treatment: keep the loading mask up until the promised verdict arrives, and advance behind it.

rymsha:
Closes #10822

The widget probe is synchronous and answered from the server's network view, but some liveview widgets (e.g. the same-origin wrapper in enonic/app-liveview-iframe#22) only learn whether they can render once their document probes from the browser. This adds the verdict half of that contract to `ExtensionRenderingHandler`:

- A widget declares `verdict: true` in its probe's `enonic-widget-data` → Content Studio frames it but **keeps the loading mask up**.
- The widget posts `liveview-render-verdict` (`{ok, contentId, url}`, IframeEventBus envelope): **ok** reveals the preview; **not ok** marks the widget failed for the current content and auto mode re-renders the remaining candidates **behind the mask** — no widget flash.
- Explicit selection keeps the widget's own failure UI; auto mode with no remaining candidate reveals the widget's failure view instead of rendering nothing; a 6s fail-open timeout guards against a widget that promises a verdict and never delivers; stale verdicts are dropped by contentId; the failed set resets on content change.
- Widgets that declare nothing reveal on load exactly as before — no timeout penalty for legacy widgets.

Companion: enonic/app-liveview-iframe#22 (wrapper + verdict emission). Independent of #10788 — works with or without the CSP adoption branch; `tsc` and eslint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
+++ b/modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
@@ -1,6 +1,7 @@
 import {DivEl} from '@enonic/lib-admin-ui/dom/DivEl';
 import {type Element} from '@enonic/lib-admin-ui/dom/Element';
 import {type IFrameEl} from '@enonic/lib-admin-ui/dom/IFrameEl';
+import {IframeEventBus} from '@enonic/lib-admin-ui/event/IframeEventBus';
 import {type Extension} from '@enonic/lib-admin-ui/extension/Extension';
 import {StatusCode} from '@enonic/lib-admin-ui/rest/StatusCode';
 import {type Action} from '@enonic/lib-admin-ui/ui/Action';
@@ -24,6 +25,28 @@ export enum PREVIEW_TYPE {
     EMPTY,
     FAILED,
     MISSING,
+}
+
+// Deferred-render protocol: a renderer whose renderability is only known in the
+// browser (after the server probe passed) declares deferReveal:true in its probe
+// widget data, then posts one RENDER_STATUS_EVENT once it knows. Content Studio
+// holds the loading mask until then and never reveals such a renderer on load.
+// Renderers echo the renderToken Content Studio put on their URL, so messages
+// from a candidate a newer render has superseded can be dropped.
+const RENDER_STATUS_EVENT = 'preview-render-status';
+
+// Safety net only: a correct renderer always posts a status, so this never fires
+// for one. It reveals anyway if a renderer dies without reporting, so a bug cannot
+// wedge the mask open. Kept above the wrapper's own 4s probe timeout.
+const RENDER_STATUS_TIMEOUT_MS = 6000;
+
+type RenderStatus = 'ready' | 'unavailable';
+
+interface RenderStatusData {
+    status?: RenderStatus;
+    contentId?: string;
+    renderToken?: string;
+    message?: string;
 }
 
 export class ExtensionRenderingHandler {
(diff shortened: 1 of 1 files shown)


modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
```diff
@@ -357,13 +472,22 @@ export class ExtensionRenderingHandler {
     protected bindListeners() {
         ViewExtensionEvent.on(this.handleExtensionEvent.bind(this));
         EmulatedDeviceEvent.listen(this.handleEmulatorEvent.bind(this));
+        IframeEventBus.get().onEvent(RENDER_STATUS_EVENT, (event) => {
+            // the bus delivers the parsed message detail, not an IframeEvent
```
pmi, line 476: It does inflate it as long as you register the class with `IframeEventBus.get().registerClass('EventName', EventClass)`, but it's the legacy approach and the DTO is preferred.

modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
```diff
@@ -266,6 +326,10 @@ export class ExtensionRenderingHandler {
                 break;
             }
         }
+        if (isOk) {
+            this.renderedExtension = extension;
+            this.renderedInAutoMode = isAuto;
```
pmi, line 331: Both of these are available to the widget and therefore can be sent from the iframe wrapper in the `RenderStatusData` event instead of creating extra fields here.

modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
```diff
@@ -331,6 +395,57 @@ export class ExtensionRenderingHandler {
...
+            return; // stale: a different content is selected now
+        }
+        if (data?.renderToken != null && Number(data.renderToken) !== this.renderSeq) {
+            return; // stale: a newer render has superseded this candidate
+        }
+        if (data?.status === 'ready') {
+            this.clearVerdictWait(true);
+            return;
+        }
+        if (!this.renderedInAutoMode || !this.renderedExtension) {
```
pmi, line 409: Add those to `RenderStatusData` event returned from the iframe wrapper

modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
```diff
@@ -52,6 +75,25 @@ export class ExtensionRenderingHandler {
 
     protected renderableChangedListeners: ((isRenderable: boolean, wasRenderable: boolean) => void)[] = [];
 
+    private failedExtensions: Set<string> = new Set<string>();
+
+    private renderedExtension: Extension;
+
+    private renderedInAutoMode: boolean = false;
```
pmi, line 82: Used to carry info between async events and should be added to the event itself.

modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
pmi, line 158: All the listeners are still notified if the page is renderable the old way, paying no attention to new verdict.

modules/lib/src/main/resources/assets/js/app/view/ExtensionRenderingHandler.ts
rymsha: This is just a demo - how to workaround CSP. 
Feel free to implement it the way you like it

PR #10839, opened 2026-06-15
Missing default values in lazy-loaded mixins and site config #10782

### LINKED ISSUE #10782: Missing default values in lazy-loaded mixins and site config
**Case 1 :** 

1. Open wizard for new site, select `Super Hero` app and enter a name .
2. Blog config tab becomes visible with red icon , so  the site is invalid:

Image

**BUG  - red icon should not be displayed , becase the  required input is selected in the Blog config  tab**

3. Click on `Blog config` tab 

Image

**BUG  - Red icon disappears in the Blog config tab, but the content remains invalid**

4. Insert a text  in the HtmlArea and press Save button 

The site gets valid

Image

**Case 2 :** 

1. Open wizard for new site 
2. Insert a name and select a app with `OptionSet` in site config: 
3. Click on `Edit` config Icon 

**BUG  - The required inputs in the site config contain values, but the site is still invalid..**

Image

ashklianko:


modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -430,6 +433,24 @@ function cloneMixins(mixins: Mixin[]): Mixin[] {
...
+
+function createSeededMixinTree(descriptor: MixinDescriptor): PropertyTree {
+    const tree = new PropertyTree();
+    try {
+        seedFormDefaults(descriptor.toForm(), tree.getRoot());
+    } catch (e) {
+        DefaultErrorHandler.handle(e);
+    }
+    return tree;
+}
```
edloidas, line 442: We try to drift away from try-catch in non-core functions.
We should go directly into the place where the error can be thrown and wrap that place and handle it there or better propagate via return `Result` from 'neverthrow' and handle it at some point where it should be handled. And don't use Default error handler. Just error log that result message.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -430,6 +433,24 @@ function cloneMixins(mixins: Mixin[]): Mixin[] {
     return mixins.map((mixin) => mixin.clone());
 }
 
+function seedMixinTree(mixin: Mixin, descriptor: MixinDescriptor): void {
```
edloidas, line 434: We seed defaults into the draft only (not persisted), so the content stays dirty and the defaults get persisted on the next save.

This holds only while `seedFormDefaults` writes exactly what the form2 renderer writes on mount: since the values already match, mounting the tab fires no tree change, so the mount-time `snapshotMixinBaseline` leaves our dirty state untouched.

If the two ever drift apart, opening a seeded tab would fire a change, the snapshot would copy draft -> persisted, and the seeded defaults would silently become not-dirty (and never get saved). Worth a regression test: seed a default, mount the tab, assert the mixin is still dirty.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -803,36 +824,59 @@ export function setContentType(contentType: ContentType): void {
...
-    seedMandatoryMixins();
+    seedMixinDefaults();
+}
+
+/**
+ * Fetches the given applications' x-data (by app key) and seeds their mandatory
+ * defaults into the draft, so a SiteConfigurator selection persists them in the
+ * same save.
+ */
+export async function seedMixinsForApplications(applicationKeys: string[]): Promise<void> {
```
edloidas: 🤖 Suggestion: this runs async I/O (`sendAndParse`) directly in a store file. Per `stores.md`, async work should live in a service that subscribes to a trigger and populates the fact store - same pattern we already use for other services (e.g. `wizardMixins.service.ts`). Consider moving `seedMixinsForApplications` there.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.test.ts
```diff
@@ -555,6 +604,48 @@ describe('wizardContent.store', () => {
         expect($wizardHasChanges.get()).toBe(true);
     });
 
+    it('seeds defaults into the draft of an already-attached mandatory mixin and marks it dirty', () => {
```
edloidas, line 589: 🤖 Note: these new tests skip the `should` prefix that `testing.md` asks for (e.g. `should seed defaults into the draft...`). It matches the existing tests in this file, so not blocking - flagging it for whenever we align the naming.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
ashklianko: Agree, should not be here at all

edloidas (approved):

PR #10860, opened 2026-06-17
Delete dialog won't close on Esc #10738

### LINKED ISSUE #10738: Delete dialog won't close on Esc
1. Open Delete dialog for a content that has inbound references
2. Press Esc

"Delete" button gets focused behind the dialog, but the dialog won't close.

skjulestad:
Focus is not set in the dialog when the delete button is disabled, this is the case when the inbound ref exists. Set focus to ignore the inbound ref button in these cases and clear names for each ref.

### DIFF
--- a/modules/app/pnpm-lock.yaml
+++ b/modules/app/pnpm-lock.yaml
@@ -184,8 +184,8 @@ importers:
         specifier: ^1.11.21
         version: 1.11.21
       dompurify:
-        specifier: ~3.4.8
-        version: 3.4.9
+        specifier: ~3.4.11
+        version: 3.4.11
       fine-uploader:
         specifier: ^5.16.2
         version: 5.16.2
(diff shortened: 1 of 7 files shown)


ashklianko (changes requested):
The code is correct and solving the issue in Delete dialog, but:

The same focus pattern (onOpenAutoFocus → preventDefault() + focus() on a possibly-disabled action button) is used in UnpublishDialogMainContent and DuplicateDialogMainContent. They share the same latent bug — focusing a disabled button is a no-op, so focus stays outside the dialog's focus scope and Esc stops working. It just doesn't surface there because their action button is only disabled transiently while loading (!ready), so focus self-heals once ready flips. Delete is the only dialog where the action stays disabled persistently (inbound refs), which is why the bug is reproducible here. Worth considering a shared fix to the focus pattern rather than a per-dialog patch. 

Try throttling network, open Duplicate and press ESC - focus shifts to the action toolbar, you can open another dialog

skjulestad:
This PR is closed, issue solved in npm-enonic-ui: https://github.com/enonic/npm-enonic-ui/issues/481

PR #10866, opened 2026-06-17
Disable inherited non-localized content editing #10728

### LINKED ISSUE #10728: Disable inherited non-localized content editing

ashklianko:
# EditLockOverlay — read-only locking

## What
Added a shared `EditLockOverlay` and used it to lock the wizard's editable surfaces when the content is read-only (`$wizardReadOnly`): ContentForm, mixin tabs (MixinView), the page tab tree (PageComponentsView, incl. detached), and the page inspect panel (PageInspectionPanel). The rename button in the toolbar is disabled in the same state.

## How
- `EditLockOverlay` wraps content in a `relative isolate` container, applies `inert` to the children when `locked`, and paints a single veil (`bg-surface-neutral/70`) on top.
- `inert` blocks all interaction (focus/click/input); the veil is the only visual dimming layer.
- Removed per-input dimming under the mask: dropped `enabled={!readOnly}` from the wrapped `FormRenderer`s and `disabled:opacity-50` from DisplayNameInput.
- `isolate` scopes the mask's z-index so it can't cover sibling panels (e.g. the context panel).
- Rename: `canRename = canRenameContentPath && !readOnly` gates the toolbar path button.

## Why
- One veil instead of per-input `disabled` styling — avoids double-dimming (mask opacity stacking on already-dimmed inputs) and gives a uniform, theme-aware read-only look.
- Single reusable component keeps the behavior consistent across all surfaces and driven by one source of truth (`$wizardReadOnly`).

## Concerns
- `inert` fully freezes the content: no scrolling, expanding sets, text selection, and it drops the form from the a11y tree — read-only is "frozen", not "viewable".

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/EditLockOverlay.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/EditLockOverlay.tsx
@@ -11,7 +11,7 @@ const EDIT_LOCK_OVERLAY_NAME = 'EditLockOverlay';
 
 export const EditLockOverlay = ({locked, contentClassName, className, children, ...rest}: EditLockOverlayProps): ReactElement => {
     return (
-        <div data-component={EDIT_LOCK_OVERLAY_NAME} className={cn('relative isolate', className)} {...rest}>
+        <div className={cn('relative isolate', className)} {...rest} data-component={EDIT_LOCK_OVERLAY_NAME}>
             <div className={contentClassName} inert={locked}>
                 {children}
             </div>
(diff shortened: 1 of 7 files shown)


edloidas (approved):

alansemenov:
Looks fine

alansemenov (approved):

PR #10871, opened 2026-06-18
SortDialog missing functionality #9882

### LINKED ISSUE #9882: SortDialog missing functionality
Sort dialog is missing the following functionality that must be implemented:
- [ ] Content must be loaded from content store from cache.
  - [ ] Content must be updated on server events.
- [ ] We must load IDs first and show skeleton for content.
- [ ] Load must be done in batches.
- [ ] List must be rendered in virtualized list

ashklianko:


edloidas (approved):

edloidas:
Looks better now, but we still lack virtualization. Rendering large amount DOM nodes would make the rendering slower and may produce visible lags on 1000+ items. This must also be fixed, but probably in a different issue

ashklianko:
Left virtualizaion out of it on purpose, we can have conflicts between dnd kit and virtualization lib, better to be done in a separate task

alansemenov:
> Left virtualizaion out of it on purpose, we can have conflicts between dnd kit and virtualization lib, better to be done in a separate task

@ashklianko can you create a task for this?

PR #10873, opened 2026-06-18
DND does not work in Image / Media selectors #10301

### LINKED ISSUE #10301: DND does not work in Image Selector
1. Open wizard with Image Selector 
2. Select an image using drag and drop

Image

### **BUG  - Drag and drop does not add a new item in the image selector**

reisfmb:


edloidas (approved):
Works great! Though I'm not really sure if dnd should look they way it looks now, as it probably should be more unified to drop zones similar to New Content Dialog and drop zones in Page Editor. But I suggest we do it separately if it's really required. We still have no design for this.

PR #10875, opened 2026-06-18
Attachment validator #10776

### LINKED ISSUE #10776: Attachment validator

ashklianko:
`AttachmentValidationError` carries only a file name (no property path), so it never fit the path-based form2 channel and was dropped. It now goes to a dedicated `$attachmentServerErrors` channel; each `AttachmentUploader` matches errors against the file names it shows and renders them inline via `FieldError`. While any attachment error is present the content tab is invalid and the wizard validity flag is cleared; the error clears on removal and a fresh set arrives on the next save. Also renamed the data-error channel (`$contentServerErrors` → `$dataServerErrors`) to disambiguate it from the attachment and component-config channels.

NB: this pull only fixes regression, we used to show this error on AttachmentUploader in legacy code. Further improvements, like showing the list of attachment errors in a separate area, to be discussed/done in a separate task.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/attachment-uploader/AttachmentUploaderList.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/attachment-uploader/AttachmentUploaderList.tsx
@@ -1,5 +1,6 @@
 import {type ReactElement} from 'react';
 import {cn, GridList, IconButton, Link} from '@enonic/ui';
+import {FieldError} from '@enonic/lib-admin-ui/form2';
 import {X} from 'lucide-react';
 import {getCmsApiUrl} from '../../../../utils/url/cms';
 
@@ -8,24 +9,29 @@ export type AttachmentUploaderListProps = {
     contentId: string | undefined;
     onRemove: (index: number) => void;
     disabled: boolean;
+    errors?: ReadonlyMap<string, string>;
 };
 
-export const AttachmentUploaderList = ({contentId, names, onRemove, disabled}: AttachmentUploaderListProps): ReactElement => {
+export const AttachmentUploaderList = ({contentId, names, onRemove, disabled, errors}: AttachmentUploaderListProps): ReactElement => {
     if (names.length === 0) return null;
 
     return (
         <GridList disabled={disabled} className="rounded p-2 -m-2">
             {names.map((name, index) => {
                 const url = getCmsApiUrl(`media/${contentId}/${encodeURIComponent(name)}`);
+                const error = errors?.get(name);
 
                 return (
                     <GridList.Row key={name} id={index.toString()} className={cn('flex items-center justify-between gap-2 py-1')}>
                         <GridList.Cell>
-                            <GridList.Action>
-                                <Link className="flex items-center gap-2 truncate" href={url} target="_blank">
-                                    {name}
-                                </Link>
-                            </GridList.Action>
+                            <div className="flex flex-col gap-0.5 min-w-0">
+                                <GridList.Action>
+                                    <Link className="flex items-center gap-2 truncate" href={url} target="_blank">
+                                        {name}
+                                    </Link>
+                                </GridList.Action>
+                                {error && <FieldError message={error} className="text-sm" />}
+                            </div>
                         </GridList.Cell>
                         <GridList.Cell>
                             <GridList.Action>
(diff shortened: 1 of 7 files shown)


modules/lib/src/main/resources/assets/js/v6/features/store/wizardValidation.store.ts
```diff
@@ -334,30 +341,45 @@ export function setServerValidationErrors(errors: ValidationError[]): void {
     const componentErrors = errors.filter(
         (e): e is ComponentConfigValidationError => e instanceof ComponentConfigValidationError,
     );
-    $contentServerErrors.set(contentErrors);
+
+    const attachmentErrors = errors.filter(
+        (e): e is AttachmentValidationError => e instanceof AttachmentValidationError && !isSystemError(e),
```
edloidas: ✴️ The attachment channel copies the `!isSystemError(e)` guard from the data channel, but the rationale doesn't carry over. For data fields, dropping `system:` errors is safe because client-side required/occurrence validation re-surfaces them with clearer messages — there's no equivalent client-side check for attachment content in v6 (the only `attachmentInvalid` reference is the test asserting the drop). So a built-in rejection like `system:cms.attachmentInvalid` is dropped entirely: it isn't shown inline (the regression this PR restores, still broken for the system case), and because `hasAttachmentErrors` stays `false` the content tab reads valid and the user can re-save a server-rejected attachment with no feedback. Legacy `AttachmentUploader.doValidate()` showed all attachment errors — `getAttachmentErrors()` filters only on `getAttachment()` being truthy, with no system filter. Built-in media validation is the common case (custom-app validators carry app keys and are shown), so this is likely to hit in practice. We should drop the `isSystemError` check on the attachment channel, or filter by specific error code rather than the blanket `system:` prefix if some codes really are covered client-side.

modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/attachment-uploader/AttachmentUploaderList.tsx
```diff
@@ -8,24 +9,29 @@ export type AttachmentUploaderListProps = {
...
 
     return (
         <GridList disabled={disabled} className="rounded p-2 -m-2">
             {names.map((name, index) => {
                 const url = getCmsApiUrl(`media/${contentId}/${encodeURIComponent(name)}`);
+                const error = errors?.get(name);
 
                 return (
-                    <GridList.Row key={name} id={index.toString()} className={cn('flex items-center justify-between gap-2 py-1')}>
+                    <GridList.Row key={`${name}-${index}`} id={index.toString()} className={cn('flex items-center justify-between gap-2 py-1')}>
```
edloidas: ✴️ `key={`${name}-${index}`}` ties the row key to position. Committed values are deduped in `handleFiles` (it skips `isDuplicate`), so names are already unique within an uploader and the previous `key={name}` was stable — folding in `index` means removing a row remounts every row after it. There's also a quiet tension with the error model: `errors.get(name)` matches by name, so if names could ever duplicate both rows would render the same error and the index in the key wouldn't help. Since names are unique, `key={name}` is the cleaner choice.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardValidation.store.ts
```diff
@@ -69,13 +75,14 @@ function runValidation(): void {
 
     const contentResult = validateForm(contentType.getForm(), draftData.getRoot(), {
         rawValues: contentRawValueMap,
-        serverErrors: contentServerErrors,
+        serverErrors: dataServerErrors,
     });
 
     const nextInvalidTabs = new Set<string>();
     const hasInvalidDisplayName = $wizardDraftDisplayName.get().trim().length === 0;
+    const hasAttachmentErrors = $attachmentServerErrors.get().length > 0;
```
edloidas, line 83: ✴️ `hasAttachmentErrors` blocks validity globally, but the inline display is scoped per uploader by file name (`useAttachmentServerErrors`). If an attachment error's file name isn't shown by any uploader, the content tab is invalid and save is blocked with nothing visible to explain why. The normal save flow keeps the file present so it matches — low risk — and the PR already defers the aggregated "errors in a separate area" view to a follow-up. A short comment here noting that deferral would help the next reader not mistake this for complete.

edloidas (changes requested):

modules/lib/src/main/resources/assets/js/v6/features/store/wizardValidation.store.ts
ashklianko: It doesn't block Save or anything else, just presents validation error sign on the content form tab even if we are missing validation on specific inputs. With attachment errors  the content's 'valid' field remains invalid,  without 'hasAttachmentErrors' and invalid icon on the tab nothing would point to validation problem. 
It is a known compromise for this issue.

PR #10886, opened 2026-06-19
Confirmation dialog appears when removing an Item Set with no dirty fields #10859

### LINKED ISSUE #10859: Confirmation dialog appears when removing an Item Set with no dirty fields
1. Open  Optionset wizar (Features)
2. Click on `Option 1` in Single selection 
3. Click on `Add` button  one more `Minimum 3` appears - OK 

Image

4. Expand the menu  and click on `Delete` menu item

Image

**BUG  -  Confirmation dialog appears when removing an Item Set with no dirty fields**

Image

reisfmb:


edloidas (commented):
This one is a bit tricky. The **fix itself is valid** according to the logic described in the issue. But I'm actually concerned about when we should show the dialog.

Previously, the check was all about the dirty state: `!isDirty() && !hasNonDefaultValues()`. So no matter what was in it, if there were changes inside, it was considered dirty.

But right now, we only check whether it is empty, not whether it is dirty. That means it's possible to remove everything from the set and not trigger the dialog, even though it had changes.

I can't say whether the previous behavior was correct, but this definitely changes how we treat those sets.

@ashklianko can you verify the correct behavior for this one?

ashklianko:
Agree with @edloidas that we used to check for 'dirty' values before, but from my perspective it looks like an overkill for item sets: to check entire non-empty form with their default config values just to know if we need to show delete dialog or not requires tons of logic to support. 
And if someone deliberately cleaned the form then I believe the delete is also going to be on purpose.
Looks good to me, can bring back the old logic if some strong arguments given to check for dirty.

ashklianko (approved):

ashklianko:
So to summarize, the new behavior is ok, so now we expect that:
- Any non-empty set will show confirmation dialog after pressing delete, no matter if it is dirty or not
- No confirmation dialog for empty set, no matter if it had values before

PR #10890, opened 2026-06-19
Rename Content dialog validation #10747

### LINKED ISSUE #10747: Rename Content dialog Name input validation is missing
1. Open a site
2. Click  on Open Rename content dialog
3. Insert a invalid name

**BUG   Name input validation is missing**  

Image

Image

Image

Image

skjulestad:
Add validation for name input, match xp's NameValidator.


ashklianko (changes requested):
- Misleading error text: any validation failure shows `path.not.available` ("Unavailable"), so the highlighted problem looks like a taken name rather than a malformed one. Add a dedicated invalid-name phrase and show it instead.

- Leading/trailing-space rejection vs trim-on-submit: the dialog submits `value.trim()`, so it transiently errors on a trailing space for a name it would submit as valid. Let's allow it. Also not looking good when typing a name like 'blog post' and when pressing space seeing error for a moment.

- `setRenameContentDialogValue` calls `.set()` twice on the invalid branch; `availabilityStatus: undefined` could go in the first write.

ashklianko (approved):

PR #10899, opened 2026-06-23
Regression: tags autocompletion #10882

### LINKED ISSUE #10882: Regression: tags autocompletion
According to [our docs](https://developer.enonic.com/docs/cms/xp8/schemas/form-items/tag):
"Tags support autocompletion based on tags defined in other items with the same property path. A tag may consist of spaces and special characters, but leading and trailing whitespace will be trimmed." 

This functionality is missing in the new UI. Check implementation details and conditions on which autocomplete was suggested in legacy and replicate in the new Tags component.

Image

skjulestad:
Add tag input suggestions with scoped allowPath support

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/index.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/index.ts
@@ -8,6 +8,7 @@ import {SiteConfiguratorDescriptor, SiteConfiguratorInput} from './site-configur
 import {AttachmentUploaderDescriptor, AttachmentUploaderInput} from './attachment-uploader';
 import {CustomSelectorDescriptor, CustomSelectorInput} from './custom-selector';
 import {ImageUploaderDescriptor, ImageUploaderInput} from './image-uploader';
+import {TagDescriptor, TagInput} from './tag';
 
 export function registerContentStudioInputTypes(): void {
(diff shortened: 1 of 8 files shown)


modules/lib/src/main/resources/assets/js/v6/features/shared/form/input-types/index.ts
```diff
@@ -19,4 +20,5 @@ export function registerContentStudioInputTypes(): void {
     InputTypeRegistry.registerType({mode: 'internal', descriptor: AttachmentUploaderDescriptor, component: AttachmentUploaderInput}, true);
     InputTypeRegistry.registerType({mode: 'internal', descriptor: CustomSelectorDescriptor, component: CustomSelectorInput}, true);
     InputTypeRegistry.registerType({mode: 'internal', descriptor: ImageUploaderDescriptor, component: ImageUploaderInput}, true);
+    InputTypeRegistry.registerType({mode: 'internal', descriptor: TagDescriptor, component: TagInput}, true);
```
edloidas, line 25: Probably worth adding a note, that we use a modified variant of a Tag here and forcefully re-register it, even though it was registered in built in types.

edloidas (changes requested):
I'm a bit confused here

edloidas (approved):

PR #10929, opened 2026-06-26
Image editor: line thickness doesn't scale #10928

### LINKED ISSUE #10928: Image editor: line thickness doesn't scale
The crop and focus overlays in the image uploader draw their outlines and crop handles in image-pixel units, since the SVG `viewBox` matches the image dimensions while the element is rendered scaled to fit.

As a result the stroke width does not scale with the displayed size: on large images (e.g. 4000px) a 2px stroke renders too thin, while on small images it looks too thick. The crop handles have the same problem and become too small on large images.

Separately, the outline uses a single red stroke, so when the underlying image is predominantly red the crop/focus line blends into the background and becomes hard or impossible to see.

Example:
Image

reisfmb:
Turns out SVG has an attribute built exactly for this: vector-effect="non-scaling-stroke". Didn't know it existed. With it, the stroke keeps a constant on-screen thickness regardless of the image dimensions — no manual recalculation on our side.

Also, as described in the issue description, the image can happen to be the same color as our line (e.g. a red photo → red crop line vanishes). To cover that, the outline is now two interleaved dashed strokes — white and red alternating — so whatever the background, one of the two always contrasts.


ashklianko (approved):

ashklianko:
The code is ok, let @alansemenov also see how it looks in action before merge

reisfmb:
> The code is ok, let @alansemenov also see how it looks in action before merge

Alan saw it and approved 😉

PR #10935, opened 2026-06-26
Invalid HtmlArea is not highlighted #10927

### LINKED ISSUE #10927: Invalid HtmlArea is not highlighted
Two mandatory fields in the screenshot - TextArea and HtmlArea - the first one is correctly highlighted (red border), the second one is not.

Image

ashklianko:


edloidas (commented):
Why wasn't it working correctly? And why do we keep other similar styles like `[&_.cke_chrome]:!border-transparent`? Are they working correctly?

ashklianko:
The selector was wrong, I think we just missed it.

Put it in the corresponding less file, checked in ff - the fix works there too

edloidas:
The question is: are we planning to get rid of less in the near future? We need to drop all less files in few months, so if those things need to stay, we need to move them to tsx's tailwind, not the other way around.
If it's just a small fix to current styling that will be removed completely soon, then it's okay.

ashklianko:
Replaced with pure tailwind, `String.raw` to preserve underline so tailwind doesn't convert it to space

edloidas (approved):
I get it. Still looks a bit ugly. But since it's part of input type, I can tolerate it.
I suggest that in the future if we need to add some hacks, that's hard to write in Tailwind and that might be removed eventually, we, instead of less, will write it in CSS and import it in our tailwind.css file (we already have short sections with hacks, so having separate layered styles is okay too).

PR #10958, opened 2026-06-30
Previous x-data remains after version revert #8825

### LINKED ISSUE #8825: Errors after reverting a site with site-configurator
1. Open wizard for new site , fill in the name input ,
2. Select `Features` 
3. Remove the selected app
4. Select another app, that contains selectors in its  site config
5. Open Versions and revert previous versions

**BUG** - Edited item is reverted - error appears in console and error-notification message

**There is the similar fixed issue**  - https://github.com/enonic/app-contentstudio/issues/8664

Image

```
main.js:2 TypeError: Cannot read properties of undefined (reading 'getProperties')
    at ImageSelector.getValueFromPropertyArray (main.js:4:726979)
    at ImageSelector.getSelectedItemsIds (main.js:2:1310563)
    at ImageSelector.isDirty (main.js:2:1312810)
    at ImageSelector.update (main.js:2:1312558)
    at InputView.update (main.js:2:273576)
    at main.js:2:568505
    at Array.map (<anonymous>)
    at FormItemLayer.update (main.js:2:568358)
    at FormView.update (main.js:4:116878)
    at main.js:2:360866
```

ashklianko:


ashklianko:
Not dependant but site configurator works properly with https://github.com/enonic/lib-admin-ui/pull/4571.

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -584,6 +584,25 @@ export function onWizardPersistedContentSet(callback: (content: Content) => void
...
+    };
+}
+
+function notifyServerMixinsChanged(): void {
+    for (const callback of serverMixinsChangedCallbacks) {
+        callback();
+    }
+}
+
+function mixinNamesChanged(previous: Set<string>, next: string[]): boolean {
```
edloidas: Maybe adding `has*` would make it better?

edloidas: Also, `next` is an array, and it _might_ contain duplicates, which would lead to `previous.size !== next.length` to be `true`, while second is `false` and in this case it would be incorrect.

And since you are using sets, the less performant version, but more readable would be:
```
const nextSet = new Set(next);
return previous.size !== nextSet.size || !previous.isSubsetOf(nextSet);
```

modules/lib/src/main/resources/assets/js/v6/features/store/wizardContent.store.ts
```diff
@@ -596,12 +615,18 @@ export function applyServerSidePersistedContent(content: Content): ApplyServerCo
...
 
+    const previousMixinNames = new Set($wizardPersistedMixins.get().map((m) => m.getName().toString()));
+
     applyDisplayNameFromServer(snapshot.displayName);
     applyNameFromServer(snapshot.name);
     applyDataFromServer(snapshot.data);
     const syncedMixinNames = applyMixinsFromServer(snapshot.mixins);
     applyWorkflowFromServer(snapshot.workflow);
 
+    if (mixinNamesChanged(previousMixinNames, snapshot.mixins.map((m) => m.getName().toString()))) {
```
edloidas: I'd at least move `mixins.map((m) => m.getName().toString())` to a separate arrow function to make `if` more compact and reuse it in code above.

edloidas (changes requested):

edloidas (approved):

PR #10974, opened 2026-07-01
Incorrect handling of a link to non-existing content in the Insert Link dialog #10939

### LINKED ISSUE #10939: Incorrect handling of a link to non-existing content in the Insert Link dialog
If there's a link to a local content (that looks like this `<a href="content://d8190c73-14ad-4654-aee5-ab83410a3687" title="My link">`) and the content with such id doesn't exist, then the dialog won't handle it and show only the selector's label, making it impossible to unselect the item and select something else. It should be handled in the same way as we do in selected options linking to missing content in a Content Selector.

Image

reisfmb:
Added ContentNotAvailable shared component with the removed-content view (icon + id + label). 
Reused it in MediaSelectorItemView and ContentSelectionItem instead of inline markup. 
ContentSelection now keeps a pathless ContentSummaryBuilder stub for unresolved ids instead of dropping them, so the selector renders a removable row. 
ContentSelectionItem shows ContentNotAvailable and hides the status badge for pathless (removed) items. HtmlAreaLinkDialogContext no longer surfaces an error toast when a linked content is missing; logs to console instead.


ashklianko (approved):

ashklianko:
Good stuff! 
'isRemoved' maybe is not the best name, 'isMissing' is more accurate, but that's fine

PR #10989, opened 2026-07-01
Content Selector is not switching to flat mode when filtered #10984

### LINKED ISSUE #10984: Content Selector is not switching to flat mode when filtered
Not sure at which point it got broken, as it used to be correct. When user enters keywords to filter the selector (Content, Media or Image Selector) it should switch to flat mode regardless of which mode (flat or tree) it currently is in.

ashklianko:


edloidas (approved):

ashklianko:
@sgauruseu tests require updating

PR #10990, opened 2026-07-01
Insert link dialog - Context in content selector is not limited with current site #10874

### LINKED ISSUE #10874: СontentSelector query results are not limited by the nearest site
**Content items with matching site name prefixes are displayed when "Show content from the entire project" is unchecked** 

1. Create 3 sites -  test1, test12, test123
2. Open  the site `test1`
3. Open` Insert link to content` modal dialog
4. Do not select the “Show content from the entire project” checkbox
Image

5. Expand the dropdown options:

**BUG - Content names that start with the same text as the current site should not be present in the options**

Image

reisfmb:
Added opt-in strictAllowedPaths flag threaded from ContentTabPanel through ContentSelector/ContentCombobox to useContentComboboxController, which filters displayed rows to the exact node or its descendants.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/htmlarea-link/ContentTabPanel.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/features/shared/dialogs/htmlarea-link/ContentTabPanel.tsx
@@ -121,6 +121,7 @@ export const ContentTabPanel = (): ReactElement => {
                         onSelectionChange={handleSelectionChange}
                         selectionMode='single'
                         allowedContentPaths={allowedContentPaths}
+                        strictAllowedPaths
                         error={!!errors.content}
                         closeOnBlur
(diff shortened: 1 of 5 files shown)


ashklianko:
Thanks for taking this on!
My one concern is that it addresses the presentation layer rather than the query: the server still matches `/content/<site>*`, a prefix that also picks up sibling sites (`site1` matching `site12`, `site123`), and the extra rows are filtered out on the client after they are fetched. Since the dropdown is server-paginated with "load more" driven by the server's total hit count, filtering after the fetch can leave a page of only sibling content reduced to zero visible rows while still reporting "has more", so the user sees an empty dropdown while matching content sits on later pages that are never reached — and it only covers the already-resolved path here (the `allPlain` guard skips `${site}`), so the regular content selectors that pass `${site}` stay unfixed. 

Check https://github.com/enonic/app-contentstudio/pull/10994

PR #10994, opened 2026-07-02
СontentSelector query results are not limited by the nearest site #10874

### LINKED ISSUE #10874: СontentSelector query results are not limited by the nearest site

ashklianko:
Fixes a cluster of regressions in the v6 content/media/image selectors where allowed-path scoping produced the wrong result set. The v6 rewrite diverged from the legacy inputs in several places; this restores correct scoping and removes an over-matching bug on the backend.

## What

Backend

- `ContentRelativePathResolver`: the bare `${site}` wildcard now resolves to the site subtree (`/<site>/*`) instead of a bare prefix (`/<site>*`), which previously also matched sibling sites sharing a name prefix (opening a selector in `site1` matched `site12`, `site123`).
- `ContentSelectorQueryJsonToContentQueryConverter`: for the bare `${site}` wildcard the query now also matches the site node itself, producing `(_path LIKE '/content/<site>' OR _path LIKE '/content/<site>/*')`; when the content is outside any site it degrades to the whole repository.

Frontend

- `useSelectorInput`: the context content is now always passed to the request. Previously it was dropped when `allowPath` was only `${site}`, so the backend could not resolve `${site}` and the selector silently listed the whole repository.
- Descriptor defaults: `ContentSelector` keeps the site-scoped default (`${site}`), while `MediaSelector` and `ImageSelector` default to no restriction (whole repository), matching the legacy behavior. All three previously defaulted to `${site}`, so media/image outside the current site was hidden.
- `htmlarea-link` content tab: the selector now sends the `${site}` placeholder plus the context content and lets the backend resolve it, instead of resolving the nearest site on the client and sending a concrete `/site` path (which reintroduced the sibling-prefix over-match). The nearest-site lookup is kept only as an "is inside a site" flag that drives the "show all content" checkbox.

Refactor

- Extracted `readAllowPath(raw, fallback)` so the three descriptors share one parser and differ only by an explicit fallback argument, removing the copy-paste that caused the default divergence.
- Extracted `resolveAllowedContentPaths` from the link dialog into a pure, testable function.

## Why

The v6 selectors are a rewrite of the legacy inputs, and several behaviors were carried over imperfectly: media/image inherited the content selector's site-scoped default, the context content was conditionally dropped, and `${site}` was resolved on the client into a prefix match that leaked into sibling sites. Together these made selectors show the wrong content — either the whole repository when they should be scoped to the site, or sibling sites when scoped. The fixes move all path resolution to the backend, where the logic already lives, and make each selector's default explicit instead of implicit.

## Tests

- Backend: added converter cases for `${site}` (site node plus subtree), `${site}` without a site (whole repository), and `${site}/*` (descendants only); updated the resolver test for the site subtree.
- Frontend: added descriptor default tests (content versus media/image), a `useSelectorInput` test pinning that context content is always sent, and unit tests for `readAllowPath` and `resolveAllowedContentPaths`.

Closes #10874

### DIFF
--- a/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentRelativePathResolver.java
+++ b/modules/app/src/main/java/com/enonic/app/contentstudio/rest/resource/content/ContentRelativePathResolver.java
@@ -17,7 +17,14 @@ public class ContentRelativePathResolver
 
     public static String resolveWithSite( final String pathExpression, final Site parentSite )
     {
-        return makeEndWithStar( makeStartWithSlash( resolveSitePath( pathExpression, parentSite ) ) );
+        final String resolved = makeStartWithSlash( resolveSitePath( pathExpression, parentSite ) );
+
+        if ( SITE_WILDCARD.equals( pathExpression ) )
+        {
+            return makeEndWithStar( makeEndWithSlash( resolved ) );
+        }
+
+        return makeEndWithStar( resolved );
     }
 
     public static boolean anyPathNeedsSiteResolving( final List<String> allowedPaths )
@@ -104,6 +111,18 @@ public class ContentRelativePathResolver
         else
         {
             return "/" + str;
+        }
+    }
+
+    private static String makeEndWithSlash( final String str )
+    {
+        if ( str.endsWith( "/" ) )
+        {
+            return str;
+        }
+        else
+        {
+            return str + "/";
         }
     }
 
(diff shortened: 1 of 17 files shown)


reisfmb (approved):
Good catch and fix Sasha!

reisfmb:
@ashklianko why was it closed? 🤔

ashklianko:
@reisfmb I needed to updated PR after the latest archtectural changes, pushed it into another branch. https://github.com/enonic/app-contentstudio/pull/11006#event-27523977578

PR #11001, opened 2026-07-02
Migrate ui-tests for CompareVersionsDialog and FullscreenDialog #11000

### LINKED ISSUE #11000: Migrate ui-tests for  CompareVersionsDialog and FullscreenDialog

sgauruseu:


ashklianko:
@sgauruseu please rebase it and squash: need to check if it works with the latest archtecture changes

One more note: when you need to update something in a pull please use 'amend' option in your commits and the force push - thus we get 1 commit per pull. Ofc it is possible to squash even with multiple commits in PR, but when we have several commits:
a) easy to miss that we have multiple commits in the pull and press rebase instead of squash 
b) it saves a little time and effort when we there's just 1 commit in the pull

ashklianko (approved):

Issue #10976, opened 2026-07-01
Can't upload a project icon


alansemenov:
After migration to the new UI we have lost a possibility to add/replace a project icon. In the legacy mode, it used to be possible via a click on project icon next to the display name input in the edit form. But now we don't have edit form for projects anymore.

In the new version uploading a project icon should work via click on the project icon in the SettingsItemStatistics panel on the right (see the screenshot). Add a tooltip and a new i18n phrase key: "Click to upload a project icon". Use the same API for upload, as before.

Important: the node (or just its icon) in the grid should get updated after the new icon is uploaded.

<img width="1069" height="401" alt="Image" src="https://github.com/user-attachments/assets/7df8035b-c036-4567-ada3-16abdbc4680a" />

sigdestad:
NB. Remember to also add a way to reset the icon if it has been customized.

alansemenov:
> NB. Remember to also add a way to reset the icon if it has been customized.

Never had it, neither for project nor for content icons. So it's a separate feature, that needs design.

PR #11005, opened 2026-07-02
Can't upload a content icon #10975

### LINKED ISSUE #10975: Can't upload a content icon
After migration to the new UI we have lost a possibility to add/replace a content icon. In the legacy mode, it used to be possible via a click on content type icon next to the display name input in the edit form.

In the new version it should work in exactly the same way, but via click on the content type icon in the Details widget (see the screenshot). Add a tooltip and a new i18n phrase key: "Click to upload a content icon". Use the same API for upload, as before.

Important: the node (or just its icon) in the grid should get updated after the new icon is uploaded.

Image

reisfmb:


reisfmb:
@ashklianko FYI: the thumbnail removal was [outside of the scope](https://github.com/enonic/app-contentstudio/issues/10976#issuecomment-4863471219) of this and the project icon tasks (as both depends on design decisions)

edloidas (approved):

PR #11041, opened 2026-07-10
Media tab should be disabled if media content is not allowed #10372

### LINKED ISSUE #10372: Suggested and Media tabs should be disabled if the content type does not exist
1. Specify a non-existent content type in the type configuration:
 ```
allowChildContentType:
- "acme"
```

2. Select the content and open `New Content Dialog`

**BUG** : **Media and Suggested tabs should be disabled** in  `New Content Dialog`
 'No content types found' -  should be shown in 'All '  tab
     

**A similar situation occurs when non-media child items are allowed**.  Media tab should be disabled
```
allowChildContentType:
- "article-non-media"
```

Image

reisfmb:


alansemenov (changes requested):
It works as expected, but can you do a small followup fix: please disable "New" button when a page template is selected. It doesn't allow any content types or media, so there's no point in allowing the New content dialog for it.

Image

Image

alansemenov (approved):

PR #11074, opened 2026-07-16
Details Panel is not updated after updating the page template #10804

### LINKED ISSUE #10804: Details Panel is not updated after updating the page template
1. Create a site. Select the site
2. Add new page template. Select `site` in support, insert a part.

**BUG  - Content Item Statistics Panel should be updated after creating the template.**  **DetailsWidgetTemplateSection** should be displayed as well

Image

Image
3. Refresh  the Browse Panel  -   DetailsWidgetTemplateSection gets visble

4 Remove `site` in support selector in the template. 
5. Switch to Browse Panel .

**BUG  -   Statistics Panel should be updated as well** the part should not be displayed in the preview

Image

### **CASE 2   - Delete the page template** 

**BUG  -   Statistics Panel is not updated after deleting the template**

reisfmb:


alansemenov (changes requested):
It does work correctly if you open the Details panel _after_ the template was created/deleted, but not if the panel was already open - see the video below. Unrelated to this bug, but I also noticed that after a template is deleted we don't refresh the editor (we do when a template is added), can you fix that as well please?

https://github.com/user-attachments/assets/a1633823-c105-4590-9093-9d65ea814134

alansemenov:
Sorry, I was on a different sandbox when I tested this, so my comment above was for a version without the fix :)

Correction: In browse mode (in the grid) it now works perfectly fine. It also refreshes the preview both when a template is update and deleted. But in the editor it's as I wrote above - if the details panel is already open before the template changes then nothing will happen, and editor won't refresh on template delete. Please fix so that it works 100% the same in both edit and browse modes.

alansemenov (approved):

PR #11075, opened 2026-07-16
Improve styling of option form in an option-set #11073

### LINKED ISSUE #11073: Incorrect styling of option form in an option-set
Add a left-border to the option form, identical to what we have for a field-set.

Now:

Image

Should be:

Image

Same for a single-select:

Image

Note that the line is aligned with the center of its option's checkbox/radiobutton.

reisfmb:


alansemenov (approved):
Looks fantastic

PR #11110, opened 2026-07-20
Incorrect behaviour of the project toggler control #11071

### LINKED ISSUE #11071: Incorrect behaviour of the project toggler control
Correct behaviour of the project toggler:
1. It should not be clickable and the arrows icon should not be visible if there's just one project available
2. We should listen to project create and project delete events on the server and:
- remove "click" event, hover effect and the arrows icon if some projects were deleted and there's just one left
- add "click" event, hover effect and the arrows icon if a project was added and now there's more than one

Image

Image

skjulestad:
Add $hasMultipleProjects to project.store.ts
Disable switch project button along with icon and aria-label when projects < 2

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/entities/project/index.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/entities/project/index.ts
@@ -9,6 +9,7 @@ export {
     onActiveProjectChanged,
 } from './activeProject.store';
 export {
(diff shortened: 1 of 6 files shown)


alansemenov:
Tested - works as expected

reisfmb (approved):
👍

edloidas (approved):

PR #11113, opened 2026-07-20
Guillotine extension won't open on a clean instance #11072

### LINKED ISSUE #11072: Guillotine extension won't open on a clean instance
1. Install Guillotine 8
2. Make sure `enonic:cs:lastselectedprojectid` is not set in LocalStorage, delete if it is
3. Open CS, don't switch the project
4. Click the Guillotine extension icon in the left toolbar.

Bug: `enonic:cs:lastselectedprojectid` is set only when a project is clicked in the project selection dialog. It should be set right when the CS is opened, with the id of the current project. Guillotine widget won't open without `enonic:cs:lastselectedprojectid` in LocalStorage.

Image

Image

skjulestad:
Persisted the resolved active project during project loading. Added regression coverage for clean loads and updated storage expectations.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/entities/project/projects.store.test.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/entities/project/projects.store.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { errAsync, okAsync } from 'neverthrow';
-import { AppError } from '../../shared/api/errors';
+import {AppError} from '../../shared/api';
 
 type MockProject = {
     getName(): string;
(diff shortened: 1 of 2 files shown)


reisfmb (commented):
@skjulestad 

Wdyt of extracting [this](https://github.com/enonic/app-contentstudio/pull/11113/changes#diff-0875b6483f42fa2d1d2de585e6c2306b4c1a3370b168f4255ab41b779f90cde1R211-R214) into a function named `persistActiveProjectId`, and then calling it after the `updateActiveProject` inside the `loadProjects`?

That would follow the existing pattern there: load the projects, then trigger the dependent follow-up actions as named function calls.

edloidas (approved):

reisfmb (approved):
👍

PR #11114, opened 2026-07-20
'Show entire content' doesn't work when the versions are identical #11036

### LINKED ISSUE #11036: "Show entire content" doesn't work when the versions are identical
Most likely they are not identical but the fields that differ are hidden. In any case the checkbox should always work.

https://github.com/user-attachments/assets/9e535bf9-3f29-4870-b4f5-d90cbc349fdc

reisfmb:


modules/lib/src/main/resources/assets/js/v6/widgets/context-panel/widget/versions/compare/CompareVersionsDialog.tsx
```diff
@@ -129,10 +124,14 @@ export const CompareVersionsDialog = ({
...
-                                    dangerouslySetInnerHTML={{ __html: diffHtml }}
-                                />
+                                <>
+                                    {isEmpty && (
+                                        <h3 className={cn('text-sm', showAllContent && 'mb-2')}>{versionsIdenticalLabel}</h3>
+                                    )}
+                                    {(!isEmpty || showAllContent) && (
+                                        <div className={deltaClasses} dangerouslySetInnerHTML={{ __html: diffHtml }} />
+                                    )}
+                                </>
```
edloidas: Let's keep all on level 1, without nested `if`'s.

modules/lib/src/main/resources/assets/js/v6/widgets/context-panel/widget/versions/compare/CompareVersionsDialog.tsx
edloidas: This one can be moved to the `buildDiffHtml`.  There is no need for that additional memo here. Result will be memorized anyway.

edloidas (changes requested):

alansemenov (approved):

Issue #9235, opened 2025-10-14
Incorrect behaviour during bulk selection in content grid using Shift+Arrow down


sgauruseu:
**Case 1** 
1. Expand the `Large tree node 2`
2. Click on the **checkbox** in the first item
3. Try to select child items by pressing `Shift+Arrow Down`.

**BUG**  -  Only some items from “Large tree node 2” and the remaining grid items are selected

https://github.com/user-attachments/assets/036517ba-0542-48e3-bab2-e887c194ab69

 
**Case 2** 

1. Expand the `Large tree node 2`
3. Select the first child item.
4. Try to select child items by pressing `Shift+Arrow Down`.
**### BUG** 
 The browser panel automatically scrolls when these keys are pressed. At some point, due to scrolling too quickly, child elements stop selecting,     but the folders in root directory  become  selected in the grid.

https://github.com/user-attachments/assets/a863a216-6e06-4032-853f-e3a9e8b1abea

PR #11142, opened 2026-07-23
Shift+Click range selection doesn't work in content grid

### LINKED ISSUE #10944: Reload doesn't expand/reselect tree nodes
Clicking the reload button in `TreeListToolbar` resets the entire tree state instead of refreshing data in place. `handleReload` calls `resetTree()`, which replaces the tree with an empty state and discards all expansion state, so every expanded node collapses back to the root level after a refresh. The multi-selection set is dropped entirely and is never restored. The active item is not reliably preserved either: `restoreActiveAfterReload` only re-applies it when the node still exists in the freshly loaded set, which after a reset contains root IDs only, so any active item that lived inside an expanded branch fails the check and `setActive(null)` is called. Because the controlled active id then leaves the items array, `VirtualizedTreeList` falls back to `items[0]` (see the note in `contentTreeSelection.store.ts`), and the first row ends up active/highlighted after every refresh regardless of what the user had selected. As a result, a user who has expanded several branches and selected items loses all of that context on reload and is left with the first item selected. Reload should refresh the content data while keeping previously expanded nodes open, preserving the current selection, and keeping (or clearing) the active item rather than jumping to the first row — dropping only items that no longer exist after the refresh.

#### Rationale

Reload is meant to fetch fresh data, not to reset the user's navigation context. Collapsing the tree, discarding the selection, and silently activating the first item makes the action disruptive and surprising, especially in large trees where re-expanding and re-selecting is tedious.

### LINKED ISSUE #11053: Shift+Click range selection doesn't work in content grid
1. Select the first row in the grid
2. Hold down the `Shift` key, and then click the 5th item in the sequence.

**Expected** : 
All items between the two clicks will be selected,

Image

**Actual:**    only one item gets selected and text is highlighted instead of rows

Image

ashklianko:


reisfmb:
@ashklianko 

_(For consistency, I'll name Shift + Click range selection as SCRS.)_

By manual testing the fix, I noticed this:
- SCRS via the row **checkbox** does not range-select;
- If I SCRS rows 10 - 15 and then still holding shift I press shift+down, I was expecting to get selection 10 - 16, but what I get is a shift 11-16. Symmetrical for shift+up, i.e, after SCRS and then shift+up I get 9-14.

Seems to me that those would be good improvements... Wdyt?

ashklianko:
> @ashklianko
> 
> _(For consistency, I'll name Shift + Click range selection as SCRS.)_
> 
> By manual testing the fix, I noticed this:
> 
> * SCRS via the row **checkbox** does not range-select;
> * If I SCRS rows 10 - 15 and then still holding shift I press shift+down, I was expecting to get selection 10 - 16, but what I get is a shift 11-16. Symmetrical for shift+up, i.e, after SCRS and then shift+up I get 9-14.
> 
> Seems to me that those would be good improvements... Wdyt?

1. Same as in Legacy, no changes here
2. There's an issue opened for shift+up/down https://github.com/enonic/app-contentstudio/issues/9235, can be fixed there

reisfmb (approved):

PR #11143, opened 2026-07-23
Wait for input to be clickable before setValue in typeTextInInput

edloidas:
Guarded the shared `typeTextInInput` page-object helper with `waitForClickable` before `setValue`. Since `setValue` clears the input first, a not-yet-expanded panel or in-flight animation could throw `element not interactable` — this surfaced as an intermittent failure in the `testInputTypes_2` Selenium suite (`datetime.config.spec`) when filtering in the Browse Filter Panel. The guard waits up to `mediumTimeout` for the input to become interactable, fixing this class of flaky failures across every text input in the suite.

### DIFF
--- a/testing/page_objects/page.js
+++ b/testing/page_objects/page.js
@@ -121,6 +121,9 @@ class Page {
     async typeTextInInput(selector, text) {
         try {
             let inputElement = await this.findElement(selector);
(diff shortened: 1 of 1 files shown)


sgauruseu (approved):

ashklianko:
Not sure if it helped - 1st run attempt still failed:
image

ashklianko (approved):

PR #11148, opened 2026-07-24
Page elements are clickable in the editor when preview is done via controller mapping

### LINKED ISSUE #11147: Page elements are clickable in the editor when preview is done via controller mapping
When page is using a page template, we lock it and display context menu on click with shortcut to the Inspect panel.
When page is customised/editable, we allow DnD of page components.

However, when page is rendered via controller mapping (like [here](https://github.com/enonic/app-hmdb/blob/master/src/main/resources/cms/site.yaml#L4-L8)) the page actually becomes clickable as a regular HTML page, including `<a>` links etc. It should behave in the same way as for page template - lock the page, display context menu on click with shortcut to the Inspect panel.

alansemenov:


modules/lib/src/main/resources/assets/js/app/wizard/page/LiveEditPageProxy.ts
```diff
@@ -5,43 +5,43 @@ import {
...
+import { LiveEditPageInitializationErrorEvent } from '../../../page-editor/event/LiveEditPageInitializationErrorEvent';
+import { IFrameEl } from '@enonic/lib-admin-ui/dom/IFrameEl';
+import { DragMask } from '@enonic/lib-admin-ui/ui/mask/DragMask';
+import { WindowDOM } from '@enonic/lib-admin-ui/dom/WindowDOM';
+import { CONFIG } from '@enonic/lib-admin-ui/util/Config';
+import { ContentUrlHelper } from '../../util/ContentUrlHelper';
+import { type Extension } from '@enonic/lib-admin-ui/extension/Extension';
+import { type WizardExtensionRenderingHandler } from '../WizardExtensionRenderingHandler';
+import { setLiveEditDraggableHost } from './LiveEditDraggableHost';
+import { type PortalComponentType } from '../../../v6/widgets/inspectors/model/page-editor/drag';
```
pmi, line 44: Did you forget to run auto-format before commit ?

pmi (approved):

modules/lib/src/main/resources/assets/js/app/wizard/page/LiveEditPageProxy.ts
edloidas: Not really. We added formatting tools (`oxfmt`) alongside linting (`oxlint`). They will run as a pre-commit hook, so they'll update only the files that are modified.

https://github.com/enonic/app-contentstudio/blob/master/.oxfmtrc.json

This makes formatting consistent and removes our reliance on editors built-in formatting tools, especially since we all use lots of them: IntelliJ IDEA, VS Code, Cursor, and Zed.

`oxfmt` also provides significantly more formatting options and is very fast, especially compared with Prettier or any built-in option. It can be introduced across our codebase without causing noticeable disruption.

The spaces in imports are probably the most visible difference from our previous formatting style. However, most of our newer code already follows this approach. Since much of the older code will eventually be rewritten or removed, we decided to keep the new rules simple and consistent.

As for linting rules, we decided to keep them minimal, follow the default recommendations. Aslo, avoid carrying over our legacy ESLint rules, even though oxlint is compatible with ESLint rules.

edloidas: We didn't run the formatter across the entire codebase to avoid massive changes and unnecessary problems with open PRs. So seeing such things is expected. However, we could format the whole codebase during the summer, when we usually have only zero to two active PRs.

PR #11173, opened 2026-07-28
Issue details: pre-stage non-mandatory items selection change #11166

### LINKED ISSUE #11166: Issue details dialog: no Apply button after toggling 'Include child items'
Contination of https://github.com/enonic/app-contentstudio/issues/11046. Create Issue dialog already shows Apply button on selection change, need to do the same in Issue details dialog too for consistency

reisfmb:
- Extracted the draft/applied selection logic into a shared helper and moved all three dialogs onto it
- Staged the Issue Details children and dependant toggles behind the Apply banner instead of saving per click 
- Protected the staged draft from reloads, keeping server changes authoritative on conflict 
- Showed the Apply banner for Task issues and gated Publish while an edit is staged 
- Added helper and store tests; enabled the E2E Apply assertion


ashklianko (changes requested):
Almost there, but there's a strange flickering when pressing Apply in issue details dialog, in the network tab I see that requests have been sent twice:
https://github.com/user-attachments/assets/35a721e0-4c2c-45aa-aa1a-deb8a15b3aab

reisfmb:
Apply ran `syncPublishDialogContext` several times, once per store write, plus again later when the issue server event echo refetched items.

Each run tore the publish context down and refetched it, blanking the footer and doubling requests.

The guard now lives in the function that owns the teardown, so the noisy effect dependencies are harmless, and the echo no longer repeats a reload the save already did.

ashklianko (approved):

PR #11222, opened 2026-08-04
Show / Hide excluded is missing #11212

### LINKED ISSUE #11212: "Hide excluded" link is missing in Request Publishing and Create Issue dialogs
See the fixed issue   https://github.com/enonic/app-contentstudio/issues/11046
The "Hide excluded" link in the Dependencies section behaves inconsistently across dialogs.

In the Publishing Wizard, when the `All` checkbox is unchecked, the `Hide excluded` link appears next to the DEPENDENCIES header, allowing
the user to hide excluded dependant items.

In the `Request publishing` and `Create Issue` dialogs, the link never appears — the area next to the DEPENDENCIES header stays empty, even though the dialogs share the same dependencies list behavior.

##  Steps to reproduce:
1. Select a content item with dependant items (e.g. a folder with images)
   and open the "Request publishing" (or "Create Issue") dialog.
2. Check "Include child items".
3. Uncheck the "All" checkbox in the DEPENDENCIES section.

Image

## Expexted
The "Hide excluded"/'Show excluded' link appears above the dependencies list, same as in the Publishing Wizard.  

Image

Image

reisfmb:
- Added the "Show/Hide excluded" toggle next to the DEPENDENCIES header in the three dialogs that lacked it, matching the Publish wizard.
- Extracted the header + toggle into one shared DependantsSeparator component, now used by all four dialogs.
- Each dialog tracks whether excluded dependants are visible, hides them from the list when toggled off, shows "All dependencies are excluded and hidden." when that empties the list, and snaps back to "shown" once nothing is excluded.
- Hiding is keyed on applied exclusions only, so a pending (unapplied) change never makes a row disappear mid-edit.
- The "All (n)" checkbox now counts and toggles only the rows actually on screen.
- Fixed a lazy-loading stall: with everything filtered out, the list stopped loading more items (this also affected the Publish wizard).
- Updated the e2e page object to find the toggle in its new position, plus new unit tests for all of the above.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/features/issues/model/issueDialogDetails.store.test.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/issues/model/issueDialogDetails.store.test.ts
@@ -20,7 +20,9 @@ import {
     $isIssueDialogDetailsSelectionSynced,
     $issueDialogDetails,
     $issueDialogDetailsDependantsSelection,
+    $issueDialogDetailsHasExcludedDependants,
     $issueDialogDetailsHasMoreDependants,
+    $showIssueDialogDetailsExcludedDependants,
     applyDraftIssueDialogDetailsSelection,
     cancelDraftIssueDialogDetailsSelection,
     loadIssueDialogItems,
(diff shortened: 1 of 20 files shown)


ashklianko (approved):
Nice!

PR #11253, opened 2026-08-10
Update alt text after it was generated in html area insert image dial…

### LINKED ISSUE #11252: Insert Image dialog - Alt text input is not updated on content change
If selected image in the dialog gets its alt text meta field updated on the server, the new value should be fetched from the server and loaded into the alt text input

pmi:
…og #11252

**NB** Complements https://github.com/enonic/app-ai-content-operator/pull/857


ashklianko:
Few things that bother me, let's look at few possible situations: for newly uploaded content and for existing image opened for editing.

- Newly uploaded: there's no indication or any UX signs that there's something happening that can update newly uploaded image's alt text. If the operator is working and updating the image's content with alt text then we end up in two potential scenarios, one is when the dialog is still open and the altText is set, and another one is when we press insert before the operator is done with updating the image's content. Thus visually identical for user operation leads to two different outcomes in HtmlArea's markup, where the inserted image get's empty/filled altText parameter depedning on how fast we pressed Insert button.
- Existing image open for editing: the altText field get's silently populated with the generated value if 'Decorative' accessibility was selected, but the value is not used since Decorative is the selected option. Also if there was no caption set before opening dialog then this server update will silently populate it too.

Potential improvements:
- Don't update fields of the dialog silently, show a hint/confirmation before applying

Feels like a separate task needed and/or deeper look on potential outcomes of applying server updates here.

edloidas (approved):

edloidas:
LGTM. There are things to polish visually, but those should be a separate part, when we come up with proper design for that.

pmi:
> * Newly uploaded: there's no indication or any UX signs that there's something happening that can update newly uploaded image's alt text. If the operator is working and updating the image's content with alt text then we end up in two potential scenarios, one is when the dialog is still open and the altText is set, and another one is when we press insert before the operator is done with updating the image's content. Thus visually identical for user operation leads to two different outcomes in HtmlArea's markup, where the inserted image get's empty/filled altText parameter depedning on how fast we pressed Insert button.

No can't do unfortunately, because alt text generation comes from a juke app that migh or migh not be installed. 
Whereas image dialog comes from content studio that should not care about apps user has installed.

> * Existing image open for editing: the altText field get's silently populated with the generated value if 'Decorative' accessibility was selected, but the value is not used since Decorative is the selected option. Also if there was no caption set before opening dialog then this server update will silently populate it too.

For existing content nothing is generated/updated in the background. 
In the dialog, alt text type radio only switches to informative option in case there was a non-null altText AND none selected before.

ashklianko (approved):

PR #11264, opened 2026-08-12
Handle duplicated content wherever created content is handled #11266

### LINKED ISSUE #11266: No UI update on content duplication
XP is changing `node.duplicate` to report an entire duplicated subtree in a single `node.duplicated` event, instead of `node.duplicated` for the copy plus a separate `node.created` for every duplicated descendant (enonic/xp#12261). Handlers that only listen to `$contentCreated` therefore stop seeing duplicated content altogether. This affects the seven v6 dialog services that reload their dependants when new content appears — publish, request-publish, unpublish, delete, duplicate, new-issue and issue-details: today the duplicated descendants trigger that reload, while the duplicated root already does not. To reproduce, open the publish dialog on a folder in one browser tab, duplicate a subtree below that folder in another tab, and observe that the dialog's dependants list no longer refreshes. The impact is a stale dependants list in an open dialog while a duplicate happens elsewhere.

Separately, `content.service`'s `$contentDuplicated` handler lacks the `markParentsWithChildren` bookkeeping that its `$contentCreated` handler performs, so a cached parent whose `hasChildren` is `false` is not corrected when content is duplicated into it.

#### Rationale

Duplicated content is new content. Every place that reacts to content appearing should react to both events, otherwise behaviour depends on which XP event happened to carry the node — which is exactly what enonic/xp#12261 is changing.

#### References

- enonic/xp#12261 — Duplication of a node tree fires different events for root/not root nodes

#### Implementation Notes

`content-fetcher`'s `$contentDuplicated` subscriber, `filter-tree.store`, `content.service`, `useContentComboboxData` and `duplicateDialog.store` already handle multi-node duplicated payloads and need no change. `content-fetcher` in particular was already written for the batched event: it skips descendants whose parent is in the same payload and reloads the parent's children rather than inserting locally.

The legacy browse panel also needs no change — it routes `onContentDuplicated` to the same handler as `onContentCreated`.

rymsha:
XP now reports a whole duplicated subtree in a single `node.duplicated` event instead of `node.duplicated` for the copy and `node.created` for each descendant (enonic/xp#12261, enonic/xp#12262), so handlers that only listened to the created event stopped seeing duplicates.

The seven dialog services that reload their dependants when new content appears — publish, request-publish, unpublish, delete, duplicate, new-issue and issue-details — now subscribe to `$contentDuplicated` alongside `$contentCreated`. Each shared callback was extracted into a named `handleCreatedContent` so both subscriptions run identical logic.

`content.service`'s `$contentDuplicated` handler also gained the `markParentsWithChildren` bookkeeping its `$contentCreated` handler already performed, so a cached parent whose `hasChildren` is `false` is corrected when content is duplicated into it.

No change was needed in `content-fetcher`, `filter-tree.store`, `useContentComboboxData`, `duplicateDialog.store` or the legacy browse panel — they already handle multi-node duplicated payloads.

Verified with `pnpm -C ./modules/lib run check` (types, lint, boundaries clean) and `test:run` (1685 tests passing).

Closes #11266

[Claude Code session](https://claude.ai/code/session_017cKmPRVUyVbKcZDho2MDtX)

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/entities/content/model/content.service.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/entities/content/model/content.service.ts
@@ -81,6 +81,7 @@ export const start = (): void => {
         $contentDuplicated.subscribe((event) => {
             if (event?.data) {
                 setContents(event.data);
+                markParentsWithChildren(event.data);
             }
         }),
         // $contentMoved is the only event for cross-parent moves; delete/create are not emitted.
--- a/modules/lib/src/main/resources/assets/js/v6/features/delete/model/deleteDialog.service.ts
+++ b/modules/lib/src/main/resources/assets/js/v6/features/delete/model/deleteDialog.service.ts
@@ -9,7 +9,9 @@ import {
     $contentArchived,
     $contentCreated,
     $contentDeleted,
+    $contentDuplicated,
     $contentUpdated,
+    type ContentEvent,
 } from '../../../shared/socket/socket.store';
 import { resolveForDelete } from '../api/delete.api';
 import {
@@ -183,6 +185,13 @@ const isDialogActive = (): boolean => {
     return open && !submitting && items.length > 0;
 };
 
+const handleCreatedContent = (event: ContentEvent | null): void => {
+    if (!event || !isDialogActive()) return;
+
+    // New content could be a child of items being deleted
+    reloadDeleteDialogDataDebounced();
+};
+
 //
 // * Service Lifecycle
 //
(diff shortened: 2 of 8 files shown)


alansemenov (changes requested):
Remove comments from AI

rymsha:
Removed in `f90d692`. Gone are the seven `// Duplicates arrive as their own event...` lines, the two-line block above the `$contentDuplicated` handler in `content.service.ts`, and a JSDoc line I had put on the extracted handler in `duplicateDialog.service.ts`.

One comment still shows as added in the diff — `// New content could be a child of items being deleted` in `deleteDialog.service.ts`. That one is pre-existing and unchanged; it just moved into the extracted function along with the code it describes. Happy to drop it too if you'd rather it went.

`check:types` and `check:lint` are clean and all 1685 unit tests pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

alansemenov (approved):

PR #11337, opened 2026-08-25
Add mobile navigation drawer to browse view #11325

### LINKED ISSUE #11325: Add mobile navigation drawer to browse view
### Summary
Add a mobile-only navigation drawer for BrowseSidebar controlled by a hamburger button in BrowseAppBar. Drawer should overlay browse content without reducing available content width. Desktop and tablet sidebar behavior must remain unchanged.
Current behavior
- Sidebar permanently occupies 60px on mobile.
- Browse content retains sidebar-related left offsets.
- Mobile navigation lacks dedicated open and close interactions.
- Narrow header content can shrink fixed-size controls.

Expected behavior
- Mobile sidebar remains closed by default.
- Hamburger icon animates into a close icon while drawer is open.
- Sidebar slides smoothly over content when opening and closing.
- Clicking outside drawer or pressing Escape closes it.
- Content and XP menu controls outside open drawer cannot receive pointer or keyboard interaction.
- Mobile browse content uses full width with no sidebar offset.

Image

Image

skjulestad:
### Summary
Add a mobile-only navigation drawer for BrowseSidebar controlled by a hamburger button in BrowseAppBar. Drawer should overlay browse content without reducing available content width. Desktop and tablet sidebar behavior must remain unchanged.
Current behavior
- Sidebar permanently occupies 60px on mobile.
- Browse content retains sidebar-related left offsets.
- Mobile navigation lacks dedicated open and close interactions.
- Narrow header content can shrink fixed-size controls.

Expected behavior
- Mobile sidebar remains closed by default.
- Hamburger icon animates into a close icon while drawer is open.
- Sidebar slides smoothly over content when opening and closing.
- Clicking outside drawer or pressing Escape closes it.
- Content and XP menu controls outside open drawer cannot receive pointer or keyboard interaction.
- Mobile browse content uses full width with no sidebar offset.

Image

Image

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
@@ -1,9 +1,10 @@
 import type { Extension } from '@enonic/lib-admin-ui/extension/Extension';
 import { Store } from '@enonic/lib-admin-ui/store/Store';
-import { Tooltip } from '@enonic/ui';
+import { cn, Tooltip } from '@enonic/ui';
 import { useStore } from '@nanostores/preact';
 import { type LucideIcon, Pen, Settings } from 'lucide-react';
-import { type ReactElement, useCallback } from 'react';
+import { type ReactElement, useCallback, useEffect } from 'react';
+import { useBreakpoints } from '../../shared/lib/hooks/useBreakpoints';
 import { useI18n } from '../../shared/lib/hooks/useI18n';
 import { DefaultProjectIcon } from '../../shared/ui/icons/DefaultProjectIcon';
 import { ProjectIcon } from '../../shared/ui/icons/ProjectIcon';
@@ -11,6 +12,8 @@ import { LegacyElement } from '../../shared/ui/LegacyElement';
 import { WidgetButton } from '../../shared/ui/WidgetButton';
 import { $activeProject, $noProjectMode } from '../../entities/project';
 import { $config } from '../../shared/config';
+import { BROWSE_SIDEBAR_ID, BROWSE_SIDEBAR_TOGGLE_ID } from './browseSidebar.constants';
+import { $isBrowseSidebarOpen, setBrowseSidebarOpen } from './model/browseSidebar.store';
 import {
     $sidebarWidgets,
     getSettingsWidget,
@@ -21,8 +24,12 @@ import {
 } from '../../widgets/context-panel/model/sidebarWidgets.store';
 
 function getWidgetIcon(widget: Readonly<Extension>): LucideIcon | undefined {
-    if (isMainWidget(widget)) return Pen;
-    if (isSettingsWidget(widget)) return Settings;
+    if (isMainWidget(widget)) {
+        return Pen;
+    }
+    if (isSettingsWidget(widget)) {
+        return Settings;
+    }
     return undefined;
 }
 
@@ -31,15 +38,25 @@ function stripVersionSuffix(version: string): string {
 }
 
 export const BrowseSidebar = (): ReactElement => {
+    const isOpen = useStore($isBrowseSidebarOpen);
+    const { sm } = useBreakpoints();
     const activeProject = useStore($activeProject);
     const noProjectMode = useStore($noProjectMode);
     const { widgets, activeWidgetId } = useStore($sidebarWidgets);
     const config = useStore($config, { keys: ['appVersion'] });
-    const name = Store.instance().get('application').getName();
+    const applicationName = Store.instance().get('application').getName();
     const version = `v${stripVersionSuffix(config.appVersion)}`;
     const settingsWidget = getSettingsWidget(widgets);
     const mainWidgets = noProjectMode ? [] : widgets.slice(0, -1);
     const footerWidget = noProjectMode ? settingsWidget : widgets.at(-1);
+    const footerWidgetButtonId = footerWidget ? `browse-sidebar-widget-${getWidgetKey(footerWidget)}` : undefined;
+    const sidebarLabel = useI18n('wcag.sidebar.label');
+    const isMobileSidebarClosed = !sm && !isOpen;
+
+    const closeSidebar = useCallback((): void => {
+        setBrowseSidebarOpen(false);
+        document.getElementById(BROWSE_SIDEBAR_TOGGLE_ID)?.focus();
+    }, []);
 
     const isWidgetActive = useCallback(
         (widget: Readonly<Extension> | undefined) => {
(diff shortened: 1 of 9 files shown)


skjulestad:
Update: Recent changes requested:
Change from project icon back to project name for mobile as per cloud console design.
Because of this I will also revert the text in the burger menu back to "Content Studio", as we do not need project name both in browse-toolbar and within the hamburger menu. Also added truncate to project name to avoid overflow in browse-toolbar.

skjulestad:
Bug found, when transition makes it so the menu "slides" open / close on screen adjustment, working on a solution for this and will push it to pr asap

skjulestad:
https://github.com/user-attachments/assets/94ac8591-e04f-4a15-aebb-4465a0c32a7f
Small bug partially fixed, need changes outside of scope to fix later, browse toolbar transition should be instant as well.

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
```diff
@@ -48,62 +75,168 @@ export const BrowseSidebar = (): ReactElement => {
...
+                                <div key={widgetKey} className="flex w-full items-center gap-1.5 sm:block sm:w-auto">
+                                    <WidgetButton
+                                        id={widgetButtonId}
+                                        label={widget.getDisplayName()}
+                                        icon={getWidgetIcon(widget)}
+                                        active={isWidgetActive(widget)}
+                                        iconUrl={widget.getIconUrl() && widget.getFullIconUrl()}
+                                        tooltipClassName="max-sm:hidden"
+                                        onClick={() => {
+                                            setActiveWidget(widget);
```
edloidas, line 145: **Picking something in the drawer should close it**

Tapping a widget only calls `setActiveWidget`. The drawer stays open, the backdrop stays up, and `.extensions-block` — which has just rendered the thing you picked — is still `inert`. So you tap Settings, and then you have to tap somewhere else to actually see Settings.

One line in both handlers (`setBrowseSidebarOpenAnimated(false)`). No `sm` guard needed — on desktop the value has no visible effect anyway.

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
```diff
@@ -48,62 +75,168 @@ export const BrowseSidebar = (): ReactElement => {
...
+        });
+        return () => {
+            previousStates.forEach(({ element, inert }) => {
+                element.inert = inert;
+            });
+        };
+    }, [isOpen, sm]);
+
+    useEffect(() => {
+        previousSmRef.current = sm;
```
edloidas: **This should also reset the drawer, and most of the animation state can go**

Two things here.

Nothing closes the drawer when you cross the breakpoint. Open it on mobile, rotate to landscape, rotate back — it's open again on top of your content. `setBrowseSidebarOpen(false)` here fixes it, and that is exactly what that setter looks like it was written for; right now it has no callers anywhere.

The bigger one: `isAnimating`, `previousSmRef` and a 500ms `setTimeout` are three mechanisms guarding one rule — "don't animate when the breakpoint flips". CSS is already doing most of that work: `sm:transition-none` covers one direction and the gate covers the other, so the ref comparison isn't buying much. I'd keep one boolean, make it local `useState` rather than a store atom, and clear it in `onTransitionEnd` so the duration only exists in the Tailwind class. Should come out around five lines.

Worth internalising the general shape: CSS does the animating, JS only decides whether it's allowed to run. Any time JS is tracking *how long* an animation takes, something has been duplicated.

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
```diff
@@ -48,62 +75,168 @@ export const BrowseSidebar = (): ReactElement => {
...
+        return () => window.removeEventListener('keydown', closeOnEscape);
+    }, [closeSidebar, isOpen]);
+
+    useEffect(() => {
+        if (sm || !isOpen) {
+            return;
+        }
+
+        const elements = [
+            sidebarRef.current?.closest('.main-app-wrapper')?.querySelector<HTMLElement>('.extensions-block'),
```
edloidas: **React shouldn't go hunting for legacy DOM**

This works, but it means the sidebar knows the internal structure of `AppWrapper`. And it fails quietly — with `.filter(Boolean)`, a rename over there just silently switches off focus containment over here. No error, no failing test, and the thing that breaks is an accessibility guarantee.

`.claude/rules/legacy-to-react.md` lists the three ways across the seam, and they're all "publish state, let the other side react to it". Both elements here are ours: `AppWrapper` owns `.extensions-block`, `main.ts` creates `xp-menu`, and both already subscribe to nanostores. Let them subscribe to `$isBrowseSidebarOpen` and set `inert` on their own elements, and this effect disappears completely.

Worth fixing now rather than later — one selector is a shortcut, five is an architecture.

edloidas (changes requested):
Hey Mathias, good work overall. The drawer does what the issue asks, and the desktop rail remains untouched, which was the fiddly part. I'd like one more round on it before it lands, though.

I can't approve these changes in their current form.

First, I left three inline comments. Only the first one is really a bug; the other two are about the approach, and I think both of them would actually make the next version *smaller*, not bigger. They address some of the problems with the current implementation, but that's not enough.

The general approach to this feature isn't quite right. It adds too much code for such a small feature, and you're trying to control the animations with JS too much. This can be done much more simply, and right now you're having to solve problems introduced by the implementation itself.

As for the UI part, I don't think this is the right approach either, it's not *complete* yet. First, there isn't proper touch support, so users won't be able to open the sidebar with a swipe.

Also, this is debatable, but I'd much rather have an X-style mobile sidebar than a Slack-style one: the sidebar should push the content to the right and dim it, rather than covering it.

Let's do the following: first, create a prototype (rework it) without any animations or touch effects and show how the mechanism would work from a UI/UX perspective. Then implement the animations once we know how the interaction should work. After that, add touch-event handling. Show it to Alan and maybe Thomas after step 1. May be a good thing to drop several prototypes for them.

In my opinion, though, you can't really do steps 2 and 3 properly without applying the same changes to the other non-animated panels as well.

modules/lib/src/main/resources/assets/js/v6/pages/browse/model/browseSidebar.store.ts
```diff
@@ -0,0 +1,10 @@
+import { atom } from 'nanostores';
+
+export const BROWSE_SIDEBAR_ID = 'browse-sidebar';
```
edloidas: Two small ones, grouped.

**These two ids.** The store never reads them — they're DOM ids for `aria-controls` and `getElementById(...).focus()`. Other stores do hold ID constants (`settings-tree.store.ts`, `pageComponents.store.ts`), but those are domain ids the store actually reasons about. These are presentation. They're shared by two components so they need *some* home, just not this one.

**`BrowseSidebar.test.tsx:131`** asserts the nav's className does *not* contain `transition`. That pins the absence of the animation, so it'll need inverting when the animation comes back — and it's the one assertion in the file testing a negative about styling rather than a behaviour.

modules/lib/src/main/resources/assets/js/v6/pages/browse/model/browseAppBar.store.ts
```diff
@@ -0,0 +1,12 @@
+import { atom } from 'nanostores';
+
+export const $isProjectSelectorVisible = atom<boolean>(true);
+export const $appName = atom<string>('');
```
edloidas: `$appName` has an absurd chain length. The chain predates you and you only moved it into a file. But the file is new, so it's worth changing while it's fresh.

Both values here come from `$activeWidget`: the display name, and `getConfig().getProperty('context') === 'project'`. 
The route they currently take is `$activeWidget` → AppWrapper subscribes → `selectExtension` → `appBar.setAppName()` → the legacy element setter → the atom → `useStore` back in the same component. 
**Six** hops from a nanostore to a nanostore. `stores.md` has a line for exactly this — *"A fact store should never derive from another fact store — use `computed`"* — and two computeds would delete this file along with the three methods on `BrowseAppBarElement` that the comment above them already calls temporary.

One trap if you do it: `$isProjectSelectorVisible` starts `true`, and a plain projection over an undefined widget gives `false`. `$appName` is fine either way thanks to the `appName || applicationName` fallback.

edloidas (changes requested):
Few structural changes need to be done.

edloidas (approved):
Nice one

PR #11360, opened 2026-08-31
Add preview panel support for phones #11339

### LINKED ISSUE #11339: Preview panel
### Summary

Make the preview panel available in phone layouts and integrate its controls into the browse toolbar. On phones, users should switch between the content list, preview, details, and search filter without desktop-style split panels. Desktop and tablet layouts should retain their existing behavior, including the ability to keep the search filter and details panel open simultaneously.

### Current behavior

- The preview panel is not directly available on phones.
- The browse toolbar does not adapt when the preview or details panel is active.
- Narrow layouts cannot present all primary actions cleanly.

### Expected behavior

- Show the preview toggle immediately to the left of the details toggle on phones.
- On phones, keep the preview, details, and search filter panels mutually exclusive.
- On desktop and tablet, preserve the existing split-panel behavior and allow the search filter and details panel to remain open together.
- Replace regular toolbar actions with relevant status and widget controls while the preview or details panel is open on phones.
- Show an `Actions` menu when the available width only supports one action.
- Hide preview-size emulation controls below tablet width.
- Avoid content-width transition animations when the mobile navigation drawer appears, disappears, or is resized.

### Acceptance criteria

- [x] Preview can be opened and closed from the phone browse toolbar.
- [x] Preview toggle is disabled when no content is selected.
- [x] On phones, opening preview closes the details and search filter panels.
- [x] On phones, opening details closes the preview and search filter panels.
- [x] On phones, opening the search filter closes the preview and details panels.
- [x] Preview toolbar shows publish status, version history, and active preview widget controls.
- [x] Details toolbar shows the active widget icon and display name.
- [x] Narrow layouts expose available actions through an `Actions` menu.
- [x] Preview-size emulation controls remain hidden on phones.
- [x] Desktop and tablet behavior remains unchanged.
- [x] On desktop and tablet, the search filter and details panel can remain open simultaneously.
- [x] Mobile navigation drawer changes and sidebar resizing do not animate the content width.

### Notes

The preview and content grid remain mounted while switching views to avoid reloading the preview iframe.

skjulestad:
### Summary

Make the preview panel available in phone layouts and integrate its controls into the browse toolbar. On phones, users should switch between the content list, preview, details, and search filter without desktop-style split panels. Desktop and tablet layouts should retain their existing behavior, including the ability to keep the search filter and details panel open simultaneously.

### Current behavior

- The preview panel is not directly available on phones.
- The browse toolbar does not adapt when the preview or details panel is active.
- Narrow layouts cannot present all primary actions cleanly.

### Expected behavior

- Show the preview toggle immediately to the left of the details toggle on phones.
- On phones, keep the preview, details, and search filter panels mutually exclusive.
- On desktop and tablet, preserve the existing split-panel behavior and allow the search filter and details panel to remain open together.
- Replace regular toolbar actions with relevant status and widget controls while the preview or details panel is open on phones.
- Show an `Actions` menu when the available width only supports one action.
- Hide preview-size emulation controls below tablet width.
- Avoid content-width transition animations when the mobile navigation drawer appears, disappears, or is resized.

### Acceptance criteria

- [x] Preview can be opened and closed from the phone browse toolbar.
- [x] Preview toggle is disabled when no content is selected.
- [x] On phones, opening preview closes the details and search filter panels.
- [x] On phones, opening details closes the preview and search filter panels.
- [x] On phones, opening the search filter closes the preview and details panels.
- [x] Preview toolbar shows publish status, version history, and active preview widget controls.
- [x] Details toolbar shows the active widget icon and display name.
- [x] Narrow layouts expose available actions through an `Actions` menu.
- [x] Preview-size emulation controls remain hidden on phones.
- [x] Desktop and tablet behavior remains unchanged.
- [x] On desktop and tablet, the search filter and details panel can remain open simultaneously.
- [x] Mobile navigation drawer changes and sidebar resizing do not animate the content width.

### Notes

The preview and content grid remain mounted while switching views to avoid reloading the preview iframe.


modules/lib/src/main/resources/assets/js/v6/pages/browse/ui/layout/BrowseLayout.tsx
```diff
@@ -94,13 +94,20 @@ export const BrowseLayout = ({
     }, []);
 
     const isMobile = mode === 'mobile';
+
+    useEffect(() => {
+        if (isMobile && isFilterOpen && (isContextOpen || isMobilePreviewOpen)) {
+            setContentFilterOpen(false);
```
edloidas, line 100: The exclusivity here only runs one way, and the store's does not cover the filter at all.
`setContextOpen` / `setMobilePreviewOpen` clear each other, but `setContentFilterOpen(true)`
clears nothing -- so this effect is the only thing enforcing filter exclusivity, and it
always resolves in favour of the panel that is already open.

That makes a programmatic filter open a silent no-op on phones. Trace:

- The dependencies context widget fires `ToggleSearchPanelWithDependenciesGlobalEvent`
- `ContentAppContainer.ts:56` re-fires it as `ToggleSearchPanelWithDependenciesEvent`
- `ContentBrowsePanel.ts:340` calls `showFilterPanel()` -> `setContentFilterOpen(true)`
- this effect sees `isMobile && isFilterOpen && isContextOpen` and calls
  `setContentFilterOpen(false)` before anything renders

So on a phone with the details panel open, "show inbound/outbound dependencies" does
nothing at all -- no filter, no error, and the context panel stays put. `SearchAndExpandItemEvent`
from `MoveContentDialog.ts:245` reaches `showFilterPanel()` the same way at
`ContentBrowsePanel.ts:352` and has the same outcome if the context panel happens to be open.

Before this change `showMobileFilter` had no `!isContextOpen` guard, so the filter overlay
did appear (stacked over the context panel -- a different bug, which this correctly fixes).

Suggestion: put the exclusion in the store next to the other two rather than in an effect,
so whichever panel is opened last wins consistently:

```ts
export function setContentFilterOpen(open: boolean): void {
    if (open) {
        $isContextOpen.set(false);
        $isMobilePreviewOpen.set(false);
    }
    ...
}
```

then this effect and the `!isContextOpen && !isMobilePreviewOpen` terms in `showMobileFilter`
can both go. It also removes a render-cascade: today the filter opens, renders, and closes
across two commits.

I could not exercise this in a running app, so please sanity-check the reachability -- the
control flow itself is unconditional on the deps listed.

edloidas (commented):
Reviewer notes on the phone preview panel work. The store consolidation into
`shared/app-state/browsePanels.store` is the right move, and keeping grid and preview
mounted through the mode change is clearly the correct call.

One item I would like settled before merge: the panel exclusivity in `BrowseLayout` is
one-directional, so a programmatic `setContentFilterOpen(true)` is silently undone while
the mobile context or preview panel is open. Details inline.

Two smaller things not worth blocking on:

`main-app-wrapper.less` drops `transition: left 0.5s ease` from `.main-app-wrapper`
outright. That kills the animation for the desktop sidebar expand/collapse too, not only
the mobile drawer -- is that intended given the "Desktop and tablet behavior remains
unchanged" criterion? Scoping it to `body:not(.mobile-mode)` would keep both.

`BrowseToolbar` now mixes two breakpoint sources in one component. `mode === 'mobile'`
comes from `LayoutTokens.contextPanel.mobileThreshold` (`_540_720`, measured on layout
width excluding the ~60px sidebar, so roughly a 780px viewport), while the surviving
`sm:` / `max-sm:` classes are Tailwind's 640px. Between those two widths the JS branch and
the CSS branch disagree about what "mobile" means. Worth a look at ~700px.

What I ran: `pnpm -C ./modules/lib run check` (types, lint, boundaries) and
`pnpm -C ./modules/lib run test:run` -- both green, 1778 tests passing. What I did not run:
the app itself, so nothing here about rendering, iframe reload behavior, or the actual
phone layout is observed -- the findings below are traced through the code only.

modules/lib/src/main/resources/assets/js/app/view/ContentItemPreviewPanel.ts
```diff
@@ -142,6 +143,11 @@ export class ContentItemPreviewPanel extends Panel implements ExtensionRenderer
 
     private setupListeners() {
         $previewRefreshRequested.listen(() => this.refresh());
+        $isMobilePreviewOpen.listen((isOpen) => {
```
edloidas, line 146: This forces a full preview re-render on every mobile preview open, which works against the
`// !` note at `BrowseLayout.tsx:102` and the PR's own "avoid reloading preview iframe".

`refresh()` -> `update(content)` -> `ExtensionRenderingHandler.render()`, and `render()` has
no same-key short-circuit: it sets `lastRenderKey`, calls `showMask()`, and re-issues
`doRender()`. The `lastRenderKey` guard that would skip a redundant render lives only in
`handleExtensionEvent` (line 325), which this path bypasses.

The content should not need it either -- `ContentItemStatisticsPanel.setItem` (line 22)
forwards every selection change to the preview panel with no visibility guard, so the
preview stays current while the panel is collapsed to zero width.

Is this a workaround for the iframe having been laid out at zero width while collapsed? If
so a comment saying that would help, since it reads as redundant otherwise. If not, dropping
the listener saves a mask, a request, and an iframe reload on every toggle -- on the one
platform where that round trip costs the most.

modules/lib/src/main/resources/assets/js/app/browse/ResponsiveBrowsePanel.ts
```diff
@@ -24,7 +23,6 @@ import { ResponsiveToolbar } from './ResponsiveToolbar';
 
 export abstract class ResponsiveBrowsePanel extends BrowsePanel {
     static MOBILE_MODE_CLASS = 'mobile-mode';
```
edloidas, line 25: Removing the `mobile-preview-on` toggle leaves its styling and its legacy API unreachable.
Worth cleaning up in this PR while the context is fresh:

- `.mobile-preview-on` rules in `styles/body.less:30`, `styles/browse/responsive-browse-panel.less:3`,
  `styles/browse/responsive-toolbar.less:6`, `styles/browse/content-browse-toolbar.less:24,35`
- `styles/body.less:27` -- `transition: top 0.5s ease-in-out` on `.mobile-mode .content-app-panel`,
  now pointless since `top` no longer changes
- `ResponsiveToolbar.onFoldClicked` has no callers left after the handler here was deleted, so
  `hideMobilePreviewButton` is a permanently `display: none` button with no behaviour;
  `enableMobileMode` / `disableMobileMode` are likewise unreachable from this path

Nothing here breaks -- the button stays hidden because `.hide-mobile-preview-button` is
`display: none` without the `mobile-preview-on` ancestor. It is just dead weight that gets
harder to identify later.

modules/lib/src/main/resources/assets/js/v6/widgets/context-panel/widget/WidgetsSelector.tsx
```diff
@@ -33,6 +42,11 @@ const WidgetsSelector = ({ widgetViews = [], externalSelectedWidgetView = undefi
     );
     const selectedWidgetLabel = selectedWidgetView?.getExtensionName();
     const inputValue = isOpen ? searchValue : selectedWidgetLabel;
+    const mobileToolbarTargetElement =
+        mobileToolbarTarget && mode === 'mobile' && isContextOpen
+            ? document.querySelector<HTMLElement>(`[data-component="${mobileToolbarTarget}"]`)
```
edloidas, line 47: A `document.querySelector` in the render body has no way to re-run when the target appears.
It resolves today only because the target div in `BrowseToolbar` is always in the DOM (it is
`hidden`, not unmounted) and both this and the toolbar re-render off the same two stores. If
either of those holds stops holding, this silently falls back to the inline layout with no
signal.

A ref-based or state-based lookup on the toolbar side -- or passing the container down --
would make the coupling explicit. Non-blocking; the styling here reads well otherwise.

modules/lib/src/main/resources/assets/js/app/browse/ResponsiveBrowsePanel.ts
```diff
@@ -4,17 +4,16 @@ import { type ViewItem } from '@enonic/lib-admin-ui/app/view/ViewItem';
...
-import { setMobilePreviewOpen } from '../../v6/pages/browse/model/browseLayout.store';
+import { setMobilePreviewOpen } from '../../v6/shared/app-state/browsePanels.store';
 import {
     $contextPanelMode,
     $isContextLayoutMeasured,
     shouldCollapseContextInitially,
-} from '../../v6/widgets/context-panel/model/contextPanelMode.store';
-import { $isContextOpen, setContextOpen } from '../../v6/widgets/context-panel/model/contextWidgets.store';
+} from '../../v6/shared/app-state/browsePanels.store';
+import { $isContextOpen, setContextOpen } from '../../v6/shared/app-state/browsePanels.store';
```
edloidas, line 16: Three separate imports from `browsePanels.store` (lines 10, 11-15, 16) -- worth merging into
one. Same pattern at `ContentBrowsePanel.ts:24-25` (`entities/project`) and
`BrowseLayout.test.tsx:4-5`.

skjulestad:
Need to implement one more feature, setting to draft meanwhile

modules/lib/src/main/resources/assets/js/v6/widgets/context-panel/widget/WidgetsSelector.tsx
```diff
@@ -33,6 +42,8 @@ const WidgetsSelector = ({ widgetViews = [], externalSelectedWidgetView = undefi
     );
     const selectedWidgetLabel = selectedWidgetView?.getExtensionName();
     const inputValue = isOpen ? searchValue : selectedWidgetLabel;
+    const mobileToolbarTargetElement = mode === 'mobile' && isContextOpen ? mobileToolbarTarget : undefined;
```
edloidas, line 61: On a phone the widget search field in the details toolbar cannot take a space or a caret move, so a widget name with a space in it cannot be typed.

Portaling the `Combobox` here puts `Combobox.Input` inside `BrowseToolbar`'s `Toolbar.Container`, whose roving-tabindex `onKeyDown` calls `preventDefault()` for `ArrowLeft`, `ArrowRight`, `Home`, `End`, `Enter` and `" "` whenever the toolbar has registered items, with no check for an editable target. `PreviewToggle` and `ContextToggle` are always registered, so that list is never empty. `Combobox.Input`'s own handler covers only Enter/ArrowUp/ArrowDown/Escape and does not stop propagation, and a portal does not stop native bubbling.

Measured on the running app with the details panel open, dispatching a cancelable `keydown` on the field and reading `defaultPrevented`:

| key | desktop 1440, rendered inline | phone 390, portaled |
| --- | --- | --- |
| Space, ArrowLeft, ArrowRight, Home, End | false | true |
| `a` | false | false |

Only the roving keys, and only when portaled.

Either move the portal target outside `Toolbar.Container`, or stop propagation from the input for the keys the toolbar claims.

modules/lib/src/main/resources/assets/js/v6/widgets/browse-toolbar/BrowseToolbar.tsx
```diff
@@ -67,20 +88,37 @@ export const BrowseToolbar = ({
...
+                    <>
+                        {mobilePreviewVersionHistoryButton}
+                        <div className="ml-3.5 min-w-0 flex-1" />
+                        {mobilePreviewWidgetSelector}
+                    </>
+                ) : showMobileContext ? null : (
+                    <>
+                        <div className="flex min-w-fit items-center gap-2 sm:min-w-0 sm:flex-1">
+                            <SearchToggle action={toggleFilterPanelAction} />
+                            <div className="sm:hidden shrink-0 min-w-fit">
```
edloidas: The `Actions` menu is hidden at exactly the widths that need it.

It is gated `sm:hidden`, a 640px **viewport** query, while the panels it accompanies switch on `$contextPanelMode`, computed from the layout's **measured** width (`totalWidth <= 720`, application sidebar excluded). The `sm:` gate predates this branch; what is new here is the second signal, so the two now disagree over a band.

Measured on the running app, sidebar rail 60px:

| viewport | toolbar width | mobile panel mode | `Actions` menu | desktop action row |
| --- | --- | --- | --- | --- |
| 800 | 740 | no | none | flex |
| 780 | 720 | yes | none | flex |
| 768 | 708 | yes | none | flex |
| 700 | 640 | yes | none | flex |
| 640 | 580 | yes | none | flex |
| 639 | 639 | yes | block | none |

From 640 to 780 the layout is in phone mode while the full desktop action row renders into a toolbar as narrow as 580px, and the menu this issue asks for stays hidden. iPad portrait, 768, sits in the middle of that band.

Gating the action rows on the same measured width the panel mode already uses would close it.

modules/lib/src/main/resources/assets/js/v6/widgets/browse-toolbar/BrowseToolbar.tsx
```diff
@@ -67,20 +88,37 @@ export const BrowseToolbar = ({
...
+                            <OverflowActionRow actions={toolbarActions} className="hidden sm:flex min-w-0 flex-1" />
+                        </div>
+                        <div className="flex-1 sm:hidden" />
+                        <div className="hidden sm:flex shrink-0 min-w-fit">
+                            <SplitActionButton actions={[publishSplitActions]} />
+                        </div>
+                    </>
+                )}
+                <PreviewToggle className="shrink-0 size-7.5" />
+                <ContextToggle className="shrink-0 size-7.5" tooltipClassName="max-md:hidden" />
```
edloidas: This shrinks the details toggle everywhere, not only on phones.

`cn()` is twMerge, so the trailing unprefixed `size-7.5` drops the component's own `size-9` at every breakpoint. Measured at 1440 on the running app, the details toggle renders 30x30 next to a 36x36 search toggle and 36px-tall action buttons.

The container values on this line's own parent are re-pinned at `sm:` (`px-2.25 max-sm:pr-4 sm:px-5`), which reads as deliberate desktop preservation — the toggle sizes just did not get the same treatment. `size-7.5 sm:size-9` would scope it, or `SearchToggle` wants the same 30px if the smaller size is the intent. Which way did you mean it?

modules/lib/src/main/resources/assets/js/v6/pages/settings/ui/SettingsTreeListToolbar.tsx
```diff
@@ -50,7 +50,7 @@ const SettingsTreeListToolbar = ({ enabled = true }: SettingsTreeListToolbarProp
     };
 
     return (
-        <div className="bg-surface-neutral flex items-center justify-between px-5 py-2.5 gap-2">
+        <div className="bg-surface-neutral flex items-center justify-between max-sm:px-3.5 px-5 py-2.5 gap-2">
```
edloidas: Two small ones, grouped.

`max-sm:px-3.5` lands on two Settings toolbars, here and in `SettingsBrowseToolbar.tsx`. Settings is not in this issue's scope, and while the change is additive below `sm` only and mirrors the browse tree's own, it is still behaviour nobody asked to change.

`setContextOpen` moved into `browsePanels.store` and picked up browse-specific side effects — it always clears `$isMobilePreviewOpen`, sets `lastOpenedPanel`, and clears the filter in mobile mode — while still being re-exported from `contextWidgets.store` to `ContentWizardPanel` and the page-editor `bridge`. I chased the state loss that implies and it does not happen: `openEditContentTab` goes through `window.open`, so the wizard is a separate tab with its own module instances and those writes are inert there. So this is an ownership question rather than a bug, but a browse-named store reached from the wizard will bite eventually.

edloidas (changes requested):
I ran this branch against a local XP 8.2.0 instance at phone (390px) and desktop (1440px) widths. The panel state machine holds up: preview, details and filter are mutually exclusive on a phone, the store owns that now rather than an effect, the preview toggle is correctly disabled with nothing selected, and the `Actions` menu does appear once the viewport drops below 640.

Two of the four comments I would not merge without: the widget search field inside the mobile toolbar, and the width signal the action rows switch on. Both reproduce against the running app, and the second one lands on this issue's own `Actions`-menu criterion at 768px. The toggle-size comment is a judgement call I would still like an answer on, and the small ones are grouped in the last comment.

This excludes the new `action.actions` key being English-only — translations land in their own commit in this repo, so that is not this PR's business.

skjulestad:
Also hid emulate screen size for mobile devices in total.

ashklianko (approved):

PR #11362, opened 2026-08-31
Filter panel mobile #11361

### LINKED ISSUE #11361: Fixes for the Filter panel in mobile mode
Mobile changes:
* Hide “Export” button
* Turn “X results” into a link “Show X results” which will collapse the panel and show the filtered grid. When nothing found (0 results), it should be a simple text “0 results”, like now.

Global change:
* If search is applied, replace the magnifying glass icon with “search-check” to show user that the grid is filtered. NB! Also for desktop.

skjulestad:
Mobile changes:

Hide “Export” button
Turn “X results” into a link “Show X results” which will collapse the panel and show the filtered grid. When nothing found (0 results), it should be a simple text “0 results”, like now.
Global change:

If search is applied, replace the magnifying glass icon with “search-check” to show user that the grid is filtered. NB! Also for desktop.

This issue was resolved by a human :)


edloidas (approved):
Nice!

PR #11376, opened 2026-09-02
Replace editor back icon with project icon #11367

### LINKED ISSUE #11367: Replace editor back icon with project icon
The editor toolbar currently uses a generic Back arrow for the control that navigates to the active project’s browse view. Replace the arrow with the active project’s icon so the destination is clearer. This is a visual change only; existing navigation behavior remains unchanged.

skjulestad:
The editor toolbar currently uses a generic Back arrow for the control that navigates to the active project’s browse view. Replace the arrow with the active project’s icon so the destination is clearer. This is a visual change only; existing navigation behavior remains unchanged.


edloidas (approved):
LGTM

PR #11379, opened 2026-09-03
Update detached page component #11368

### LINKED ISSUE #11368: Fixes for Page Component view in mobile resolution
Page component needs to update icons from "Chevrons" to "Netowrk" and "X" to fit better with current logic.
Detached page component should also be resizable for easier handling for both mobile and desktop users.
Currently there is no logic for handling resizing, and with long lists moving content around can be a hastle.

Solution update icons and add resize logic, let users decide how much screen space detached page component shall take to allow full access to content.

Added small fix where scrolling within detached page component would scroll the entire screen.

skjulestad:
Update icons and add resize logic, let users decide how much screen space detached-page-component shall take to allow full access to content.

Added small fix where scrolling within detached page component would scroll the entire screen.


modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/layout/detached-page-components/DetachedPageComponentsView.tsx
```diff
@@ -0,0 +1,394 @@
...
+            desiredHeight: number,
+            maximumHeight?: number,
+            viewport: ViewportBounds = getViewportBounds(),
+        ): void => {
+            const range = getPanelHeightRange(snapshot, viewport, maximumHeight);
+            const resized = resizePanelToHeight(snapshot, desiredHeight, viewport, maximumHeight);
+            setLayout((current) => ({
+                ...current,
+                ...resized,
+                height: getManualPanelHeight(resized.height, range.max),
```
edloidas: ### The height a user sets from the top handle is discarded, and the panel's bottom edge jumps down

Resizing from the top edge should grow the panel upward with its bottom staying put. Past a point it does the opposite. `getManualPanelHeight` returns `undefined` — auto height — whenever the clamped height reaches `range.max`, and for the top edge `range.max` is `min(startTop + startHeight - viewport.top, fitHeight)`. So when the tree is taller than the space above the panel's bottom, the cap comes from the viewport rather than the content, the height just set is thrown away, and the render recomputes `maxHeight` off the new smaller `layout.top` (line 312) so the auto-sized panel grows past the height it had reached.

Panel at `top: 96, height: 300` (bottom at 396), viewport 600, tree 700: dragging the top handle to the top gives `{top: 0, height: undefined}`, `maxHeight` becomes `min(600, 700) = 600`, the box renders 600 tall, and the bottom moves from 396 down to 600. `End` on the focused top handle does the same in one keypress, since `desiredHeight` is `range.max` exactly. The bottom edge is unaffected — there `maxHeight` happens to equal the height reached.

Line 104 discards it on a different trigger: collapse a node so `fitHeight` falls below the set height and the choice is gone; expanding sizes to content rather than back to it.

The helper cannot see why the cap was hit — `fitHeight` means auto is right, the viewport limit means the height must stay manual. Passing `fitHeight` rather than `range.max` here is the smaller version of that, if you want it.

modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/layout/detached-page-components/DetachedPageComponentsView.tsx
```diff
@@ -0,0 +1,394 @@
...
+        const header = panelHeaderRef.current;
+        const content = panelFitContentRef.current;
+        if (panel == null || header == null || content == null) {
+            return;
+        }
+
+        let resizeFrame = 0;
+        const syncToViewport = (): void => {
+            const nextViewport = getViewportBounds();
+            setViewportBounds(nextViewport);
```
edloidas: Small one, and the only one. `syncToViewport` always writes fresh objects — `setViewportBounds(nextViewport)` and the `setLayout` spread below it — so `Object.is` can never bail out, unlike `syncFitHeight` ten lines down which guards with `height === current.height ? current : {...}`. Mounting the panel with a render-counting stub in place of `PageComponentsView` gives 2 renders instead of 1, the extra one from the effect's synchronous `syncToViewport()`, and each no-op `resize` event adds another (3, then 4) with `innerWidth`/`innerHeight` unchanged. Worth an equality check on both writes, since `PageComponentsView` is not memoized and re-renders along with the panel.

edloidas (changes requested):
I reviewed this against #11368, mounted the panel, and measured the resize arithmetic and the rendered box directly.

The feature holds up in the main. The icons are Network and X as the issue asks, both handles are keyboard-operable with real `aria-valuemin`/`valuemax`/`valuenow`, the bottom-edge drag behaves correctly at every size I tried, `pointer-coarse:h-8` gives the handles a usable touch target, and the pointer listeners are torn down on `pointerup`, `pointercancel` and `lostpointercapture`. Clamping against `visualViewport` bounds rather than `innerWidth`/`innerHeight` is a real improvement on what it replaces.

One of the two comments I would not merge without: resizing from the top edge discards the height the user set once the tree is taller than the space above the panel, and the bottom edge then jumps down — which is the issue's own "let users decide how much screen space" case. The other is small, an equality guard on two state writes.

Two things I looked at and deliberately left out. There is no horizontal resize and long component names stay truncated, but the issue frames the problem as long lists and vertical space, so widening reads as a separate request rather than a gap here. And `getViewportBounds()` now runs in a `useState` initializer, which drops the `typeof window === 'undefined'` guard the old file had — nothing in this repo renders the component without a DOM, so it costs nothing today.

reisfmb (commented):
@skjulestad 

I haven't fully reviewed it, but I'd suggest the following:

### Consider extracting the panel behaviour into a hook

`DetachedPageComponentsView.tsx` is ~250 lines of behaviour before its first return: refs, three pieces of state, the viewport/ResizeObserver effect, the header drag, and both resize paths. The render itself is a small fraction of the file.

A single `hooks/useDetachedPanelLayout(active)` alongside the component would own the refs, the state, the effect, the five handlers and the derived height math, returning them to a view that's then focused on the markup: roughly 240 lines in the hook, 130 in the component.

Finally, two smaller follow-ups it opens up:
- splitting layout into separate position and height state
- pulling the resize handle out as its own component

Probably worth doing only after the fixes proposed by @edloidas lands 😉

skjulestad:
Thanks! Resize math is already extracted and the remaining logic are tightly coupled to component refs and markup. Hook would move complexity without reuse, should we not keep it local or might we reuse it in the nearby future? 

Current feedback taken into account but I might need to update tests, let's see if they fail first!

@reisfmb

reisfmb:
@skjulestad 

Yeah, I was afraid of that as well when I quickly checked the code... fair point! The refs/markup coupling is a real objection, and the hook doesn't dissolve it, just relocates it across two files.

The two follow-ups I'd still push for, since they cut complexity rather than move it:

- splitting layout into separate position and height state: one object for two concerns is what makes the merge in applyResize fragile, and it's the same site as the top-edge bug
- pulling the resize handle out as a component: the .map callback currently builds a snapshot, computes a range and assembles four class fragments inline; as a component its inputs are just props

If any part of the hook idea is worth keeping, it's only the viewport/ResizeObserver effect: refs in and viewportBounds/fitHeight out, so no markup knowledge.

skjulestad:
I extracted the resize handle into its own component. I kept layout as one state because top-edge resizing updates position and height atomically; splitting it would make coordination more fragile. I also kept the viewport/ResizeObserver effect local since it has no reuse yet.
Selenium exposed the resize handle overlapping the toolbar, so its hit area now stays inside the panel while the visual indicator remains centered on the border.

PR #11380, opened 2026-09-03
Fix touch behavior in page components tree #11371

### LINKED ISSUE #11371: Fix touch behavior in page components tree
### Current behavior

On touch devices, tapping a component in the page components tree opens its context menu. This prevents selecting the component and showing its selection on the page.

### Expected behavior

A single tap should select the component and synchronize the selection with the page. A long press should open the component context menu, matching the behavior directly on the page.

#### Rationale

Consistent touch behavior makes component selection usable and predictable.

skjulestad:
Restores standard touch behavior in the page components tree. A single tap now selects the component and synchronizes it with the page, while a long press opens the context menu.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/page-components/PageComponentsContextMenu.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/page-components/PageComponentsContextMenu.tsx
@@ -1,14 +1,7 @@
 import { ContextMenu, getIsMobile } from '@enonic/ui';
 import { useStore } from '@nanostores/preact';
 import { Box, Columns2, PenLine, Puzzle } from 'lucide-react';
-import {
-    type MouseEvent as ReactMouseEvent,
-    type ReactElement,
-    type ReactNode,
-    useCallback,
-    useMemo,
-    useState,
-} from 'react';
+import { type ReactElement, type ReactNode, useCallback, useMemo, useState } from 'react';
 import { SaveAsTemplateAction } from '../../../../../../app/wizard/action/SaveAsTemplateAction';
 import { FragmentComponent } from '../../../../../../app/page/region/FragmentComponent';
 import { ComponentPath } from '../../../../../../app/page/region/ComponentPath';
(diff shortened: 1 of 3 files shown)


edloidas (commented):
Looks fine overall - tap selects the component, which is what the issue asks for.

One question: did you test the long press on a real iPhone or iPad? Asking because I suspect it won't open the menu there. Safari on iOS hasn't fired a `contextmenu` event on long press since iOS 13.1, and `ContextMenu.Trigger` in `@enonic/ui` only listens for `onContextMenu` - there's no pointerdown/hold timer in it. Android Chrome synthesizes `contextmenu` from a hold so it works there, but on iOS nothing fires it. The rows are `select-none` too, so a long press would do nothing at all - no menu, no callout. And there's no other way into the menu on a phone, no kebab button on the row.

To be clear, this PR didn't break it. The library has never had a touch path; the old `onClickCapture` just hid that by turning a tap into a synthetic `contextmenu`. Radix handles it with a ~700ms pointerdown timer, which is roughly what our Trigger needs.

So - if it does work on a real device, ignore all this. If it doesn't, I'd file an issue on enonic/npm-enonic-ui to add long press to `ContextMenu.Trigger` and land that first. This PR then needs no changes, just the version bump.

skjulestad:
@edloidas I don't have iphone or android :'/ but I can check with one of my coworkers!

ashklianko:
Tested locally on my iphone in Safari and Chrome - neither is working, touch does not select a page component

ashklianko:
Doesn't work in dev tools open for mobile view. I suggest the issue is assigned to me or Mikita (in this repo or npm-enonic-ui)

ashklianko:
To be fixed im mobile:

- Tapping on the components tree must select (visually too!) an item (not working), scroll to the selected component in preview (working), select the component in Context panel BUT NOT OPEN IT.
- Scrolling in the tree scrolls background form instead (bug)

So it seems that PR is partially working, but needs more updates to match expected behavior.

skjulestad:
Scroll bug has now been resolved here: https://github.com/enonic/app-contentstudio/pull/11379

skjulestad:
iphone to be solved outside this issue

edloidas:
Fix failing tests.
Create Enonic UI issue and refer it here.
Then we are done and ready to merge.

skjulestad:
@edloidas issue created: https://github.com/enonic/npm-enonic-ui/issues/537

PR #11397, opened 2026-09-07
Improve content editor layout on narrow screens  #11369

### LINKED ISSUE #11369: Improve content editor layout on narrow screens
Replace the editor action control with a localized Actions dropdown containing visible actions.
Change the display name text size to 1.5rem in compact layouts.
Use 8px horizontal padding in the editor preview toolbar without changing vertical padding.
Keep publication status visible without overflowing long labels.
Group widget and refresh controls together in the editor layout.

skjulestad:
Replace the editor action control with a localized Actions dropdown containing visible actions.
Change the display name text size to 1.5rem in compact layouts.
Use 8px horizontal padding in the editor preview toolbar without changing vertical padding.
Keep publication status visible without overflowing long labels.
Group widget and refresh controls together in the editor layout.

### DIFF
--- a/modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/DisplayNameInput.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/DisplayNameInput.tsx
@@ -190,7 +190,7 @@ export const DisplayNameInput = (): ReactElement => {
                     onClick={startEditing}
                     onFocus={startEditing}
                     className={cn(
-                        'block w-full min-w-64 border-0 border-l-1 bg-transparent text-left text-[2rem] font-semibold',
+                        'block w-full min-w-64 border-0 border-l-1 bg-transparent text-left text-[1.5rem] sm:text-[2rem] font-semibold',
                         'px-2.5 py-1 pl-4.5 rounded-none',
                         'hover:not-disabled:border-l-4 hover:not-disabled:pl-3.75',
                         'focus:outline-none focus:ring-0 focus:ring-offset-0',
@@ -222,7 +222,7 @@ export const DisplayNameInput = (): ReactElement => {
                     disabled={readOnly}
                     className={cn(
                         'block w-full min-w-64 resize-none overflow-hidden whitespace-pre-wrap break-words bg-transparent',
-                        'border-0 border-l-1 text-[2rem] font-semibold px-2.5 py-1 pl-4.5',
+                        'border-0 border-l-1 text-[1.5rem] sm:text-[2rem] font-semibold px-2.5 py-1 pl-4.5',
                         '[&:hover,&:focus]:border-l-4 [&:hover,&:focus]:pl-3.75 placeholder:text-subtle/50 rounded-none',
                         'transition-highlight focus:outline-none disabled:select-none disabled:cursor-not-allowed',
                         'focus:ring-0 focus:ring-offset-0 focus:border-transparent',
--- a/modules/lib/src/main/resources/assets/js/v6/widgets/preview-panel/ui/PreviewToolbar.tsx
+++ b/modules/lib/src/main/resources/assets/js/v6/widgets/preview-panel/ui/PreviewToolbar.tsx
@@ -14,12 +14,14 @@ type PreviewToolbarProps = {
     item?: ContentSummaryAndCompareStatus | null;
     onRefresh?: () => void;
     hideInMobileMode?: boolean;
+    editorLayout?: boolean;
 };
 
 const PreviewToolbar = ({
     item = null,
     onRefresh,
     hideInMobileMode = false,
+    editorLayout = false,
 }: PreviewToolbarProps): ReactElement | null => {
     const mode = useStore($contextPanelMode);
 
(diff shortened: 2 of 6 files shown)


modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/DisplayNameInput.tsx
```diff
@@ -190,7 +190,7 @@ export const DisplayNameInput = (): ReactElement => {
                     onClick={startEditing}
                     onFocus={startEditing}
                     className={cn(
-                        'block w-full min-w-64 border-0 border-l-1 bg-transparent text-left text-[2rem] font-semibold',
+                        'block w-full min-w-64 border-0 border-l-1 bg-transparent text-left text-[1.5rem] sm:text-[2rem] font-semibold',
```
edloidas, line 193: ### The 1.5rem size flips at a 640px viewport, so the wizard's own mobile layout still shows 2rem

I swept the viewport against a local instance on this branch. `font-size` on the display name is 24px at 639px wide and 32px at 641px — `sm:` is a viewport media query, so it switches at 640px.

The wizard's own mobile layout switches at 720px, not 640px: the live panel measures 0px wide at 719px and 360px at 721px, which matches `LayoutTokens.contextPanel.mobileThreshold` (`ResponsiveRanges._540_720`, and `isFitOrSmaller(size)` is `size <= 720`).

So the whole 641–719px band is the wizard in its mobile layout with a 2rem display name. At 700px both the button here and the textarea on line 225 measured 32px.

Separately, `sm:` cannot see the form pane, and that pane is resizable to `FORM_MIN_WIDTH = 360` (`WizardLayout.tsx:18`) on a wide screen, where the name also stays 2rem.

A container query on the form pane covers both cases. `max-[720px]:` covers the first and keys it to the threshold the layout already uses.

modules/lib/src/main/resources/assets/js/v6/widgets/preview-panel/ui/PreviewToolbar.tsx
```diff
@@ -30,18 +32,31 @@ const PreviewToolbar = ({
...
                 aria-label="Preview toolbar"
                 className={cn(
-                    '@container bg-surface-neutral h-15 px-5 py-3.75 flex items-center justify-between border-b border-bdr-soft',
+                    '@container bg-surface-neutral h-15 py-3.75 flex items-center justify-between border-b border-bdr-soft',
+                    editorLayout ? 'px-2' : 'px-5',
                     hideInMobileMode && mode === 'mobile' && 'hidden',
                 )}
             >
-                <PreviewToolbarVersionHistoryItem contentSummary={item.getContentSummary()} />
+                <PreviewToolbarVersionHistoryItem contentSummary={item.getContentSummary()} showStatus={editorLayout} />
```
edloidas, line 40: ### `showStatus` keeps the version-history label inline where it used to collapse, and a long status label then overflows the row

`FrameContainer.ts:26` is the only site passing `editorLayout: true`, unconditionally, so `showStatus` is always true here — dropping `hidden @sm:inline` from the label and `@max-sm:p-0` from the button.

At a 730px window the live panel is already at its 360px minimum (`LIVE_MIN_WIDTH`, `WizardLayout.tsx:19`) and the toolbar measures 369px — below its own `@sm` of 384px. Measured there, the button is 117px; collapsing the label and zeroing the padding brings it to 36px — 81px of a 369px row. With English labels nothing overflows: `scrollWidth` and `clientWidth` are both 369.

It overflows once the status string is long. Substituting the real `status.scheduled` value from `phrases_ru.properties:388`, `Запланирован на публикацию`, the button goes to 268px and `scrollWidth` to 390 against a `clientWidth` of 369 — the refresh control is cut off at the right edge. That was a text substitution in the live page, not a genuinely scheduled Russian-locale content, so read the 21px as the size of the gap.

`editorLayout ? 'px-2' : 'px-5'` satisfies the padding clause on its own. If the always-on label is wanted, it needs room to shrink.

modules/lib/src/main/resources/assets/js/v6/widgets/preview-panel/ui/PreviewToolbar.tsx
```diff
@@ -30,18 +32,31 @@ const PreviewToolbar = ({
...
 
-                <div className="flex gap-2 @md:gap-5 flex-nowrap shrink-0">
-                    <PreviewToolbarEmulatorSelector />
-                    <PreviewToolbarWidgetSelector />
-                </div>
+                {editorLayout ? (
```
edloidas, line 42: Two small ones, grouped.

The `editorLayout` branch also regroups the right-hand controls — the emulator selector becomes its own `justify-between` child and refresh moves in beside the widget selector — and the new group is plain `gap-2`, dropping the `@md:gap-5` the other branch keeps. At the default width those three children measure 117 / 86 / 164px. The issue asks for padding only, so this reads as a design decision rather than a consequence of one. Is that correct?

In `PreviewToolbarVersionHistoryItem`, lines 48 and 54 both became `mobile || showStatus`, but `startIconClassName` on line 51 still keys on `mobile` alone — and that line predates this branch. Measured live, the editor toolbar's History icon is 14x14 beside a full label, where the mobile button gets `size-5`. Which of the two is right is your call; today they disagree.

modules/lib/src/main/resources/assets/js/v6/pages/wizard/ui/content-wizard-tabs/DisplayNameInput.tsx
skjulestad: Decided to stick with screen size for now to avoid text size jump as container is resizable.

modules/lib/src/main/resources/assets/js/v6/widgets/preview-panel/ui/PreviewToolbar.tsx
skjulestad: This is design choice yes, will update issue description.

edloidas (approved):

PR #11406, opened 2026-09-09
Issue 11339

skjulestad:
Main item labels and paths remain single-line and truncate when needed.
Status appears on a separate row below the label and path.
Reference icons appear directly after status.
Remove actions remain in the upper-right corner.
Other items and dependencies use the same status-row alignment.
Reference icons appear white before visiting and blue after visiting.
Tablet and desktop retain their existing layout.


skjulestad:
PR already done and closed, commit to wrong branch reverted.

PR #11407, opened 2026-09-09
Dialog content items patch for mobile #11370

### LINKED ISSUE #11370: Dialog content items patch for mobile
### Summary
Content metadata and actions compete for limited horizontal space in mobile dialogs. Reorganize item rows below the sm breakpoint so labels, paths, statuses, and actions remain readable and consistently aligned. Preserve existing tablet and desktop layouts.

### Current behavior
- Status badges reduce space available for main item labels and paths.
- Reference actions and status badges use inconsistent placement between main and dependent items.
- Long item names can push action icons out of alignment.

### Expected behavior
- Main item labels and paths remain single-line and truncate when needed.
- Status appears on a separate row below the label and path.
- Reference icons appear directly after status.
- Remove actions remain in the upper-right corner.
- Other items and dependencies use the same status-row alignment.
- Reference icons appear white before visiting and blue after visiting.
- Tablet and desktop retain their existing layout.

### Acceptance criteria

- [ ] Delete, Duplicate, Sort, Publish, Unpublish, Request Publishing, and Create Issue dialogs display correctly on mobile.
- [ ] Main labels and paths remain single-line with truncation.
- [ ] Main item status appears below its label and path.
- [ ] Reference icons appear after status on mobile.
- [ ] Remove actions remain in the upper-right corner.
- [ ] Other items and dependencies retain single-line labels.
- [ ] Reference-link visited state changes from white to blue.
- [ ] Existing tablet and desktop layouts remain unchanged.

<sub>Drafted with AI assistance</sub>

skjulestad:
Main item labels and paths remain single-line and truncate when needed. Status appears on a separate row below the label and path. Reference icons appear directly after status.
Remove actions remain in the upper-right corner.
Other items and dependencies use the same status-row alignment. Reference icons appear white before visiting and blue after visiting. Tablet and desktop retain their existing layout.


modules/lib/src/main/resources/assets/js/v6/features/shared/ContentReferencesLink.tsx
```diff
@@ -42,7 +44,8 @@ export const ContentReferencesLink = forwardRef<HTMLAnchorElement, ContentRefere
                 ref={ref}
                 className={cn(
                     className,
-                    'self-stretch px-2 visited:text-main active:bg-transparent data-[active=true]:bg-transparent',
+                    'self-stretch justify-center px-2 text-alt [--color-link-visited:var(--color-blue)] active:bg-transparent data-[active=true]:bg-transparent',
```
edloidas, line 47: My brother in Christ, why are you patching the global visited link color instead of applying new `visited:` styles and and overriding `visited:text-link-visited` in Link?

edloidas: `text-alt` us always `white` on light and dark themes, so on light theme it will be invisible on light background.

edloidas (changes requested):
In addition to comments, `statusBelowLabelBelowSm`, `mainItemButtonLayoutBelowSm` and others are just a bad smell. They would bring much complexity, that would backfire in the future. This has to be dome simpler. There are lots of things that are already hard to track, and excessive.
Styling can be done with CSS. If it's not universal and require different component, create one.
`rightSlotAfterStatusBelowSm` is just bad name and usecase, this must not be a boolean flag, and should not be stacked with anything else. If any, I see it transformed to `slotsOrder: 'status-first' | ...` or something like that.

So, avoid boolean flags for styling. If you need some rendering order or disable some slots, use it, but keep expressive and avoid boolean flags, that may conflict with others in the future.

skjulestad:
Applied fixes.
NOTE: I added px-1 as focus-ring for show ref button on desktop was overlapping.
After applied:
image

Before applied:
image

PR #11413, opened 2026-09-10
Layout fragments cannot be nested inside other layouts #11363

### LINKED ISSUE #11363: Layout fragments cannot be nested inside other layouts
Regression: in legacy CS we could not add a fragment with a layout into another layout, but in the new ui CS it is possible.

https://developer.enonic.com/docs/content-studio/stable/editor/page-editor#layout

skjulestad:
Prevent layout fragment from nesting inside another layout.


ashklianko (changes requested):
Found an issue: add an empty fragment into a Layout component, drag this empty fragment out of layout and try assigning it a real fragment with layout - it says that "Layout within layout is not allowed." though we moved this fragment out

edloidas (approved):
Layout fragments can no longer be dropped into a layout from the Page Components tree, restoring the legacy restriction. An unresolved fragment is treated as a layout until its content loads, so a fast drag right after opening the page cannot slip a layout fragment into a layout either. Fragment resolution is scoped to the latest rebuild, so a slow response from an earlier page state cannot flip a node's flag after the tree has changed.

PR #11428, opened 2026-09-14
Menu improvements in mobile resolution #11416

### LINKED ISSUE #11416: Menu improvements in mobile resolution
In the menu panel:
1. Make the panel wider, so that it almost reaches the avatar icon 
2. Don’t apply background to the “X” icon
3. Replace the project icon with “project toggle” icon which will open the project selection dialog 
4. Show project name between “X” and the project toggle icon, add ellipsis wrap. Make sure there’s enough place between “…” and the project toggle icons when a long project name is cut.
5. Remove “Content Studio” above the menu
6. Make sure entire row is highlighted for the current menu item, not just the icon

In the toolbar
1. Remove the project toggle icon - we have it inside the panel now
2. Show project name only when you’re inside the “main” Content Studio. If you’re in an extension (Archive, GraphQL Playground etc.) show the name of the extension (see the screenshot).

skjulestad:
Expand panel
Remove background from the sidebar menu toggle.
Replace the project icon with the project selection button. Add a truncated project name in the sidebar menu.
Match layout for the sidebar menu with appbar, adjusted positioning of the name to fit with both, aligner to the edge of the action-button instead of action button text. Highlight the entire row per selected widget in the sidebar.

Remove Content Studio from the sidebar menu.
Remove project toggle from mobile BrowseAppBar.tsx Show only project name on browse content, show extension or settings for settings.


modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseAppBar.tsx
```diff
@@ -10,16 +10,15 @@ import { useI18n } from '../../shared/lib/hooks/useI18n';
...
 import type { IssueStatsJson } from '../../../app/issue/json/IssueStatsJson';
 import { LegacyElement } from '../../shared/ui/LegacyElement';
 import { BROWSE_SIDEBAR_ID, BROWSE_SIDEBAR_TOGGLE_ID } from './browseSidebar.constants';
 import { $isBrowseSidebarOpen, setBrowseSidebarOpen } from './model/browseSidebar.store';
 
 const $widgetState = computed($activeWidget, (activeWidget) => ({
     appName: activeWidget?.getDisplayName() ?? '',
-    isIssuesButtonVisible: isMainWidget(activeWidget),
-    isProjectSelectorVisible: activeWidget == null || activeWidget.getConfig().getProperty('context') === 'project',
+    isContentBrowseActive: activeWidget == null || isContentBrowseWidget(activeWidget),
```
edloidas, line 21: This now only matches `studio:main`, but Archive sets `context: "project"` in its descriptor (`app-contentstudio-plus/src/main/resources/admin/extensions/archive/archive.yaml`), so it used to match here too.

So on desktop, open Archive and there's no project name and no switcher. The new switcher lives in the sidebar, which is `sm:hidden`, and the `ProjectIcon` on the rail isn't clickable — so you can't tell whose archive you're looking at, or change project, without going back to Content first.

The new `BrowseSidebar` test covers this exact case with `plus:archive`, but only the mobile path.

Simplest fix is to keep the `context === 'project'` check in the condition.

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseAppBar.tsx
```diff
@@ -50,10 +49,13 @@ export const BrowseAppBar = (): ReactElement => {
     const isMobileSidebarOpen = !sm && isSidebarOpen;
 
     return (
-        <header className="bg-surface-neutral h-15 px-3.5 sm:pl-5 py-2 pr-24 flex items-center gap-2.5 border-b border-bdr-soft">
+        <header className="bg-surface-neutral h-15 px-3.5 sm:pl-5 py-2 pr-24 flex items-center gap-2 border-b border-bdr-soft">
             <Toggle
                 id={BROWSE_SIDEBAR_TOGGLE_ID}
-                className={cn('relative z-20 size-9 shrink-0 p-0 sm:hidden', !hasMultipleProjects && 'max-sm:mr-auto')}
+                className={cn(
+                    'relative z-20 size-9 shrink-0 p-0 active:bg-transparent data-[active=true]:bg-transparent sm:hidden',
```
edloidas: Please, don't forget to test things on both themes: light and dark.

Dropping the background here leaves `text-alt` behind. It's paired with `bg-btn-active` in Button's base, and tailwind-merge only removes the background half. `text-alt` is white, and the header behind it is white in light theme, so the X vanishes while the panel is open. I rendered it with the project's own compiled CSS and got `color: rgb(255, 255, 255)` on `rgb(255, 255, 255)`.

Dark theme is fine. Escape and the backdrop still close the panel, so nobody's stuck — you just can't see the button.

`active:text-main data-[active=true]:text-main` alongside the two background overrides sorts it.

image

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseSidebar.tsx
```diff
@@ -98,92 +95,100 @@ export const BrowseSidebar = (): ReactElement => {
...
+                    'bg-surface-neutral absolute z-20 flex h-dvh w-[clamp(220px,100%_-_100px,340px)] flex-col items-start gap-5 border-r border-t-0 border-bdr-soft px-3.5 pt-3',
                     'sm:z-auto sm:w-15 sm:translate-x-0 sm:items-center sm:gap-10 sm:px-1.75 sm:py-2.5 sm:pointer-events-auto',
                     isOpen ? 'translate-x-0' : '-translate-x-full pointer-events-none',
                 )}
                 aria-label={sidebarLabel}
                 aria-hidden={isMobileSidebarClosed || undefined}
                 inert={isMobileSidebarClosed}
             >
-                <div className="flex h-11 w-full shrink-0 items-center justify-end sm:contents">
+                <div className="flex ml-11 self-stretch items-center sm:hidden">
```
edloidas, line 106: TBD. With a single project there's no icon here at all — the issue said replace the project icon, and the desktop button stays visible-but-disabled in that situation.

And the row sets no gap: with a long name the ellipsis ends up 8px from the switcher glyph, which is just the button's own padding. Fine if that's what you meant by "enough place".

edloidas (changes requested):
Two things I'd fix before this goes in, both on the app bar. Everything else looks right to me — the panel width, the row highlight, the title switching to the extension name.

Two small ones I didn't put on lines: the new sidebar header doesn't check `noProjectMode`, so with no available project you get an empty `<h1>` where the app name used to be on mobile. And the mobile version label is `aria-hidden` now, while the `<p>` that carried it is `max-sm:hidden`, so on mobile the version isn't in the accessibility tree at all.

modules/lib/src/main/resources/assets/js/v6/pages/browse/BrowseAppBar.tsx
skjulestad: Incoming fix for this.

alansemenov:
Issues discovered after the last testing:
1. Horizontal separator between the header and the toolbar is gone, both in desktop and mobile resolutions

2. Project name/switcher in the heading was supposed to be removed only in mobile resolution. Now it's gone everywhere and the font has changed. It should remain 100% as before (check Nightly) in the desktop resolution.

3. Looks especially bad in CS+ / Archive extension (project selector is gone, all horizontal separators are screwed up)

was:
image

now
image

PR #11430, opened 2026-09-15
Favicon does not adapt to the browser color scheme #11429

### LINKED ISSUE #11429: Favicon does not adapt to the browser color scheme
The Content Studio favicon is a solid black tree on a transparent background, so it renders as a black-on-dark smudge in any browser running a dark tab strip. Every declared icon — the `favicon-*.png` set plus the android and apple-touch variants — is the same single-colour artwork, and there is no light variant to fall back to. To reproduce, switch the OS or browser to dark appearance and open any Content Studio tab. The icon is effectively unreadable, which makes the tab hard to pick out of a row of pinned or narrow tabs.

Separately, the `mask-icon` link points at `{{assetsUri}}/icons/icon-black.svg`, which does not exist in the app module's assets and returns 404.

#### Rationale

Content Studio already ships a dark theme and declares `color-scheme: light dark`, so a favicon that only works on light backgrounds is an unfinished part of that support. The tab icon is the most persistent piece of branding the app has, and it is the one surface still hardcoded to a single theme.

#### Implementation Notes

An SVG favicon can carry its own `prefers-color-scheme` media query, so one file covers both themes while the PNG set stays as the fallback. Note that the content wizard swaps the favicon for the content's own icon (`clearFavicon`/`updateFavicon` in `main.ts`), and Chrome prefers an SVG icon over the PNG ones — the SVG link therefore has to leave the document while that swap is in effect.

ashklianko:
Added `favicon.svg` carrying a `prefers-color-scheme` media query, black on light and white on dark. Declared it in `main.html` ahead of the existing PNG set, which stays as the fallback. Repointed `mask-icon` from the missing `icons/icon-black.svg` to the new file. Detached the SVG link in `clearFavicon` and restored it in `updateFavicon`, since Chrome prefers it over the PNGs the wizard swaps for the content icon.


modules/app/src/main/resources/assets/icons/favicons/favicon.svg
```diff
@@ -0,0 +1,24 @@
...
+            fill: #000000;
+        }
+
+        @media (prefers-color-scheme: dark) {
+            path {
+                fill: #ffffff;
+            }
+        }
+    </style>
+    <g fill-rule="evenodd" transform="translate(-25, 0)">
```
edloidas, line 13: Drop transform and update viewBox to `viewBox="2 0 326 326"`

modules/app/src/main/resources/assets/js/main.ts
edloidas, line 167: Probably wrong now.

modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -182,13 +182,22 @@ let invalidEditUrlNotificationPending: boolean = false;
...
         faviconCache[href] = link;
         link.setAttribute('href', ImgEl.PLACEHOLDER);
     });
+
+    // Chrome picks the svg favicon over the png ones, so it has to leave the document entirely
+    // while the tab shows the content icon
+    $('link[rel*=icon][type="image/svg+xml"]').each((index, link: HTMLElement) => {
+        detachedSvgFavicon = link;
+        link.remove();
+    });
```
edloidas, line 200: We will be dropping jQuery. Don't use it here.

modules/app/src/main/resources/assets/js/main.ts
```diff
@@ -182,13 +182,22 @@ let invalidEditUrlNotificationPending: boolean = false;
 
 let wizardParams: ContentWizardPanelParams | undefined;
 
+let detachedSvgFavicon: HTMLElement | undefined;
```
edloidas, line 185: `detachedSvgFavicon` mirrors state the DOM already holds. It fuses _which_ element was detached with _whether_ anything is detached. Rewrite, so that the first is `svgFaviconLink`, a const resolved once at load - the link never changes, so there's nothing to discover at detach time. The second is `svgFaviconLink.isConnected`, read from the DOM instead of mirrored beside it. Once both are covered, nothing is left for `detachedSvgFavicon` to hold, so it goes away along with its two reassignment sites.

edloidas (changes requested):

reisfmb (approved):

edloidas (approved):

PR #11442, opened 2026-09-23
Inspect after inserting a component on mobile #11372

### LINKED ISSUE #11372: Inspect after inserting a component on mobile
Open inspect panel and close detached page component after inserting component into detached page component.

skjulestad:
Open the inspection panel and close the detached page component on mobile after inserting a component.


ashklianko (changes requested):
The "Inspect" menu item doesn't collapse the panel. On mobile, InspectItem (PageComponentsContextMenu.tsx:241) opens the inspector too (bridge.ts:104). It doesn't collapse the detached page components panel, though, so the panel stays on top of the inspector. That's the same situation this PR fixes for insert. A more robust fix would collapse the panel once, in the mobile INSPECT handler: in bridge.ts next to setContextOpen(true), or in a shared helper. That would cover the "Inspect" item and INSPECT events from the preview/iframe, not just insert. If that's out of scope for #11372, at least add the same call to InspectItem.

